PR #4, opened 2017-07-13
Repository mirroring 4458

ikapelyukhin:


lib/rmt/downloader.rb
```diff
@@ -0,0 +1,129 @@
...
+    filename
+  end
+
+  def make_request(remote_file, local_file, checksum_type, checksum_value, &complete_callback)
+    uri = URI.join(@repository_url, remote_file)
+    downloaded_file = Tempfile.new('rmt')
+
+    request = Typhoeus::Request.new(uri.to_s, followlocation: true)
+    request.on_headers do |response|
+      if (URI(uri).scheme != 'file' and response.code != 200)
```
ikapelyukhin: Yep, there are no headers for `file:///` scheme apparently, so `response.code == 0` :-)

lib/rmt/mirror.rb
```diff
@@ -0,0 +1,86 @@
...
+    @downloader = RMT::Downloader.new(
+      @repository_url,
+      temp_dir.to_s,
+      @logger
+    )
+
+    begin
+      @downloader.download('repodata/repomd.xml.key')
+      @downloader.download('repodata/repomd.xml.asc')
+    rescue RMT::Downloader::Exception # rubocop:disable Lint/HandleExceptions
```
ikapelyukhin: The repository might be unsigned, so those files wouldn't exist and the signatures aren't referenced in the metadata.

lib/rmt/downloader.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  def verify_checksum(filename, checksum_type, checksum_value)
+    hash_function = checksum_type.gsub(/\W/, '').upcase.to_sym
+    unless (KNOWN_HASH_FUNCTIONS.include? hash_function)
+      raise Exception.new("Unknown hash function #{checksum_type}")
+    end
+
+    digest = Digest.const_get(hash_function).file(filename)
+
+    raise Exception.new('Checksum doesn\'t match') unless (checksum_value == digest.to_s)
```
ikapelyukhin: Will replace with `RMT::Downloader::Exception`.

lib/rmt/downloader.rb
```diff
@@ -0,0 +1,129 @@
...
+    local_file
+  end
+
+  def download_multi(files)
+    @queue = files
+    @hydra = Typhoeus::Hydra.new(max_concurrency: @concurrency)
+
+    @concurrency.times { download_one }
+
+    loop do
```
ikapelyukhin: Rubocop rules :-)

suse-tests-pass:
@thutterer will review your pull request :dancers: check https://trello.com/c/XXJTyKsa/4458-13-rmt-serve-mirrored-rpms-via-https-with-no-auth
  @thutterer: Please review this pull request using our guidelines:

lib/rmt/downloader.rb
```diff
@@ -0,0 +1,137 @@
+module RMT
```
thutterer: That shouldn't be required here as you have the `lib/rmt.rb` file which already does that.
Haven't tried this yet, but I'm quite sure you can just remove `module RMT; end` here.

lib/rmt/mirror.rb
```diff
@@ -0,0 +1,86 @@
+module RMT
```
thutterer: see other comment

lib/rmt/rpm.rb
```diff
@@ -0,0 +1,10 @@
+module RMT
+  module Rpm
+  end
+end
+
+require_relative 'rpm/file_entry'
```
thutterer: Any reason to not put the requires to the very top of the file?

lib/rmt/rpm/base_parser.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+  end
+
+  protected
+
+  def get_attribute(attributes, name)
+    attributes.select {|e| e[0] == name }.first[1]
+  end
+
```
thutterer: Too many empty lines.
 (me playing rubocop)

lib/rmt/downloader.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      verify_checksum(downloaded_file.path, checksum_type, checksum_value) if (checksum_type and checksum_value)
+    rescue StandardError => e
+      downloaded_file.unlink
+      raise e
+    end
+
+    FileUtils.mv(downloaded_file.path, local_file)
+
+    @logger.info("D #{File.basename(local_file)}")
```
kirushik: Not particularly helpful log message, is it?

Does `D` stand for `DEBUG`? Maybe you can use `.debug` method instead then?

ikapelyukhin: `D` for `Download` (I think) -- inspired by output of SMT :grimacing:

lib/rmt/rpm/base_parser.rb
ikapelyukhin: But Rubocop doesn't complain :-)

lib/rmt/downloader.rb
ikapelyukhin: Yeah, that's left over since before `lib/rmt.rb` came to life.  I'll clean that up.

bin/rmt-mirror
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env ruby
+
+require_relative '../config/boot'
+require 'pathname'
+require 'config'
+require 'pp'
```
thutterer: left-over `pp`  from developing, I guess

lib/rmt/mirror.rb
```diff
@@ -0,0 +1,86 @@
...
+end
+
+class RMT::Mirror
+
+  def initialize(mirroring_base_dir, repository_url, local_path, mirror_src = false, logger = nil)
```
thutterer: A method with so many arguments should use [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments). Then the ordering does not matter anymore and an invocation is way more readable ... I had to check the method definition here to understand what was going on in `bin/rmt-mirror`

spec/lib/rmt/downloader_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        it('has correct content') { expect(File.read(subject)).to eq(content) }
+      end
+    end
+  end
+
+  describe '#download_multi' do
+    let(:files) { %w(package1 package2 package3) }
+    let(:checksum_type) { 'SHA256' }
+    let(:queue) do
+      queue = []
```
kirushik:
```ruby
files.map do |file|
  RMT::Rpm::FileEntry.new(
    file,
    checksum_type,
    Digest.const_get(checksum_type).hexdigest(file),
    :rpm
  )
end
```
without any intermediate collection `query` would be mo Ruby-ish.

bin/rmt-mirror
```diff
@@ -0,0 +1,32 @@
...
+
+require_relative '../lib/rmt'
+require_relative '../lib/rmt/mirror'
+require_relative '../lib/rmt/rpm'
+require_relative '../lib/rmt/downloader'
+
+repository_url = ARGV[0]
+local_path = ARGV[1]
+
+unless (repository_url and local_path)
```
thutterer: Not sure if this is part of this card, but currently you do not validate user input nor catch its exceptions. Calling `./bin/rmt-mirror foo bar` fails with an URI::BadURIError exception and a stacktrace.

spec/lib/rmt/downloader_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      files.each do |file|
+        expect(WebMock).to(
+          have_requested(:get, "http://example.com/#{file}").with(headers: headers)
+        )
+      end
+    end
+
+    it 'saved all files' do
+      files.each do |file|
+        expect(File.exist?(File.join(dir, file))).to eq(false)
```
kirushik: Am I misreading this, or we're expecting all the files **not** to exist here?

bin/rmt-mirror
ikapelyukhin: I guess later on it will read repos configuration from the DB. There's a different card about configuring which repos to mirror. For now it takes repo URL as parameter.

spec/lib/rmt/downloader_spec.rb
ikapelyukhin: Whoops, example description is incorrect. It is supposed to check that it doesn't write files if there were exceptions raised.

lib/rmt/mirror.rb
```diff
@@ -0,0 +1,86 @@
...
+      repomd_parser.parse
+
+      repomd_parser.referenced_files.each do |reference|
+        @downloader.download(reference.location, reference.checksum_type, reference.checksum)
+        @primary_files << reference.location if (reference.type == :primary)
+        @deltainfo_files << reference.location if (reference.type == :deltainfo)
+      end
+
+      local_repo_dir = File.join(@mirroring_base_dir, @local_path)
+      FileUtils.mkpath(local_repo_dir) unless Dir.exist?(local_repo_dir)
```
thutterer: Fails with a stacktrace if the user who runs rmt-mirror lacks the privileges to mkdir that folder.

bin/rmt-mirror
thutterer: :ok:

lib/rmt/downloader.rb
kirushik: `↓` might be a better symbol?

ikapelyukhin: Well, smiling cat face would be even better. :-)
On a serious note though, I guess it depends on the ability of the terminal to render all the funky symbols.

ikapelyukhin:
@thutterer Yep, that's the version of Nokogiri that we currently have in Leap 42.3 repos (which is what is used for packaging the RPM).

lib/rmt/downloader.rb
kirushik: And `×` for errors

thutterer: > On a serious note though, I guess it depends on the ability of the terminal to render all the funky symbols.

Yep, as much as I :heart: emojis, let's stick to ASCII please :wink: 

But related to this conversation (and unrelated to the pull request itself), I'd suggest to fine-tune the Logger at some point. Having the microseconds of the timestamp in the log feels like wasted bytes to me:

```
I, [2017-07-25T15:50:07.755331 #19185]  INFO -- : D yp-tools-2.14-6.1.x86_64.rpm
```

Also, does the first character "I" also mean "INFO"? Then one of the two is redundant.

lib/rmt/rpm.rb
ikapelyukhin: Hmm, for better modularity. All RPM stuff is required in RPM module. You can require just the downloader without any of the RPM stuff.

lib/rmt/downloader.rb
```diff
@@ -72,7 +58,7 @@ def process_queue
       begin
         make_request(remote_file, local_file, request_fiber, queue_item[:checksum_type], queue_item[:checksum])
       rescue RMT::Downloader::Exception => e
-        @logger.info("E #{File.basename(local_file)} - #{e}")
+        @logger.warn("× #{File.basename(local_file)} - #{e}")
```
thutterer: What kind of char is `×`? Is it supported on _all_ terminals?
Also, is it appropriate to use for logging? Does it have any meaning? Looks weird to me.

lib/rmt/downloader.rb
```diff
@@ -117,7 +114,7 @@ def make_request(remote_file, local_file, request_fiber, checksum_type = nil, ch
 
     FileUtils.mv(downloaded_file.path, local_file)
 
-    @logger.info("D #{File.basename(local_file)}")
+    @logger.info("↓ #{File.basename(local_file)}")
```
thutterer: Same as for `×` ... might look nice but not sure if we should invent our own symbol language.

bin/rmt-mirror
```diff
@@ -23,10 +22,15 @@ Config.load_and_set_settings(
...
+begin
+  RMT::Mirror.new(
+    mirroring_base_dir: Settings.mirroring.base_dir,
+    repository_url: repository_url,
+    local_path: local_path,
+    mirror_src: Settings.mirroring.mirror_src,
+    logger: Logger.new(STDOUT)
+  ).mirror
+rescue RMT::Mirror::Exception => e
+  warn e.to_s
```
kirushik: Isn't it `fatal` level if you're going to exit with an error code right now?

lib/rmt/mirror.rb
```diff
@@ -1,9 +1,10 @@
-module RMT
-end
-
 class RMT::Mirror
 
-  def initialize(mirroring_base_dir, repository_url, local_path, mirror_src = false, logger = nil)
+  class RMT::Mirror::Exception < RuntimeError
```
thutterer: For these empty classes, I personally prefer `class RMT::Mirror::Exception < RuntimeError; end`
I'm quite sure there is a RuboCop rule for that as-well. But I'm not insisting on that. It should just be same ruleset as in Glue :policeman:

bin/rmt-mirror
ikapelyukhin: I intended that a way to output to STDERR, not as a way to raise an exception.

lib/rmt/mirror.rb
```diff
@@ -21,18 +22,30 @@ def mirror
...
+      local_path: temp_dir.to_s,
+      logger: @logger
     )
 
     begin
       @downloader.download('repodata/repomd.xml.key')
       @downloader.download('repodata/repomd.xml.asc')
-    rescue RMT::Downloader::Exception # rubocop:disable Lint/HandleExceptions
+    rescue RMT::Downloader::Exception
+      @logger.info('Repository metadata signatures are missing')
```
thutterer: Is this really info-level (Non-critical) or should that better be a warning? I don't know, up to you.

ikapelyukhin: Those files are repository metadata signatures, and unsigned repositories exist. And these signatures aren't referenced in the XML.

lib/rmt/downloader.rb
ikapelyukhin: It's a multiplication sign, it should be in base ASCII charset. But yeah, it looks unorthodox :-)

lib/rmt/downloader.rb
```diff
@@ -0,0 +1,132 @@
...
+      downloaded_file.write(chunk)
+    end
+    request.on_complete do |response|
+      request_fiber.resume(response) if request_fiber.alive?
+    end
+
+    response = Fiber.yield(request) # yields headers
+
+    begin
+      if (URI(uri).scheme != 'file' and response.code != 200)
```
kalabiyau: why and over && ?

ikapelyukhin: @kalabiyau Y U HATE WORDS :grimacing:

thutterer: @ikapelyukhin They have lower precedence and are usually avoided for things other than flow-control. See https://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

Better change it to `&&` here.

Dockerfile
```diff
@@ -10,4 +10,4 @@ WORKDIR /srv/www/potato/
 
 RUN ruby -e 'require "securerandom"; puts SecureRandom.hex(64)' > .secret
 
-CMD tail -f /dev/null
+CMD bash -c 'bundler.ruby2.4 --system --without with_c_extensions && rm -rf .bundle/config && rails db:migrate && rails s -P /tmp/rmt.pid'
```
kalabiyau: could we split the line for the readability but maintaining one layer?

ikapelyukhin: @kalabiyau I'll look into rebuilding the container based on Leap 42.3 image.

PR #5, opened 2017-07-26
Make Travis check the master and branches

kalabiyau:
🥔

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,9 @@ rvm:
 script:
   - 'bundle exec rspec spec'
   - 'bundle exec rubocop'
+services:
+  - postgresql
+before_script:
+  - psql -c "CREATE DATABASE potato;" -U postgres
+  - psql -c "CREATE USER potato WITH PASSWORD 'potato';" -U postgres
+  
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ services:
 before_script:
   - psql -c "CREATE DATABASE potato_test;" -U postgres
   - psql -c "CREATE USER potato WITH PASSWORD 'potato';" -U postgres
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: ruby
+rvm:
+  # The version used at SLE12
+  - 2.1.2
```
ikapelyukhin: Hmm, which version are we aiming at? The RPM package is build for Ruby 2.4.

.travis.yml
```diff
@@ -13,4 +13,3 @@ services:
 before_script:
   - psql -c "CREATE DATABASE potato;" -U postgres
   - psql -c "CREATE USER potato WITH PASSWORD 'potato';" -U postgres
```
ikapelyukhin: `grant all privileges on database potato to potato;`?

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: ruby
+rvm:
+  # SLES15 sp0
+  - 2.4.1
+
+  # Future versions
+  - ruby-head
+script:
+  - 'bundle exec rspec spec'
+  - 'bundle exec rubocop'
```
thutterer: Does Travis stop and report after it hit the first error? If yes, I would move rubocop one line up (before rspec) as rubocop usually fails faster than a large number of rspec tests.
Could save us some time.

ikapelyukhin: Maybe rubocop should be a separate Travis task?

ikapelyukhin: It seems like Travis runs rubocop even if rspec failed: https://travis-ci.org/SUSE/rmt/jobs/258048380

thutterer: Hm okay then it doesn't really matter.

PR #17, opened 2017-08-11
Add git to Dockerfile

ikapelyukhin:
I've forked `vcr` gem, it fails to install in docker image without git.


sandrew:
I believe you need only `git-core`

tmuntan1:
@sandrew Sorry, I loaded the pull request before you added your message. Looking into that now as well.

tmuntan1:
@sandrew You were right, it also works with `git-core` instead.

ikapelyukhin:
@sandrew @tmuntan1 :+1:

PR #19, opened 2017-08-15
Use Thor for a better CLI

thutterer:
I tried to keep the amount of changes as small as possible.
But I think I covered all functionality. 
Will have a second look tomorrow. Also need to think about which commands need a verbose-option, or if all should have one (like the commented line would do).
Have not yet touched the specs :grin:


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'pg', '0.18.3'
 
 gem 'nokogiri', '~> 1.8.0'
 gem 'rails', '~> 5.1'
+gem 'thor'
```
ikapelyukhin: Let's specify the version?

lib/rmt/repo_manager.rb
```diff
@@ -1,115 +1,81 @@
...
-    conditions[:mirroring_enabled] = true unless (@options[:all])
+    conditions[:mirroring_enabled] = true unless (scope == :all)
 
     rows = []
     repositories = Repository.where(conditions)
     repositories.all.each do |repository|
       rows << [ repository.id, repository.name, repository.distro_target, repository.description, repository.mirroring_enabled ]
     end
 
-    @out.puts Terminal::Table.new headings: %w[ID Name Target Description Mirror?], rows: rows
```
ikapelyukhin: I'm not sure if using `@out = out || StringIO.new` for output is a good idea or not.
Rubocop's complaining about using `puts`, also might come handy for capturing output in tests.

ikapelyukhin: Or  even rather replace `puts` it with logger.

ikapelyukhin:
So the output with no params currently looks this:

```
Commands:
  rmt-repos disable TARGET  # Disable a repository or product
  rmt-repos enable TARGET   # Enable a repository or product
  rmt-repos help [COMMAND]  # Describe available commands or one specific command
  rmt-repos list            # List all repositories
  rmt-repos version         # Show version
```

Let's maybe order it more logically:
```
list
enable
disable

version
help
```

ikapelyukhin:
Also it doesn't seem obvious that a command (like `list`) might have options.


Rails doesn't list `help` as a command, but prints the following line at the bottom of the output:
```
All commands can be run with -h (or --help) for more information.
```

Docker also doesn't have `help` listed among the commands, but has the following message:
```
Run 'docker COMMAND --help' for more information on a command.
```

ikapelyukhin:
We could also build a huge monolithic `rmt` binary, that does everything (mirroring, manage repos, etc.) via commands. 😕

lib/rmt/repo_manager.rb
```diff
@@ -1,115 +1,81 @@
 require 'rmt'
 require 'optparse'
```
ikapelyukhin: If we are using `thor`, we don't need `optparse` anymore?

Gemfile
thutterer: > Let's specify the version?

Not sure what the best practice here is. Because Thor is already a required gem for railties. Which probably already specifies the version somehow. 
So not really sure what is reasonable to set here.

thutterer:
> We could also build a huge monolithic rmt binary, that does everything (mirroring, manage repos, etc.) via commands.

Yes, maybe and no ...

__Yes__, that might be a way to go and Thor supports that. See section "Subcommands" at http://whatisthor.com/ Let's keep it simple for now, but consider that when we revisit the CLI in that special Trello card.

__Maybe__ this would become a "does way too much stuff" situation, if we don't do it right. So we should not do it _just because_ we can.

__No__, because there is already a `rmt` binary packaged:
`Program 'rmt' is present in package 'star', which is installed on your system.` :disappointed:

lib/rmt/repo_manager.rb
thutterer: > Rubocop's complaining about using `puts`

Yes, but is the Rails/Output cop complaining. Does not really apply for a CLI tool. I think I'll just disable this cop for these files :relaxed: 

> also might come handy for capturing output in tests

AFAIR you can capture STDOUT with RSpec for a while now. See https://stackoverflow.com/a/16507680/3449673

> Or even rather replace puts it with logger.

Somethink like [this](https://stackoverflow.com/a/6407200/3449673) could do.
I'll give it a try.

bin/rmt-mirror
```diff
@@ -6,16 +6,7 @@ $LOAD_PATH.unshift File.expand_path(File.join(__dir__, '../lib/'))
...
-
-require_relative '../app/models/application_record.rb'
-
-Dir[File.join(__dir__, '..', 'app', 'models', '**', '*.rb')].each { |file| require File.expand_path(file) }
-
-db_config_path = File.join(__dir__, '../config/database.yml')
-db_config = YAML.safe_load(ERB.new(File.read(db_config_path)).result, [], [], true)
-
-ActiveRecord::Base.establish_connection(db_config['development'])
+require 'database'
```
ikapelyukhin: Good stuff! :+1:

lib/suse/connect/api.rb
```diff
@@ -1,11 +1,14 @@
 require 'rmt'
 require 'rmt/http_request'
 require 'json'
+require 'byebug'
```
ikapelyukhin: Hmm?

Gemfile
ikapelyukhin: Yep, OK! Its' version is locked by `railties`.

lib/rmt/scc_sync.rb
```diff
@@ -29,8 +43,18 @@ def sync
     end
 
     @logger.info('Done!')
+  rescue SUSE::Connect::Api::InvalidCredentialsError
+    @logger.error('SCC credentials not valid.')
+  rescue Interrupt
+    puts 'Aborting! You need to rerun this command to have a consistent state.'
```
ikapelyukhin: Maybe this should also be logged? (You can send SIGINT to a background process, even though it is unlikely)

lib/rmt/repo_manager.rb
```diff
@@ -1,115 +1,80 @@
...
+require 'rmt/cli'
 
-class RMT::RepoManager
+# rubocop:disable Rails/Output
 
-  def initialize(argv, out = nil)
-    @argv = argv
-    @options = {}
-    @affected_repos = 0
-    @out = out || StringIO.new
```
ikapelyukhin: Make `StringIO` great again!

lib/rmt/scc_sync.rb
```diff
@@ -1,13 +1,27 @@
...
+    @logger.level = options[:verbose] ? 0 : 1
   end
 
+  class_option :verbose, aliases: '-v', type: :boolean
+
+  desc 'sync', 'Synchronize database with SCC'
   def sync
+    unless (Settings.scc.username && Settings.scc.password)
+      puts 'SCC credentials not set.'
+      exit 1
```
ikapelyukhin: Maybe let's subclass an exception, raise it here, catch it in `bin/rmt-repos` and `exit` there?
There's already similar code in `rmt-mirror`:

```ruby
  rescue RMT::Mirror::Exception => e
    warn e.to_s
    exit 1
  end
```

thutterer: Updated! To review your code example: I think it just say `abort e.to_s` :wink:

PR #22, opened 2017-08-18
Implement rmt-products list command

funglaub:
Not sure about requiring 'erb'...

### DIFF
--- a/lib/database.rb
+++ b/lib/database.rb
@@ -1,4 +1,5 @@
 require 'active_record'
+require 'erb'
 
 require_relative '../app/models/application_record.rb'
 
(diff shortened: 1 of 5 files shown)


ikapelyukhin:
@funglaub It doesn't look like you are rendering any templates in this PR, I don't see why `erb` would be required.
Also some Rubocop checks are failing, but those should be fixable using Rubocop's auto-correction (`rubocop -a`).

funglaub:
Sorry about that, I will have a look at that in the evening :)

ikapelyukhin:
@funglaub Sure, no problem. :+1:

lib/database.rb
```diff
@@ -1,4 +1,5 @@
 require 'active_record'
+require 'erb'
```
ikapelyukhin: OK, I've taken another look -- requiring `erb` here is actually fine, there's some rendering happening in database.yml.
(In other `bin` scripts `erb` gets pulled in by `rmt/config`)

PR #23, opened 2017-08-21
Rmt products command

funglaub:
Since I accidentally deleted the source branch, let's give this another try!

See also https://github.com/SUSE/rmt/pull/22


funglaub:
Thanks, appreciated! :)

PR #35, opened 2017-09-25
Fixes installation on darwin

kalabiyau:
see https://github.com/puma/puma/issues/1380


kalabiyau:
@ikapelyukhin I think we could target that major version for SLES15

PR #36, opened 2017-10-04
Remove SUSE repos without auth_tokens after sync 4582

ikapelyukhin:
Also splitted RMT::SCC and RMT::CLI::SCC specs into separate files.


suse-tests-pass:
@lagartoflojo will review your pull request :dancers: check https://trello.com/c/2F4BcmSK/4582-research-products-present-in-scoped-api-call-which-have-no-token-in-the-repos-call

spec/lib/rmt/scc_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+            expect(db_subscription[key]).to eq(subscription[key.to_sym])
+          end
+
+          expect(db_subscription[:kind]).to eq(subscription[:type])
+          expect(Subscription.statuses[db_subscription.status]).to eq(subscription[:status])
+        end
+      end
+    end
+
+    context 'removes SUSE repositories without auth tokens' do
```
lagartoflojo: I think this should be `describe` instead of `context`, since it's describing a special functionality rather than being a different context.

ikapelyukhin: I usually do `describe #method`, and everything that's inside is a `context`. Currently it's in `describe #sync` block.
How about `describe #remove_suse_repos_without_tokens`?

lagartoflojo: `describe #remove_suse_repos_without_tokens` that would work :)

ikapelyukhin: Okey-doke, done.

PR #39, opened 2017-10-18
Add a new subscription type to enum

ikapelyukhin:


ikapelyukhin:
@tmuntan1 I've pushed file permissions fix into this PR as well.

PR #41, opened 2017-10-23
Conditional RPM spec Ruby version

ikapelyukhin:


ikapelyukhin:
Downgraded Ruby version to 2.4, because it seems like rubocop doesn't have a parser for 2.5 yet and dies with:
```
Error: Unknown Ruby version 2.5 found in `.ruby-version`.
Known versions: 1.9, 2.0, 2.1, 2.2, 2.3, 2.4
```

PR #44, opened 2017-10-26
Properly handle licence mirroring errors

ikapelyukhin:


lib/rmt/cli/mirror.rb
```diff
@@ -19,6 +21,7 @@ def self.mirror(repository_url = nil, local_path = nil)
 
     repositories.each do |repository|
       begin
+        puts "Mirroring repository #{repository.name}"
```
tmuntan1: very useful!

PR #51, opened 2017-11-08
Offline sync

thutterer:


lib/rmt/cli/scc.rb
```diff
@@ -1,13 +1,20 @@
 class RMT::CLI::SCC < RMT::CLI::Base
 
   desc 'sync', 'Synchronize database with SCC'
+  option :to_dir, desc: 'Write info to a directory instead of database'
```
ikapelyukhin: How about `rmt-cli scc export`? It seems nicer than `--to-dir`.

lib/rmt/cli/mirror.rb
```diff
@@ -22,7 +24,9 @@ def self.mirror(repository_url = nil, local_path = nil)
     repositories.each do |repository|
       begin
         puts "Mirroring repository #{repository.name}"
-        mirror_one_repo(repository.external_url, repository.local_path, repository.auth_token)
+        repo_url = repository.external_url
+        repo_url.sub!(/.*(?=SUSE)/, "file://#{from_dir}/") if from_dir
```
ikapelyukhin: Consider looking at `Repository.make_local_path` rather than using a regexp, to have all the local path extraction logic in one place.

lib/rmt/cli/sync.rb
```diff
@@ -0,0 +1,20 @@
+class RMT::CLI::Sync < RMT::CLI::Subcommand
+  default_task :scc
+
+  desc 'scc', 'Sync database with SUSE Customer Center', hide: true
+  def scc
+    abort 'This RMT is in offline-mode. Use `sync airgap` if you want to sync from a portable storage.' if Settings.airgap.offline
+    RMT::SCC.new(options).sync
+  end
+
+  desc 'airgap', 'Store or read data at Airgap storage for offline usage'
```
ikapelyukhin: `Airgap storage` sounds like it some kind of special storage device. Air gap is a type of network architecture, not a type of a storage device.

ikapelyukhin:
I'm not sure if setting up things in the config file is a good idea -- it makes it harder to setup and troubleshoot (i.e., the user will do this stuff manually once a month, mount the storage to a different mount point and forget where the settings are).

I think it makes things easier if you have everything in the command line arguments, compared to have something hidden from you in a config file.

Also, right now the airgap-related stuff is all over the place:

* `rmt-cli sync airgap`
* `rmt-cli repos dumpdb`
* `rmt-cli mirror airgap`

Which is not very cool, considering it's a rather rare use-case. I'd prefer to isolate everything related to import/export in a more compact fashion and without the use of any settings in the config file, i.e.:

1. On the offline client:
  * `rmt-cli export settings /mnt/storage` -- to write repo settings into a file;

2. On the online client:
  * `rmt-cli export scc-data /mnt/storage` -- to dump SCC data to JSON files;
  * `rmt-cli export repos /mnt/storage` -- to mirror the repos, the settings file is in the same place;

3. Back on the offline client:
  * `rmt-cli import scc-data /mnt/storage` -- to import SCC data;
  * `rmt-cli import repos /mnt/storage` -- to import the RPMs.

Gemfile
```diff
@@ -15,6 +15,7 @@ gem 'mysql2', '~> 0.4.9'
 
 gem 'nokogiri', '~> 1.8.0'
 gem 'thor'
+gem 'rsync'
```
ikapelyukhin: `rsync` shouldn't stay. Fare thee well, `rsync`. :grimacing:

lib/rmt/downloader.rb
```diff
@@ -85,10 +85,14 @@ def verify_checksum(filename, checksum_type, checksum_value)
 
   def make_request(remote_file, local_file, request_fiber, checksum_type = nil, checksum_value = nil)
     uri = URI.join(@repository_url, remote_file)
-    uri.query = @auth_token if @auth_token
+    uri.query = @auth_token if (@auth_token && uri.scheme != 'file')
```
lagartoflojo: no parens necessary

ikapelyukhin: @lagartoflojo Rubocop doesn't complain though!
Everything's nicer with parens. And semicolons :D

thutterer:
Thanks for the reviews! I'll give Ivan's idea with the `import`/`export` sub-commands a shot!

thutterer:
> `rmt-cli export settings /mnt/storage` -- to write repo settings into a file;

@ikapelyukhin This format would break `rmt-cli export settings --help`, same as with the custom repos. What would work is `rmt-cli export settings --path=/mnt/storage` but that is rather verbose. 
What would you prefer? Not supporting `--help` or the longer `--path`?

spec/lib/rmt/cli/main_spec.rb
```diff
@@ -115,7 +101,7 @@
           end
         end
 
-        context 'Mysql2::Error with cannot connect error' do
```
ikapelyukhin: What's wrong with contexts?

thutterer: Well, one is a context, the other describes a thing :nerd_face:  Technically they are almost always interchangeable but readability suffers if you use them wrong.

ikapelyukhin: Oh, OK. It actually doesn't make sense when you try and read it with contexts. :+1:

lib/rmt/cli/export.rb
```diff
@@ -0,0 +1,25 @@
+# rubocop:disable Rails/Output
+
+class RMT::CLI::Export < RMT::CLI::Base
+
+  desc 'settings PATH', 'Store repository settings at given path'
+  def settings(path)
+    File.write(File.join(path, 'repos.json'), Repository.only_mirrored.pluck(:id).to_json)
```
ikapelyukhin: Hmm, yeah, for custom repos we'd need a bit more than just the ids.

lib/rmt/cli/main.rb
```diff
@@ -1,36 +1,34 @@
...
-# rubocop:disable Rails/Output
-
 class RMT::CLI::Main < RMT::CLI::Base
 
   class_option :version, desc: 'Show RMT version', type: :boolean, aliases: '-v', required: false
   class_option :debug, desc: 'Enable debug output', type: :boolean, aliases: '-d', required: false
 
+  desc 'sync', 'Sync database with SUSE Customer Center'
+  def sync
+    RMT::SCC.new(options).sync
```
ikapelyukhin: 1. That has that same issue with first level commands and `--help` flag.
2. `sync` is not very descriptive (mirroring is also a kind of sync).

lib/rmt/cli/mirror.rb
```diff
@@ -1,46 +1,55 @@
...
+    RMT::CLI::Base.handle_exceptions { mirror }
+  end
 
-    if repository_url
-      local_path ||= Repository.make_local_path(repository_url)
-      mirror_one_repo(repository_url, local_path)
-      return
-    end
+  desc 'custom URL', 'Mirror a custom repository URL'
+  def custom(url, path = nil)
```
ikapelyukhin: :+1:

package/rmt.conf
```diff
@@ -29,3 +29,7 @@ http_client:
   proxy_auth:
   proxy_user:
   proxy_password:
+
+air_gap:
+  offline: false
```
ikapelyukhin: Those are leftovers, then?

spec/lib/rmt/cli/repos_spec.rb
```diff
@@ -15,9 +15,9 @@
     end
 
     context 'without parameters' do
-      let(:argv) { [] }
+      let(:argv) { ['enable'] }
```
ikapelyukhin: The context is called `without parameters`, and now there's a parameter `enable` :grimacing:

spec/lib/rmt/cli/repos_spec.rb
```diff
@@ -59,9 +59,9 @@
     end
 
     context 'without parameters' do
-      let(:argv) { [] }
+      let(:argv) { ['disable'] }
```
ikapelyukhin: Same issue with the context name here.

lib/rmt/cli/export.rb
thutterer: Well, but then we should have worked on the "custom repos" card first to define how their information is stored. For the current feature set of RMT, this repos.json format it all it needs. Let's extend it when we get there. I don't want to start another card before finally finishing this one.

lib/rmt/cli/main.rb
thutterer: `scc sync` is also not much more descriptive. And it is as easy to confuse with `mirror` than just `sync` (imo), so I don't really care anymore. I can revert if you prefer that.

@kalabiyau  We need some real users (and some planned time for sitting with them) to give us some better understanding on how clear/confusing it is. Before that, we can move commands around all we want and it won't get much better :expressionless:

package/rmt.conf
thutterer: Uh :hankey: , yes, will remove.

spec/lib/rmt/cli/repos_spec.rb
thutterer: Good point :smile: I will have a look.

lib/rmt/cli/main.rb
ikapelyukhin: @thutterer Well, `scc sync` doesn't have that `--help` issue. Otherwise I don't care much about this as well :D

lib/rmt/cli/export.rb
ikapelyukhin: Yep! Just saying that the format is likely to change soon, which might be an issue if SUMA would want to use it.

spec/lib/rmt/cli/repos_spec.rb
thutterer: Well, look again: I added what the test was supposed to do, but did not. It is about testing `rmt-cli repos enable` without any parameters (like an Id)... before my changes, all the test did was calling `rmt-cli repos` ... which was not was the test was all about.

spec/lib/rmt/cli/repos_spec.rb
thutterer: Same as above.

lib/rmt/cli/export.rb
thutterer: Agreed, I would suggest to not call this feature "done done" in the sense of "being ready to be picked up by SUMA". Let's do that when we have the custom repos feature defined and implemented. Okay with you, @kalabiyau ?

lib/rmt/cli/main.rb
thutterer: Neither does `version --help` work :wink: But it really puzzles me why Thor does not support that (but it supports the same thing with subcommands, e.g. `rmt-cli repos enable --help` does work) ... so I'm wondering if we should hack Thor to support it.

For now, I'd like to leave it as it is and get some user insights before we invest any more time into reshuffling the commands layout.

ikapelyukhin:
When running specs it's flooding the output with the following message (on both my machine and on Travis):

```
[WARNING] Attempted to create command "__help_without_any_instance__" without usage or description. Call desc if you want this method to be available as command or declare it inside a no_commands{} block. Invoked from "/home/ivan/.rvm/gems/ruby-2.4.1/gems/rspec-mocks-3.6.0/lib/rspec/mocks/any_instance/recorder.rb:234:in `alias_method'".
```

Could you please do what it wants and wrap whatever causing this in `no_commands{}` block?

spec/lib/rmt/scc_spec.rb
```diff
@@ -131,4 +141,66 @@ def scc
...
+        end
+      end
+    end
+  end
+
+  describe '#import' do
+    let(:path) { '/tmp/usb' }
+
+    context 'with bad path or missing files' do
+      it 'raises an error before it touches the database' do
```
ikapelyukhin: I've had `/tmp/usb` directory on my machine for some reason, and this test was failing.
It would be better if it was either wrapped into FakeFS block or used a random directory name.

thutterer:
> When running specs it's flooding the output with the following message (on both my machine and on Travis)

This seems to come from using `expect_any_instance_of` on the CLI classes... As the usage of "any_instance_of" is already a smell, I guess this needs some refactoring and not just usage of `no_commands` block... check the output: these are not actual methods that get called but the name of the test... kinda:
```
[WARNING] Attempted to create command "__help_without_any_instance__" without usage or description.
```

I'll see what I can do ...

spec/lib/rmt/scc_spec.rb
```diff
@@ -131,4 +141,69 @@ def scc
...
+          FileUtils.mkdir_p(path)
+          File.write(File.join(path, 'organizations_products.json'), products.to_json)
+          File.write(File.join(path, 'organizations_subscriptions.json'), subscriptions.to_json)
+          File.write(File.join(path, 'organizations_repositories.json'), all_repositories.to_json)
+
+          described_class.new.import(path)
+        end
+      end
+
+      include_examples 'saves in database'
```
ikapelyukhin: Those shared examples are somehow order dependent/have some context from the previous run.
If you run the whole spec file -- it works, however when running just the `decribe '#import'` block they fail.
Hmm, maybe we should enable random execution order for all the specs?

lib/rmt/cli/mirror.rb
```diff
@@ -1,46 +1,55 @@
...
+      if repos.empty?
+        warn 'There are no repositories marked for mirroring.'
+        return
+      end
+
+      repos.each do |repository|
+        begin
+          puts "Mirroring repository #{repository.name} from #{from_dir || 'SCC'} to #{base_dir}"
+          repo_url = repository.external_url
+          repo_url.sub!(/.*(?=SUSE)/, "file://#{from_dir}/") if from_dir
```
ikapelyukhin: Hmm, we have some repos like Nvidia and such, I'm not sure if the URL always contains `SUSE` string.

spec/lib/rmt/cli/export_spec.rb
```diff
@@ -59,18 +61,31 @@
...
       end
+
+      context 'with exceptions during mirroring' do
+        it 'outputs exception message' do
+          FakeFS.with_fresh do
+            FileUtils.mkdir_p path
+            File.write("#{path}/repos.json", repo_ids.to_json)
+
+            expect(mirror_double).to receive(:mirror)
+            expect(mirror_double).to receive(:mirror).and_raise(RMT::Mirror::Exception, 'black mirror')
```
ikapelyukhin: `black mirror` :scream:

lib/rmt/cli/import.rb
```diff
@@ -17,7 +17,8 @@ def repos(path)
       end
 
       repos.each do |repo|
-        repo.external_url.sub!(/.*(?=SUSE)/, "file://#{path}/") # FIXME: this probably won't work for some repos
+        require 'byebug'; byebug
```
ikapelyukhin: Wat?! :scream:

thutterer:
@ikapelyukhin About this one FIXME you added:
We could replace this line with:
```ruby
repo.external_url = "file://#{path}#{URI(repo.external_url).path}"
```

But, we cannot, because that conflicts with the URL mangling you did [there](https://github.com/SUSE/rmt/blob/master/app/models/repository.rb#L19) ... So we either have to repeat this logic here, or we get rid of it. Not sure if that was okay, so I leave it up to you to do the one or the other.

Also, I created #55 which I found today, but I would vote for doing this in a separate PR later.
Same for the other FIXME I added in 12f1fd8.

Other than that, I think this is ready to be reviewed and tested. Could you please address the issue above and then ask @digitaltom to review when you are done?

One hint for the reviewer:
When implementing and testing, I always used an absolute path in all the commands (e.g. `rmt-cli export repos /mnt/usb` but I just realized that I don't enforce this. So a user could do `rmt-cli export repos ../foo` or whatever. I did not test this and have no idea if or where it fails. Please give this some extra attention and form an opinion if this is something we want to support or not. Personally I think it would not be worth spending too much extra time on this, as usually people have no reason not to provide the full path to the mounted storage device.

ikapelyukhin:
@thutterer OK, I'll look into path mangling.

lib/rmt/cli/products.rb
```diff
@@ -1,32 +1,37 @@
...
-      warn 'No products found in the DB. Run "rmt-cli scc sync" to synchronize with your SUSE Customer Center data first.'
+    puts Terminal::Table.new(headings: headings, rows: rows) unless rows.empty?
+
+    if options.all
+      warn 'Run "rmt-cli sync" to synchronize with your SUSE Customer Center data first.' if rows.empty?
     else
-      puts Terminal::Table.new headings: headings, rows: rows
+      warn 'No matching products found in the database.' if rows.empty?
+      puts 'Only enabled products are shown by default. Use the `--all` option to see all products.'
     end
```
lagartoflojo: `rows.empty?` gets called in 2 `if`s and 1`unless`...
How about...

```ruby
    if rows.empty?
      if options.all
        warn 'Run "rmt-cli sync" to synchronize with your SUSE Customer Center data first.'
      else
        warn 'No matching products found in the database.'
      end
    else
      puts Terminal::Table.new(headings: headings, rows: rows)
      puts 'Only enabled products are shown by default. Use the `--all` option to see all products.' unless options.all
    end
```

lib/rmt/downloader.rb
```diff
@@ -85,36 +85,26 @@ def verify_checksum(filename, checksum_type, checksum_value)
...
+
+    if (URI(uri).scheme == 'file' && !File.exist?(uri.path))
```
lagartoflojo: no parens necessary

lib/rmt/downloader.rb
```diff
@@ -85,36 +85,26 @@ def verify_checksum(filename, checksum_type, checksum_value)
...
 
     begin
-      response = request.receive_body
-      if (response.return_code && response.return_code != :ok)
-        raise RMT::Downloader::Exception.new("#{remote_file} - return code #{response.return_code}")
-      end
-
-      downloaded_file.close
-
+      request.receive_body
```
lagartoflojo: Why let `#receive_headers` handle its exceptions but not `#receive_body`?

lib/rmt/downloader.rb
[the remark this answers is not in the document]
ikapelyukhin: Rubocop doesn't complain!!! :scream:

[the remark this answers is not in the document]
lagartoflojo: I think it's the third time you say that, but... so what? Just because rubocop doesn't complain doesn't mean it doesn't look ugly...

README.md
```diff
@@ -51,11 +51,38 @@ desired location.
...
+#### Inital Setup
+
+##### On the Online RMT
+
+- `rmt-cli export scc-data /mnt/usb` will get the required JSON responses from SCC and save them as files at the specified path.
+
+##### On the Offline RMT
+
+- `rmt-cli import scc-data /mnt/usb` will read the JSON-files from given path and fill the local database.
+- Now use `rmt-cli repos enable` to mark repos for mirroring.
```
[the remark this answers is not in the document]
thutterer: I think I had it in an earlier version and then trimmed the text down a bit. But yes, I can add it again.

lib/rmt/cli/export.rb
```diff
@@ -0,0 +1,37 @@
...
+  Run this command on an online RMT.
+  It will look in PATH for a repos.json file which has to contain a list of repository IDs.
+  Usually, this file gets created by an offline RMT with `export settings`.
+
+  `export repos` will mirror these repositories to this PATH, usually a portable storage device.
+  REPOS
+  def repos(path)
+    needs_path(path) do
+      repos_file = File.join(path, 'repos.json')
+      repos_ids = JSON.parse(File.read(repos_file))
```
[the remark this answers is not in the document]
thutterer: Yes, we definitely should. I'll add a TODO now and probably fix it here while you review the rest of the PR :)

lib/rmt/cli/import.rb
```diff
@@ -0,0 +1,28 @@
...
+      if repos.empty?
+        warn 'There are no repositories marked for mirroring.'
+        return
+      end
+
+      repos.each do |repo|
+        external_url = repo.external_url
+        repo.external_url = 'file://' + path + Repository.make_local_path(repo.external_url)
+        mirror(repo)
+        repo.update(external_url: external_url) # make sure we do not permanently change the URI
```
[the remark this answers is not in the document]
thutterer: It is! :ghost: And I don't like it myself, at all! I'll mark this comment with a fat `TODO` and see if I can workaround with `update_attribute`.

The background:
The repo resource gets updated later during mirroring to store the new timestamp. Other changes to the repo get saved along side. This line was a "hotfix" for that, but I guess I can refactor the problematic model methods to only update only one column.

lib/rmt/downloader.rb
thutterer: I pass the ball to @ikapelyukhin here: :football:

lib/rmt/downloader.rb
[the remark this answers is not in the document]
ikapelyukhin: @lagartoflojo The whole point of using Rubocop is so that issues like that don't pop up during the review (because fixing some minor style issue manually every time is a huge annoyance and waste of time -- that should happen automagically) and also so that all the rules are uniform and codified (other reviewers don't complain about this particular issue).

So it either has to be a Rubocop rule, or you'll have to accept parens :-)

P.S. On a side note, many things are ugly. JavaScript is ugly, let's stop using JavaScript. :-)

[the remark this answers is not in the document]
thutterer: But Hernan does! :laughing:

lib/rmt/downloader.rb
ikapelyukhin: I was doing refactoring to fix offenses about method over-complexity. Right now it handles both `#receive_body` and `#verify_checksum` methods exceptions -- but I'll see if I can move `#verify_checksum` into `Request` class as well.

lib/rmt/downloader.rb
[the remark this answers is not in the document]
ikapelyukhin: Y'all need some Perl code :grimacing:

lib/rmt/downloader.rb
ikapelyukhin: Done!

lib/rmt/downloader.rb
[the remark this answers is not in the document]
thutterer: > So it either has to be a Rubocop rule, or you'll have to accept parens :-)

Hm, not sure if they even is such a rule, but if it was, I would vote against forcing one or the other option. Because sometimes it helps with readability and a pair of parens can save you, future-you or any other developer a few seconds of thinking. And sometimes the logic of an "complex" if-statement is so obvious that the parens itself become a distraction. How you perceive this, is different per developer and can always lead to discussions.
So I'm fine with or without ... In this case, _I_ don't think there is a need for them, so I removed them. Which I probably should not have done to let you guys fight it out :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
ikapelyukhin: @thutterer There's a rule that fixes simple issues like `puts 1 if (1)`.

But yeah -- sometimes you need the parens. That's what I'm saying -- if Rubocop doesn't complain, then it's not an issue.

P.S. Does that mean that I can demand that you add the parens when I'm doing the reviews? Because I like parens! :grimacing:

spec/lib/rmt/cli/main_spec.rb
```diff
@@ -5,106 +5,101 @@
...
 
-    context 'handles exceptions with --debug' do
-      let(:argv) { ['mirror', '--debug'] }
+    describe 'exception handling' do
+      let(:exception_class) { RMT::CLI::Error }
+      let(:error_message) { 'Dummy error' }
 
       before do
-        expect(RMT::CLI::Mirror).to receive(:mirror) { raise RMT::CLI::Error, 'Dummy exception' }
+        expect(described_class).to receive(:help) { raise exception_class, error_message }
```
lagartoflojo:
```ruby
expect(described_class).to receive(:help).and_raise(exception_class, error_message)
```

spec/lib/rmt/cli/mirror_spec.rb
```diff
@@ -8,7 +8,7 @@
       subject(:command) do
         repository
         expect_any_instance_of(RMT::Mirror).not_to receive(:mirror)
```
lagartoflojo: Expectations inside subject 😱

spec/lib/rmt/cli/mirror_spec.rb
```diff
@@ -18,69 +18,45 @@
...
-    end
 
-    context 'with exceptions during mirroring' do
-      subject(:command) do
-        repository
-        expect_any_instance_of(RMT::Mirror).to receive(:mirror).at_least(:once) do
-          raise RMT::Mirror::Exception, 'Test double exception'
+      context 'with exceptions during mirroring' do
+        subject(:command) do
+          expect_any_instance_of(RMT::Mirror).to receive(:mirror).and_raise(RMT::Mirror::Exception, 'black mirror')
```
lagartoflojo: I would move this line to the it block with an allow:
```ruby
allow_any_instance_of(RMT::Mirror).to receive(:mirror).and_raise(RMT::Mirror::Exception, 'black mirror')
```
We don't "expect" RMT::Mirror to raise the exception, what we care about is the output to STDERR.

lib/rmt/cli/products.rb
thutterer: Thanks for the suggestion. I know my version looked overly complicated.
But in yours, you missed the case where it outputs `Only enabled products are shown by default. Use the `--all` option ...` when you you have `rows.empy?` and not used `--all`. So I moved that line out of that if-block ... hope it is okay now.

spec/support/fakefs_autoload.rb
```diff
@@ -0,0 +1,20 @@
+# Make FakeFS play nicely with autoload
```
lagartoflojo: Is there a source for this file?

spec/lib/rmt/scc_spec.rb
lagartoflojo: Hmm not good. Is this still the case?

spec/lib/rmt/scc_spec.rb
```diff
@@ -1,81 +1,91 @@
...
-        end
-      end
-      let(:api_double) { double }
+describe RMT::SCC do
+  let!(:products) { JSON.parse(file_fixture('products/dummy_products.json').read, symbolize_names: true) }
+  let!(:subscriptions) { JSON.parse(file_fixture('subscriptions/dummy_subscriptions.json').read, symbolize_names: true) }
+  let(:extension) { product[:extensions][0] }
+  let(:all_repositories) do
+    products.flat_map do |product|
+      [product, product[:extensions][0]].flat_map { |item| item[:repositories] }
```
lagartoflojo: image

spec/lib/rmt/scc_spec.rb
```diff
@@ -1,81 +1,91 @@
 require 'rails_helper'
+require 'fakefs/spec_helpers'
```
lagartoflojo: Is this necessary? And if so, shouldn't this line go in `rails_helper`?

spec/lib/rmt/scc_spec.rb
```diff
@@ -1,81 +1,91 @@
...
+      before do
+        allow(Settings).to receive(:scc).and_return OpenStruct.new(username: 'foo', password: 'bar')
+        described_class.new.sync
+      end
+
+      include_examples 'saves in database'
+    end
   end
 
   describe '#remove_suse_repos_without_tokens' do
```
lagartoflojo: I guess the `expect`s in lines 120-123 should/could just be `allow`s?

spec/support/fakefs_autoload.rb
ikapelyukhin: What do you mean? I wrote it. I have found any existing solution for FakeFS to handle ActiveSupport autoload magic.

lagartoflojo: I thought it might have been copied from StackOverflow or something. If not, then nothing to do :)

lib/rmt/cli/import.rb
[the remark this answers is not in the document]
thutterer: @ikapelyukhin @lagartoflojo 
I checked and this could be avoided by changing [refresh_timestamp!](https://github.com/SUSE/rmt/blob/master/app/models/repository.rb#L23) to only:
```ruby
  update_column(:last_mirrored_at, DateTime.now.utc)
```

But while this fixes the current issue, it masquerades the underlying problem too much, imo.
What do you think? Should I change it to that?
Or better add a `TODO` comment here and leave it for now?

Because I don't want to start another round of changes in this PR.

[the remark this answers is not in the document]
ikapelyukhin: @thutterer Maybe remove `refresh_timestamp!` method call from `RMT::Mirror` class and call it manually from wherever it makes sense to be called?

spec/support/fakefs_autoload.rb
ikapelyukhin: We can put it on StackOverflow! :D

lib/rmt/cli/import.rb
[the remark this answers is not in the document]
lagartoflojo: Btw, you can just do: `touch(:last_mirrored_at)`

spec/lib/rmt/cli/export_spec.rb
```diff
@@ -89,9 +90,9 @@
             FileUtils.mkdir_p path
             File.write("#{path}/repos.json", repo_ids.to_json)
 
+            expect(RMT::Mirror).to receive(:from_repo_model).exactly(repo_ids.count).times.and_return(mirror_double)
             expect(mirror_double).to receive(:mirror)
             expect(mirror_double).to receive(:mirror).and_raise(RMT::Mirror::Exception, 'black mirror')
```
lagartoflojo: Still a duplicate `expect(mirror_double).to receive(:mirror)` (lines 94 & 95), is that intentional?

thutterer:
@ikapelyukhin I think e12b57c fixes the `uninitialized constant`. Outside of rspec this never was a problem. I guess, because we have ActiveSupport in the bin/rmt-cli already.

@lagartoflojo Good to merge?

PR #52, opened 2017-11-09
Get users information on how to provide feedback

kalabiyau:
I believe that feedback possibility should be that prominent that one cannot avoid it.
It is a small additional cli item which should give info that user can reach out to github issues to leave their feedback

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,4 @@ coverage/
 config/secrets.yml.key
 config/secrets.yml.enc
 config/rmt.local.yml
-
+public/repo
--- a/lib/rmt/cli/main.rb
+++ b/lib/rmt/cli/main.rb
@@ -33,6 +33,14 @@ class RMT::CLI::Main < RMT::CLI::Base
     puts RMT::VERSION
   end
 
+  desc 'feedback', 'How to provide feedback about RMT'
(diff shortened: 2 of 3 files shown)


kalabiyau:
@tmuntan1 proposed to put that statement in the default output of rmt-cli. So we have two options.

ikapelyukhin:
@kalabiyau How would we prefer to receive feedback? GitHub issues, mailing list?

kalabiyau:
@ikapelyukhin I would love to keep in github, as it is opensource project and being transparent is a big part of it

thutterer:
> @tmuntan1 proposed to put that statement in the default output of rmt-cli. So we have two options.

Good idea, you could just add it after https://github.com/SUSE/rmt/blob/master/lib/rmt/cli/base.rb#L18 and it would show with `rmt-cli` (without any command) and all help commands. (Expect where the command is not a subcommand group. Currently these are sync, version and help itself.)

But still, this is probably the most prominent place. Even more prominent than "hiding" the URL inside a feedback-command which the user first has to discover.

@kalabiyau Can you resolve the merge conflicts? I think it makes a lot of sense to merge this while we are still in an early stage :baby_symbol:

PR #53, opened 2017-11-17
Unbundle some of the dependencies

ikapelyukhin:


ikapelyukhin:
Those are really ugly hacks :cry:
They should be killed with :fire::fire::fire:

ikapelyukhin:
Let's close it for now.

PR #56, opened 2017-12-12
Deduplicate files by copy/hardlink during download 4701

tmuntan1:


lib/rmt/downloader.rb
```diff
@@ -25,12 +27,15 @@ def initialize(repository_url:, local_path:, auth_token: nil, logger: nil)
 
   def download(remote_file, checksum_type = nil, checksum_value = nil)
     local_file = make_local_path(remote_file)
+    was_deduplicated = deduplicate(checksum_type, checksum_value, local_file)
 
-    request_fiber = Fiber.new do
-      make_request(remote_file, local_file, request_fiber, checksum_type, checksum_value)
-    end
+    unless was_deduplicated
```
ikapelyukhin: Maybe just `return local_file if (was_deduplicated)`?

lib/rmt/downloader.rb
```diff
@@ -56,25 +61,30 @@ def process_queue
...
 
-    # The request is wrapped into a fiber for exception handling
-    request_fiber = Fiber.new do
-      begin
-        make_request(remote_file, local_file, request_fiber, queue_item[:checksum_type], queue_item[:checksum])
-      rescue RMT::Downloader::Exception => e
-        @logger.warn("× #{File.basename(local_file)} - #{e}")
-      ensure
-        process_queue
+    was_deduplicated = deduplicate(queue_item[:checksum_type], queue_item[:checksum], local_file)
```
ikapelyukhin: It probably makes sense to move that into a `# Skip over files that already exist` loop above.

lib/rmt/downloader.rb
```diff
@@ -56,25 +61,30 @@ def process_queue
...
 
-    @hydra.queue(request_fiber.resume)
   end
 
   def verify_checksum(filename, checksum_type, checksum_value)
     hash_function = checksum_type.gsub(/\W/, '').upcase.to_sym
     hash_function = :SHA1 if (hash_function == :SHA)
 
-    unless (KNOWN_HASH_FUNCTIONS.include? hash_function)
+    unless KNOWN_HASH_FUNCTIONS.include? hash_function
```
ikapelyukhin: Why you hate parens :cry:

spec/lib/rmt/downloader_spec.rb
```diff
@@ -6,6 +6,10 @@
   let(:downloader) { described_class.new(repository_url: 'http://example.com', local_path: dir) }
   let(:headers) { { 'User-Agent' => "RMT/#{RMT::VERSION}" } }
 
+  after do
+    FileUtils.remove_entry(dir)
```
ikapelyukhin: We've got FakeFS now in @thutterer's pull request, check it out -- I think it makes sense to use it everywhere.

lib/rmt/checksum_verifier.rb
```diff
@@ -0,0 +1,32 @@
+class RMT::ChecksumVerifier
+
+  KNOWN_HASH_FUNCTIONS = %i[MD5 SHA1 SHA256 SHA384 SHA512].freeze
+
+  class Exception < RuntimeError
+  end
+
+  ##
+  # This method verifies that files are what they say that they are.
```
ikapelyukhin: I'm not sure how helpful that description is (and if it is needed at all). The method name is `verify_checksum`, it would have been weird if it did anything else :grimacing:

lib/rmt/downloader.rb
tmuntan1: I love them in other languages, but not in ruby. ^_^

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,60 @@
...
+  # * The checksum of the source file is verified to prevent data integrity issues (corrupted files deduplicated).
+  # * If any error is encountered, this method will return nil and ignore that there was a previous source.
+  #
+  # Params:
+  # +checksum_type+:: The type of checksum (MD5, SHA256, ...)
+  def self.get_local_path_by_checksum(checksum_type, checksum)
+    file = DownloadedFile.find_by({ checksum_type: checksum_type,
+                                    checksum: checksum })
+    return nil unless file
+    RMT::ChecksumVerifier.verify_checksum(checksum_type, checksum, file.local_path)
```
ikapelyukhin: That might be a bit to CPU aggressive (i.e., potentially calculating checksums on half a terabyte every night).
I think it would be sufficient to check that file exists and has the correct size (the size is available in the metadata and can also be stored).
For verifying data integrity it would be good to have a separate command (implemented as a separate feature) that you'd run manually if something's actually broken.

suse-tests-pass:
@jezuhke will review your pull request :dancers: check https://trello.com/c/gUwi6Per/4701-5-rmt-files-de-duplication
  @jezuhke: Please review this pull request using our guidelines:

suse-tests-pass:
@lagartoflojo will review your pull request :dancers: check https://trello.com/c/gUwi6Per/4701-5-rmt-files-de-duplication

app/models/downloaded_file.rb
tmuntan1: This should only be called if the file doesn't exist and a duplicate was found. After the first run, I don't think that it would be too much of a CPU issue, but it would be a better idea to add the filesize anyways.

ikapelyukhin: I was under assumption that it is always called, but I didn't actually run the code yet :grimacing:

tmuntan1: We now check by the filesize in my latest commit.

spec/lib/rmt/downloader_spec.rb
tmuntan1: As I ran into lots of small issues while trying this, I don't think that it is worth reworking all of these functioning tests right now.

This would require rewriting specs that deduplication doesn't touch and I feel that this is outside the scope here. I ran into full btrf filesystem issues, so I cleaned up after ourselves for the previous and new tests.

ikapelyukhin: OK, maybe I'll refactor it all myself later to use FakeFS.

spec/support/fixtures.rb
```diff
@@ -0,0 +1,7 @@
+def fixture_file(relative_path)
+  File.read(fixture_file_path(relative_path))
```
ikapelyukhin: There's already a default RSpec method called `file_fixture`.
You could do something like `file_fixture('products/dummy_products.json').read`.

spec/support/fixtures.rb
```diff
@@ -0,0 +1,7 @@
...
+end
+
+def fixture_file_path(relative_path)
+  ::Rails.root.join('spec', 'fixtures', 'files', relative_path).to_s
```
ikapelyukhin: Same -- `file_fixture` returns the path.

spec/lib/rmt/config_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    context 'handles empty string' do
+      before { deduplication_method('') }
+      it('defaults to hardlink') { expect(described_class.deduplication_by_hardlink?).to be_truthy }
+    end
+
+    context 'handles everything' do
+      before { deduplication_method(Class.new) }
+      it('defaults to hardlink') { expect(described_class.deduplication_by_hardlink?).to be_truthy }
```
ikapelyukhin: But is it really supposed to handle "everything"?

`railsconfig` gem actually supports [config schema validation](https://github.com/railsconfig/config#validation), so rather then testing weird things like `Class.new`, maybe let's add validation instead. Because `Class.new` is not a valid input.

spec/lib/rmt/downloader_spec.rb
```diff
@@ -101,6 +105,99 @@
...
+          .with(headers: headers)
+          .to_return(status: 200, body: rpm_content, headers: {})
+        stub_request(:get, 'http://example.com/repo1/foo.drpm')
+          .with(headers: headers)
+          .to_return(status: 200, body: drpm_content, headers: {})
+      end
+
+      context 'handles duplicated files by copy' do
+        before do
+          deduplication_method(:copy)
```
ikapelyukhin: I would maybe add `.to have_been_made.once` to make sure it makes only one HTTP request, but that's not too critical.

spec/lib/rmt/checksum_verifier_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+    context 'checksum is correct' do
+      it 'does not raises a exception' do
+        expect do
+          described_class.verify_checksum('MD5', '3858f62230ac3c915f300c664312c63f', test_file_path)
+        end.not_to raise_error
+      end
+
+      it 'handles strange formatting of MD5' do
+        expect do
+          described_class.verify_checksum('Md5', '3858f62230ac3c915f300c664312c63f', test_file_path)
```
ikapelyukhin: I would reduce the amount of examples by testing only `wEiRdLy` formatted/cased strings -- everything is upcased/downcased in the method anyway.
Also, hash value also contains case-sensitive characters.
Both aren't critical though.

spec/lib/rmt/config_spec.rb
tmuntan1: everything is an object in ruby

ikapelyukhin: Sure, but we expect only very specific object with only two very specific values. :-)
Everything else is invalid.

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,30 @@
+require 'rmt/checksum_verifier'
+
+class DownloadedFile < ApplicationRecord
+
+  def self.add_file!(checksum_type, checksum, local_path)
+    return nil unless local_path.match?(/\.(rpm|drpm)$/)
```
ikapelyukhin: That might not be the best logic:

In `RMT::Mirror` there are `mirror_metadata` and `mirror_license` methods, which download all of the metadata. In those methods all the files are known to be metadata and those files we don't want to deduplicate.

`mirror_data` method parses `primary.xml` file of the repo and downloads files that are mentioned there. However, it can reference files with arbitrary names (and extensions). While most commonly in the modern SUSE repos the extensions are `.rpm` and `.drpm`, that doesn't necessarily apply for older or third-party repos.

I would suggest moving all of the deduplication method calls out of `RMT::Downloader` class into `mirror_data` method `RMT::Mirror` class.

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,30 @@
...
+    file_size = File.size(local_path)
+    DownloadedFile.create({ checksum_type: checksum_type,
+                            checksum: checksum,
+                            local_path: local_path,
+                            file_size: file_size })
+  rescue StandardError => e
+    # de-duplication is not critical, just move on
+    logger.error e.message
+    logger.error e.backtrace.each { |line| logger.error line }
```
ikapelyukhin: So far everywhere else in the code we are not showing stack traces unless debug option is specified.

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def self.get_local_path_by_checksum(checksum_type, checksum)
+    file = DownloadedFile.find_by({ checksum_type: checksum_type,
+                                    checksum: checksum })
+    return nil if file.nil? || file.file_size != File.size(file.local_path)
+    file.local_path
+  rescue StandardError => e
+    # de-duplication is not critical, just return nil to download again
+    logger.error e.message
+    logger.error e.backtrace.each { |line| logger.error line }
```
ikapelyukhin: Same issue as above.

ikapelyukhin: Also, `logger` is not defined here -- when CLI app is running, Rails is not loaded.
When running `rmt-cli mirror`, I'm getting:
```
rmt/app/models/downloaded_file.rb:25:in `rescue in get_local_path_by_checksum': undefined method `error' for nil:NilClass (NoMethodError)
	from rmt/app/models/downloaded_file.rb:19:in `get_local_path_by_checksum'
	from rmt/lib/rmt/file_utils.rb:6:in `deduplicate'
```

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,30 @@
...
+  rescue StandardError => e
+    # de-duplication is not critical, just move on
+    logger.error e.message
+    logger.error e.backtrace.each { |line| logger.error line }
+  end
+
+  def self.get_local_path_by_checksum(checksum_type, checksum)
+    file = DownloadedFile.find_by({ checksum_type: checksum_type,
+                                    checksum: checksum })
+    return nil if file.nil? || file.file_size != File.size(file.local_path)
```
ikapelyukhin: Testing on FAT filesystem with hardlinks, here I'm getting:

```
No such file or directory @ rb_file_s_size - /home/ivan/work/rmt/public/repo/SUSE/Updates/SLE-SERVER-INSTALLER/12-SP2/x86_64/update/x86_64/hwinfo-21.31_21.38-11.17_15.7.2.x86_64.drpm
/home/ivan/work/rmt/app/models/downloaded_file.rb:21:in `size'
/home/ivan/work/rmt/app/models/downloaded_file.rb:21:in `get_local_path_by_checksum'
```

It might make sense to chech that `file.local_path` exists first.

ikapelyukhin: The exception occurs when the file record exists in the DB, but the file has been deleted.

spec/support/fixtures.rb
tmuntan1: good catch, updated

lib/rmt/file_utils.rb
```diff
@@ -0,0 +1,18 @@
+require 'fileutils'
+
+class RMT::FileUtils
+
+  def self.deduplicate(checksum_type, checksum_value, destination)
+    local_path = DownloadedFile.get_local_path_by_checksum(checksum_type, checksum_value)
+    return false unless local_path
+
+    if RMT::Config.deduplication_by_hardlink?
+      ::FileUtils.ln(local_path, destination)
```
ikapelyukhin: If the FS doesn't support hardlinks, this would die with:
```
.rvm/rubies/ruby-2.4.1/lib64/ruby/2.4.0/fileutils.rb:287:in `link': Operation not permitted @ rb_file_s_link - (rmt/public/repo/SUSE/Updates/SLE-SERVER-INSTALLER/12-SP2/x86_64/update/x86_64/hwinfo-21.23_21.38-1.13_15.7.2.x86_64.drpm, rmt/public/repo/SUSE/Updates/SLE-DESKTOP-INSTALLER/12-SP2/x86_64/update/x86_64/hwinfo-21.23_21.38-1.13_15.7.2.x86_64.drpm) (Errno::EPERM)
	from .rvm/rubies/ruby-2.4.1/lib64/ruby/2.4.0/fileutils.rb:287:in `block in ln'
	from .rvm/rubies/ruby-2.4.1/lib64/ruby/2.4.0/fileutils.rb:1476:in `fu_each_src_dest0'
```
It's actually OK that it dies, but it would be nice to have a better error message that says something along the lines of `Your filesystem doesn't allow hardlinks, try using copy instead`.

lib/rmt/downloader.rb
```diff
@@ -68,7 +75,14 @@ def process_queue
     @hydra.queue(request_fiber.resume)
   end
 
+  def deduplicate(checksum_type, checksum_value, destination)
+    return false unless RMT::FileUtils.deduplicate(checksum_type, checksum_value, destination)
+    @logger.info("↓ #{File.basename(destination)}")
```
ikapelyukhin: A different symbol would be good. `→` maybe?
SMT has `D` for download and I think `L` for link.

app/models/downloaded_file.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.add_file!(checksum_type, checksum, local_path)
+    return nil unless local_path.match?(/\.(rpm|drpm)$/)
+    file_size = File.size(local_path)
+    DownloadedFile.create({ checksum_type: checksum_type,
+                            checksum: checksum,
+                            local_path: local_path,
+                            file_size: file_size })
+  rescue StandardError => e
+    # de-duplication is not critical, just move on
+    ::Rails.logger.debug e.message
```
ikapelyukhin: That still doesn't work, because Rails are still not loaded. :confused:

spec/lib/rmt/downloader_spec.rb
tmuntan1: This was not worthwhile in the end, wontfix.

PR #60, opened 2017-12-19
Clean up the mess that the specs output into STDOUT/STDERR

ikapelyukhin:
Since the last merge running specs on master they've started to produce some extra output to the console, which is not nice :hankey:


ikapelyukhin:
@thutterer Wanna finish those pending specs too, maybe? :grimacing:

ikapelyukhin:
Added an experimental spec helper to prevent messy STDOUT/STDERR. Any thoughts on it? :grimacing:

thutterer:
> What about those pending specs? Or should we wait until custom repos are implemented?

[Yes!](https://github.com/SUSE/rmt/blob/clean_up_output_mess_in_specs/spec/lib/rmt/cli/mirror_spec.rb#L45) :smiley:

PR #62, opened 2017-12-20
Use Fuubar formatter

thutterer:
:cool: https://github.com/thekompanee/fuubar
I like it. Any objections? @ikapelyukhin

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,1 @@
---require spec_helper
+--require spec_helper --format Fuubar --color spec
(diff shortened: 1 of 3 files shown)


ikapelyukhin:
@thutterer Is it instafailing? I like to see all the specs turn red and fall into despair. :D
But sure, let's try it -- we can revert anytime (or move `.rspec` out of git).

thutterer:
> Is it instafailing? I like to see all the specs turn red and fall into despair

It keeps on running all the other specs, but it gives you the output of the failing specs right after they failed. So don't have to wait for the whole suite to finish before you can look into the first failing one.

ikapelyukhin:
Oh, OK. That it's fine.

Issue #58, opened 2017-12-15
"repos --all" should tell about "sync" when DB is still empty


thutterer:
On an empy database (before the first `sync`)
```
rmt-cli repos --all
```
currently outputs
```
No repositories enabled.
```

But it should behave the same way than `products --all` which in this case already outputs
```
Run "rmt-cli sync" to synchronize with your SUSE Customer Center data first.
```

ikapelyukhin:
@thutterer A good issue for the trainees to work on? :grimacing:

fschueller:
See https://github.com/SUSE/rmt/pull/63.

PR #63, opened 2017-12-20
Print message about syncing on empty database

fschueller:
#58

### DIFF
--- a/lib/rmt/cli/repos.rb
+++ b/lib/rmt/cli/repos.rb
@@ -68,7 +68,11 @@ class RMT::CLI::Repos < RMT::CLI::Base
     end
 
     if rows.empty?
-      warn 'No repositories enabled.'
+      if options.all
+        warn 'Run "rmt-cli sync" to synchronize with your SUSE Customer Center data first.'
+      else
+        warn 'No repositories enabled.'
+      end
     else
       puts Terminal::Table.new headings: ['ID', 'Name', 'Description', 'Mandatory?', 'Mirror?', 'Last mirrored'], rows: rows
     end
--- a/spec/lib/rmt/cli/repos_spec.rb
+++ b/spec/lib/rmt/cli/repos_spec.rb
@@ -102,6 +102,16 @@ RSpec.describe RMT::CLI::Repos do
       it 'outputs success message' do
         expect { command }.to output("No repositories enabled.\n").to_stderr
       end
+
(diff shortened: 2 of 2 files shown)


lib/rmt/cli/repos.rb
```diff
@@ -68,10 +68,15 @@ def list_repositories(scope: :enabled)
...
 
     if rows.empty?
-      warn 'No repositories enabled.'
+      print_warning
     else
       puts Terminal::Table.new headings: ['ID', 'Name', 'Description', 'Mandatory?', 'Mirror?', 'Last mirrored'], rows: rows
     end
   end
 
+  def print_warning
```
thutterer: Why an extra method for these two lines?
If it was reused anywhere, it would make more sense. But it isn't and only makes my :eyes: do extra work.

But maybe we can actually share some code with the `products` command?

Have a look at the list command in `lib/rmt/cli/products.rb`. It still behaves a bit different: It always mentions the `--all` command when you only do `products`. I think we should do the same here. With that change, the logic of when to output what becomes a bit complicated - as you can see in that method there.

So maybe there is a good way to generalize this method and move into base.rb?

Not sure if you wanna do this in this PR though, or just create another issue for that.

Whatever you prefer... If you go for the second option, I'd vote for not having this extra `print_warning` message as long as it is not reused anywhere.

fschueller: I also don't like the extra method, but the Code Climate check was crying about complexity :/ Previously it was very straightforward, like in `lib/rmt/cli/products.rb`. But the extra method passes the test.

Technically I'd like to avoid duplication. Probably best to move it to a shared method, so we can get rid of some `if`'s in both classes.

I'll review what's done in `products.rb` and I suppose I'll try improving it it in this PR, if that complies with the rules.

thutterer: Code Climate issues not mandatory to fix right away. It doesn't block the merge when it complains.
And especially this complexity checks can be tricked by moving things in extra methods, which not always means it is really less complex then. So better trust your own feeling more than CC :smiley: 

Cool, have a look into that DRYing up idea! But still, not mandatory to do this in this PR.

fschueller: After having a look at this, it might be a little more work to do it properly. I'll see what I can whip up something working by tomorrow.

PR #65, opened 2017-12-21
update yard gem to fix a security vulnerability

thutterer:


thutterer:
@tmuntan1 was quicker! :zap:

PR #72, opened 2018-01-03
SLE15 product tree support -- 4702

ikapelyukhin:


ikapelyukhin:
I want to add one more spec for the `ProductSerializer`, but the rest should be ready for the review.

suse-tests-pass:
@lagartoflojo will review your pull request :dancers: check https://trello.com/c/p2hEsVT1/4702-5-rmt-sles-15-product-tree-support

ikapelyukhin:
OK, that should be it.

spec/models/product_spec.rb
```diff
@@ -65,4 +65,54 @@
...
+    let(:level_two_extension_b_mirrored) do
+      FactoryGirl.create(:product, :extension, :with_mirrored_repositories, base_products: [level_one_extension], root_product: base_b)
+    end
+
+    before do
+      level_two_extension_a_not_mirrored
+      level_two_extension_b_not_mirrored
+      level_two_extension_a_mirrored
+      level_two_extension_b_mirrored
+    end
```
lagartoflojo: This looks funny, why not use `let!`?

spec/serializers/v3/product_serializer_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe 'SLES extension tree' do
+    subject(:serializer) { described_class.new(sles15, root_product: sles15, base_url: base_url) }
+
+    let(:top_extension) { serializer.as_json[:extensions].first }
+    let(:nested_extension) { top_extension[:extensions].first }
+
+    it 'has base system extension' do
+      expect(top_extension[:name]).to eq(basesystem.name)
+    end
```
lagartoflojo: These can be re-written as:
`its([:name]) { is_expected.to eq basesystem.name }`

Same for the tests below.

lagartoflojo:
This PR is not working for me:

```
> ./bin/rmt-cli sync
I, [2018-01-18T17:26:45.890722 #24033]  INFO -- : Cleaning up the database
I, [2018-01-18T17:26:45.921413 #24033]  INFO -- : Downloading data from SCC
I, [2018-01-18T17:26:45.921473 #24033]  INFO -- : Updating products
I, [2018-01-18T17:28:13.144281 #24033]  INFO -- : Updating repositories
/home/hernan/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/activerecord-5.1.3/lib/active_record/core.rb:189:in `find': Couldn't find Repository with 'id'=1094 (ActiveRecord::RecordNotFound)
	from /code/rmt/lib/rmt/scc.rb:152:in `update_auth_token'
	from /code/rmt/lib/rmt/scc.rb:33:in `block in sync'
	from /code/rmt/lib/rmt/scc.rb:32:in `each'
	from /code/rmt/lib/rmt/scc.rb:32:in `sync'
	from /code/rmt/lib/rmt/cli/main.rb:11:in `sync'
	from /home/hernan/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/command.rb:27:in `run'
	from /home/hernan/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/invocation.rb:126:in `invoke_command'
	from /home/hernan/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor.rb:387:in `dispatch'
	from /code/rmt/lib/rmt/cli/base.rb:22:in `block in dispatch'
	from /code/rmt/lib/rmt/cli/base.rb:33:in `handle_exceptions'
	from /code/rmt/lib/rmt/cli/base.rb:22:in `dispatch'
	from /home/hernan/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/base.rb:466:in `start'
	from ./bin/rmt-cli:40:in `<main>'
```

The same command works fine in master.

app/serializers/v3/product_serializer.rb
```diff
@@ -11,11 +11,12 @@ def repositories
...
-             :friendly_name, :product_class, :cpe, :free, :description, :eula_url, :repositories, :product_type, :extensions
+             :friendly_name, :product_class, :cpe, :free, :description, :eula_url, :repositories, :product_type,
+             :extensions, :recommended, :available
 
   def extensions
-    object.mirrored_extensions.map do |extension|
-      ::V3::ProductSerializer.new(extension, base_url: base_url).attributes
+    object.extensions.for_root_product(root_product).map do |extension|
+      ::V3::ProductSerializer.new(extension, base_url: base_url, root_product: root_product).attributes
     end
```
lagartoflojo: In [glue](https://github.com/SUSE/happy-customer/blob/master/glue/app/serializers/v3/product_serializer.rb#L5) we simply have a `has_many :extensions`, and all the `instance_options` are passed down to the child serializers automatically, so you don't need to specify `base_url` or `root_product`.
That leaves you with:

```ruby
has_many :extensions, serializer: V3::ProductSerializer

def extensions
  object.extensions.for_root_product(root_product)
end
```

spec/models/product_spec.rb
ikapelyukhin: Because Rubocop complains about it:

```
spec/models/product_spec.rb:93:5: C: RSpec/LetSetup: Do not use let! for test setup.
    let!(:level_two_extension_a_not_mirrored) { FactoryGirl.create(:product, :extension, base_products: [level_one_extension], root_product: base_a) }
```

I guess using the `before` block makes the initialization order more explicit.
But I can disable the check if you think it doesn't make sense.

spec/serializers/v3/product_serializer_spec.rb
ikapelyukhin: Not really, it's not the `subject` of the spec.

For the top level specs I'd need to do:

```
subject { described_class.new(sled15, root_product: sled15, base_url: base_url)[:extensions].first }
```

And for nested extensions I'd need to do:

```
    subject { described_class.new(sled15, root_product: sled15, base_url: base_url)[:extensions].first[:extensions].first }
```

That doesn't seem to make anything better, though.

app/serializers/v3/product_serializer.rb
ikapelyukhin: I remember I was looking into this at the very beginning and the current code is the solution that works correctly.

The problem is that for some reason it doesn't serialize all of the attributes of the relation, namely the ones that are declared on the serializer and don't exist on the model (like `available` and `recommended`). With the `has_many` relation they seem to be just missing.

If you have any idea what's going on with that -- feel free to share.

ikapelyukhin:
@lagartoflojo, syncing should work without issues now.

spec/models/product_spec.rb
lagartoflojo: Hmm strange rule. It must be new. We use `let!` all over the place (even in RMT hehe(

ikapelyukhin: [It seems it doesn't complain](https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/let_setup.rb#L42) about all of the `let!` calls, but only about the ones that aren't called from anywhere.

ikapelyukhin: [Here's](https://github.com/backus/rubocop-rspec/issues/94) the discussion about why it's useful.

lagartoflojo: I see. With that logic, they actual `create` calls should be moved into the `before` block then.

ikapelyukhin: Moving into the `create` block would make it impossible to reference them later in the `expect` block.
A couple of those aren't referenced, so yes, those could be moved in, but I'd say it's not very important.

lagartoflojo: Ok, let's leave it as is then.

app/models/product.rb
```diff
@@ -69,6 +77,10 @@ def change_repositories_mirroring!(conditions, mirroring_enabled)
     repositories.where(conditions).update_all(mirroring_enabled: mirroring_enabled)
   end
 
+  def recommended_for?(root_product)
```
lagartoflojo: I think we're missing unit tests for this

spec/models/product_spec.rb
```diff
@@ -115,4 +115,24 @@
...
+    it 'not recommended for base product A' do
+      extension = FactoryGirl.create(:product, :extension, base_products: [base_a], recommended: false)
+      expect(extension.recommended_for?(base_a)).to be(false)
+    end
+
+    it 'not recommended for unrelated product B' do
+      extension = FactoryGirl.create(:product, :extension, base_products: [base_a], recommended: true)
+      expect(extension.recommended_for?(base_b)).to be(false)
+    end
+  end
```
lagartoflojo: To follow the http://www.betterspecs.org/ convention, I would write these tests like this.
(Take it or leave it 😄)

```ruby
describe '#recommended_for?' do
  let(:base) { create :product }
  let(:extension) { create(:product, :extension, base_products: [base], recommended: recommended) }
  
  subject { extension.recommended_for?(queried_base) }
  
  context 'when the extension is recommended for its base' do
    let(:recommended) { true }
    let(:queried_base) { base }
    
    it { is_expected.to be true }
  end

  context 'when the extension is not recommended for its base' do
    let(:recommended) { false }
    let(:queried_base) { base }
    
    it { is_expected.to be false }
  end

  context "when the queried base is not the extension's base" do
    let(:recommended) { true }
    let(:queried_base) { create(:product) }
    
    it { is_expected.to be false }
  end
end
```

ikapelyukhin: Yep, I definitely like it! Just commit those changes yourself, maybe?

lagartoflojo:
🙈 rubocop....

ikapelyukhin:
:robot:

spec/models/product_spec.rb
ikapelyukhin: @thutterer ^^^

PR #74, opened 2018-01-04
Raise error on add_file! when not unique

tmuntan1:


ikapelyukhin:
Doesn't make much sense to handle `ActiveRecord::RecordNotUnique`, however it might make sense to handle other exceptions like `ActiveRecord::RecordNotSaved` and to output an error message.

But for that purpose you wouldn't be able to catch `StandardError` when calling `add_file!`.

PR #75, opened 2018-01-05
Fix build on ruby2.5, bundler isn't in stdlib anymore

ikapelyukhin:


Makefile
```diff
@@ -4,9 +4,6 @@ VERSION       = 0.0.2
 all:
 	@:
 
-init:
-	if [ ! -d package/.osc ]; then rm -f package/* && osc co systemsmanagement:SCC:RMT rmt-server -o package; fi;
```
ikapelyukhin: Running `make dist` removed my changes and checkout out OBS working copy (and the same would happen for everybody who has made some changes but never ran `make dist` before).

I've removed it and documented `osc` workflow instead.

README.md
```diff
@@ -151,17 +151,25 @@ EOFF
 
 The package is build in the OBS at: https://build.opensuse.org/package/show/systemsmanagement:SCC:RMT/rmt-server
```
lagartoflojo: is built*

Makefile
lagartoflojo: In [connect](https://github.com/SUSE/connect/blob/master/PACKAGE.md#step-2-refresh-your-package) the instructions are a bit simpler maybe? Since we don't make the reader check out the OBS project in a separate directory, outside the working directory. What do you think?

ikapelyukhin: OBS is essentially also a version control system, I personally find it very awkward to have the same files under two VCSes at the same time (might result in lost changes in any of them).

I also use a workflow with a separate directory with SMT, because it has 1 git repo (with two branches for SLE11 and SLE12) and 2 OBS projects -- can't check them out both in package directory. So for me using a separate directory seems quite natural :-)

README.md
ikapelyukhin: Fixed.

Makefile
lagartoflojo: Ok! Let's see how it goes. We can improve it later.

PR #76, opened 2018-01-08
Added custom repositories - Card 4721

tmuntan1:


suse-tests-pass:
@thutterer will review your pull request :dancers: check https://trello.com/c/B3g1bJtM/4721-8-rmt-custom-repos
  @thutterer: Please review this pull request using our guidelines:

spec/lib/rmt/cli/repos_spec.rb
```diff
@@ -93,82 +93,84 @@
...
+          end
+
+          it 'outputs success message' do
+            expect { command }.to output(expected_output).to_stdout
+          end
         end
 
-        it 'outputs success message' do
-          expect { command }.to output(expected_output).to_stdout
+        context 'list all' do
```
thutterer: `describe 'list --all'`

app/models/repository.rb
```diff
@@ -28,8 +55,8 @@ def change_mirroring!(mirroring_enabled)
     update_column(:mirroring_enabled, mirroring_enabled)
   end
 
-  def self.remove_suse_repos_without_tokens!
-    where(auth_token: nil).where('external_url LIKE ?', 'https://updates.suse.com%').delete_all
+  def id
```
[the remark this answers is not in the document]
tmuntan1: When I started, hundreds of tests failed without the conversion of string to int, but with all the other changes that came after, I seemed to fix the issue in the background.

When I reran the test, it was only offline mode that failed which was easy enough to fix.

suse-tests-pass:
@wstephenson will review your pull request :dancers: check https://trello.com/c/B3g1bJtM/4721-8-rmt-custom-repos
  @wstephenson: Please review this pull request using our guidelines:

lib/rmt/cli/repos.rb
```diff
@@ -1,13 +1,17 @@
 class RMT::CLI::Repos < RMT::CLI::Base
 
-  default_task :list
+  include ::RMT::CLI::RepoPrintable
+
+  desc 'custom', 'Custom Repositories'
```
thutterer: all other descriptions follow the pattern "verb object bla", see yourself with `rmt-cli repos`. So until we come up with something better, I would change it to "List and modify custom repositories"

lib/rmt/cli/custom_repos.rb
```diff
@@ -0,0 +1,90 @@
+class RMT::CLI::CustomRepos < RMT::CLI::Base
+
+  include ::RMT::CLI::RepoPrintable
+
+  desc 'add', 'Adds a custom repository to a product'
```
thutterer: Change verb to infinite form: "Add" instead of "Adds" ... same for all descriptions.

app/services/product_service.rb
```diff
@@ -0,0 +1,7 @@
+class ProductService
+
+  def get_service(product)
+    Service.find_or_create_by(product_id: product.id)
```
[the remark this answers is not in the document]
thutterer: FYI, the comments with a :question: are more for myself, to clarify when I read other parts of the code. Nothing to do here (yet) :smiley:

lib/rmt/cli/custom_repos.rb
```diff
@@ -0,0 +1,90 @@
...
+  option :name, aliases: '-n', type: :string, desc: 'The name of the custom repository', required: true
+  option :url, aliases: '-u', type: :string, desc: 'Absolute external URL to this repository', required: true
+  option :update, type: :boolean, desc: 'Update repository instead of ignore when it already exists', default: false
+  option :product_id, aliases: '-p', type: :string, required: true,
+         desc: 'The id of the product where the repository should be added. Use `rmt-cli products list` to see products'
+  def add
```
thutterer: Your options name, url and product_id are not really optional. They are all required. For this kind of data, we should use arguments, not options. This also makes the command shorter to type.
So please change to `def add(name, url, product_id)` and remove the options.
Also, let's think if we really need the last remaining option "update yes/no" ... I don't see its value. You could solve it by just always updating. Maybe add a user confirmation. Or block updating and output a warning, that a repo with this url already exists, plus the hint to just delete that one, in case someone wants to "update" it... Less options make it easier to test and easier to understand for users. IMO.

spec/factories/repositories.rb
```diff
@@ -1,11 +1,13 @@
 FactoryGirl.define do
   factory :repository do
+    sequence(:id) { |n| n }
```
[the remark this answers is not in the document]
tmuntan1: we no longer have auto-increment in change from int to string

spec/lib/rmt/scc_spec.rb
```diff
@@ -31,8 +31,8 @@
       all_repositories.map.each do |repository|
         db_repository = Repository.find(repository[:id])
 
-        (db_repository.attributes.keys - %w[external_url mirroring_enabled local_path]).each do |key|
-          expect(db_repository[key]).to eq(repository[key.to_sym])
+        (db_repository.attributes.keys - %w[external_url mirroring_enabled local_path custom]).each do |key|
+          expect(db_repository[key].to_s).to eq(repository[key.to_sym].to_s)
```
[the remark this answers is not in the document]
tmuntan1: ids being both strings and ints, but this may change depending on what you think in the end

app/models/repository.rb
```diff
@@ -7,17 +9,38 @@ class Repository < ApplicationRecord
...
+      SecureRandom.uuid.delete('-')[0...6]
+    end
+
+    def remove_suse_repos_without_tokens!
+      where(auth_token: nil).where('external_url LIKE ?', 'https://updates.suse.com%').delete_all
+    end
+
+    def only_mirrored_ids
+      Repository.only_mirrored.pluck(:id).map do |string|
+        integer = Integer(string) rescue false
```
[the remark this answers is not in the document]
tmuntan1: to not collide with scc

app/services/product_service.rb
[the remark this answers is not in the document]
tmuntan1: this was just to avoid having the class calling it to have knowledge of how the database works. I could also just place this in the model.

What do you think?

db/migrate/20180103085245_add_is_custom_to_repositories.rb
```diff
@@ -0,0 +1,7 @@
+class AddIsCustomToRepositories < ActiveRecord::Migration[5.1]
```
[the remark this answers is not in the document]
tmuntan1: add_is_custom_to_repositories.rb -> AddIsCustomToRepositories

It looks the same to me.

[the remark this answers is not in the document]
thutterer: > It looks the same to me.

I meant, the field is not called `is_custom` anymore... but we will drop this field now anyway, so nevermind :)

lib/rmt/cli/repos.rb
```diff
@@ -1,13 +1,17 @@
 class RMT::CLI::Repos < RMT::CLI::Base
 
-  default_task :list
```
thutterer: @tmuntan1 Any reason why you removed this? I'm not insisting on keeping it. One could argue that it makes it harder to get to the help output... but anyway, we should keep it consistent with how `products` command behaves.

tmuntan1: the default task hides that there are subcommands under it.

app/models/repository.rb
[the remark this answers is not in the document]
thutterer: @tmuntan1 this is not required anymore, is it?

app/services/product_service.rb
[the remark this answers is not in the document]
thutterer: > What do you think?

For me, this is just extra complication without benefit. I would just call `find_or_create_by` directly in the model.

app/services/repository_service.rb
```diff
@@ -0,0 +1,45 @@
+class RepositoryService
+
+  class InvalidExternalUrl < RuntimeError
+  end
+
+  def repository_by_id(repository_id)
```
thutterer: This method is called in 1 place and has 1 line. Kill with fire! :fire:

app/services/repository_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    Repository.find_by(id: repository_id)
+  end
+
+  def repository_by_url(url)
```
thutterer: This method is called in 1 place and has 1 line. Kill with fire! :fire:

app/services/repository_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    Repository.find_by(external_url: url)
+  end
+
+  def create_repository(product_service, url, attributes, is_custom_repository = false)
```
thutterer: This is the only thing meaningful this ServiceObject does. Please remove the other methods, and rename this service to `CreateRepositoryService` or something. Make sure it is the only public method and follow other best practices from [here](https://hackernoon.com/the-3-tenets-of-service-objects-c936b891b3c2).

app/models/repository.rb
```diff
@@ -1,5 +1,3 @@
-require 'securerandom'
```
tmuntan1: This was already removed in master https://github.com/SUSE/rmt/blob/master/app/models/repository.rb ^_^

lib/rmt/cli/mirror.rb
```diff
@@ -1,29 +0,0 @@
...
-    if repos.empty?
-      warn 'There are no repositories marked for mirroring.'
-      return
-    end
-    repos.each { |repo| mirror(repo) }
-  end
-
-  desc 'custom URL', 'Mirror a custom repository URL'
-  def custom(url, path = nil)
-    url += '/' unless url.end_with?('/')
```
tmuntan1: Was there a reason to remove this functionality?

It could still be useful to mirror things like openSUSE repositories until we also can also get custom products in.

tmuntan1:
@thutterer I pushed new changes that addresses most of your feedback.

As for the testing of aliases, I have not found a better way to implement this yet.

lib/rmt/cli/custom_repos.rb
```diff
@@ -65,15 +65,15 @@ def remove(repository_id)
       return
     end
 
-    repository_service.remove_repository(repository)
+    Repository.remove(repository)
```
thutterer: Not sure how to put it, but something feels a bit wrong here: First you check here if the repo is custom. Then you call a method `remove` which does the check again. Which is not obvious by it's name... Who will remember in 3 month that remove does sth different than destroy? After all these checks you could just use the destroy method here, I think. Or if you want a more save method on the model, which still does the check itself, you might want to call it without repeating the checks here. And instead, just make a conditional that outputs the warning when the returned value of the method was false. Something like that could remove this repeated checking.

app/services/create_repository_service.rb
```diff
@@ -1,16 +1,8 @@
-class RepositoryService
+class CreateRepositoryService
```
thutterer: I still don't like this thing too much. Just look on how it is called from `lib/rmt/scc.rb:140`:

```ruby
@repository_service.create_repository(service, repo_item[:url], repo_item)
```

- When the class is called `CreateRepositoryService`, then its instance should be called `create_repository_service`, no different.
- A service has one purpose, ergo one public method. The class name already holds the information what it does, so the method should not repeat it. The method should be called sth like `call` or `run` and that should be consistent throughout the project (which is easy because it is the first service ... and I'm not even sure we really need this concept for this purpose)
- The method interface cries for refactoring. You send in stuff from `repo_item` two times.
- Inside the method is a `# TODO` which either needs to be done now, or needs a comment wtf is happening there. It seems you punch the id into the scc_id field for non-custom repos :fist_oncoming: Why does this `attributes` object have an `id` in the first place? Maybe this can be cleaned up on the calling side and this mangling can be removed from here.

app/models/repository.rb
```diff
@@ -37,8 +37,8 @@ def by_url(url)
       Repository.find_by(external_url: url)
     end
 
-    def remove(repository)
-      return unless repository.custom?
+    def remove_if_custom(repository)
+      return false unless repository.custom?
```
thutterer: You can omit the `false` here and return nil: `repository.destroy! unless repository.custom?`

app/models/repository.rb
```diff
@@ -7,17 +7,41 @@ class Repository < ApplicationRecord
...
+    def by_id(repository_id, custom: false)
+      return Repository.find_by(id: repository_id) if custom
+      Repository.find_by(scc_id: repository_id)
+    end
+
+    def by_url(url)
+      Repository.find_by(external_url: url)
+    end
+
+    def remove_if_custom(repository)
```
thutterer: oh man, it took me three iterations of reviewing this method to come up with this suggestion:

Why not make this an instance method? :rofl: 

```ruby
def remove_if_custom
  destroy! if custom?
end
```

lib/rmt/scc.rb
```diff
@@ -133,27 +133,17 @@ def create_product(item)
...
+      create_repository_service.call(service, repo_item[:url], repo_item)
     end
   end
 
   def update_auth_token(item)
     uri = URI(item[:url])
     auth_token = uri.query
-
-    Repository.find(item[:id]).update! auth_token: auth_token
+    Repository.find_by(scc_id: item[:id]).update! auth_token: auth_token
```
thutterer: Ah, here you should also use the new `by_id` method!

app/models/repository.rb
```diff
@@ -7,17 +7,37 @@ class Repository < ApplicationRecord
...
+    def make_local_path(url)
+      uri = URI(url)
+      path = uri.path.to_s
+      path.gsub!(%r{^/repo}, '') if (uri.hostname == 'updates.suse.com')
+      path
+    end
+
+    def by_id(repository_id, custom: false)
+      return Repository.find_by(id: repository_id) if custom
+      Repository.find_by(scc_id: repository_id)
```
ikapelyukhin: Two ID fields after all? That brings tears to my eyes :cry:

app/services/create_repository_service.rb
```diff
@@ -0,0 +1,32 @@
...
```
ikapelyukhin: It's a factory. You could call it `RepositoryFactory` instead, which seems nicer IMO.

app/models/repository.rb
```diff
@@ -28,8 +48,12 @@ def change_mirroring!(mirroring_enabled)
...
   end
 
-  def self.remove_suse_repos_without_tokens!
-    where(auth_token: nil).where('external_url LIKE ?', 'https://updates.suse.com%').delete_all
+  def custom?
+    scc_id.nil?
+  end
+
+  def remove_if_custom
+    destroy! if custom?
```
ikapelyukhin: It's a very simple one-line method and it seems it is used only once. Is it really required?

app/models/repository.rb
```diff
@@ -7,17 +7,37 @@ class Repository < ApplicationRecord
...
+      path
+    end
+
+    def by_id(repository_id, custom: false)
+      return Repository.find_by(id: repository_id) if custom
+      Repository.find_by(scc_id: repository_id)
+    end
+
+    def by_url(url)
+      Repository.find_by(external_url: url)
```
ikapelyukhin: It's a very simple one-line method and it seems it is used only once. Is it really required?

spec/factories/repositories.rb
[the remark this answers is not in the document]
ikapelyukhin: That should no longer be the case?

app/models/repository.rb
tmuntan1: It is used to prevent other classes having knowledge of the database columns.

app/models/repository.rb
tmuntan1: I would say that the extra protection in this method is better than having the possibility of the same logic being repeated and implemented incorrectly. This is our protection from deleting products from the database.

app/services/create_repository_service.rb
tmuntan1: I agree with you about factories, but this is following the service pattern in rails. Because this pattern was also used in happy-customer, the same pattern was repeated.

app/models/repository.rb
ikapelyukhin: I still can just call `Repository.destroy!` anywhere in the code or in the console. Does having this method really protect against anything?

Adding a destroy hook and raising an exception if the repo is non-custom would prevent this.

app/services/create_repository_service.rb
ikapelyukhin: It's just a naming thing. The fact that some other project uses services shouldn't prevent you from calling this a factory.

app/models/repository.rb
ikapelyukhin: But I think you are trying to solve the problem that doesn't exist/can't be solved. Other developers wouldn't know that this method exists, and would just call `destroy!` instead -- can't protect against that.

app/models/repository.rb
ikapelyukhin: To be honest, I don't think that having knowledge of the DB structure in other places is such a big deal. This column will most likely never change, so it will not need to be refactored. The developer has still to know that the column exists in the DB. They also have to know that this method exists.

public/repo/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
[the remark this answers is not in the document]
tmuntan1: You need to keep `public/repo` anyways, so a `public/repo/.gitkeep` would be necessary.

If you run any command after deleting the repo directory, you get the following issue:
```
bin/rmt-cli export data ~/rmt 
```
```
Mirroring base directory (/home/thomas/Projects/rmt/public/repo) is not writable by user 'thomas'
Run as root or adjust the permissions.
```

Try `rm -rf public/repo/*`?

app/models/repository.rb
tmuntan1: I like to keep the responsibility of knowing about the internals of the database to as few locations as possible. When your codebase is small, this is not as much of an issue, but as it grows, having database calls everywhere can become a mess and much harder to refactor.

As to not knowing that the method exists, that is a symptom to the problem. It helps if the developer has only one set of places to look.

app/models/repository.rb
tmuntan1: Good idea, I just pushed a new commit with callback

spec/factories/repositories.rb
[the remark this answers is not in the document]
tmuntan1: I'm not sure why, but tests fail everywhere when I remove this line.

thutterer:
All my "requests for changes" from earlier are done now. So I'm dismissing that review. But I think I have committed too much code and thinking into this PR to be unbiased enough to officially "approve" it alone. At least as long as @ikapelyukhin is unhappy ;)

This PR is not (yet) the full feature anyway. For me, this is still missing to call it a day:

- So far it is impossible to see the URL after I added a custom repo. Also I cannot see to which products it is connected to.
So we are definitely missing a details-view for custom repos or something like that. 

- The `repos custom add` still feels a bit confusing. Given a new URL, it creates a new repo and connects it to given product_id. Given a known URL but different product_id, it connects the existing repo to that product_id too, okay, but it allows me to accidentally change the name of the existing repo at the same time. That's a bit weird.
And there is now way to delete these connections to products yet.
So I guess we need more straightforward commands, something like `create/remove` (a custom repo) and `connect/disconnect` (a custom repo to/from a product).

- It is confusing that I have to connect my custom repo to a product, but even if that product was never enabled, the repo itself gets enabled automatically and it gets mirrored next time. So we still need to define how this should behave.

- Shouldn't removing a custom repo also remove the downloaded files from disk? Should this be a cli option?

ikapelyukhin:
Do custom repos work currently with offline mode?

thutterer:
> Do custom repos work currently with offline mode?

No they don't. We had a discussion today in the office how to do it. Maybe that's why I forgot to add it. Too obvious :grimacing: Updated my comment.

PR #83, opened 2018-01-24
Metadata caching 4761

ikapelyukhin:


suse-tests-pass:
@tmuntan1 will review your pull request :dancers: check https://trello.com/c/EWnnSc3b/4761-rmt-metadata-redownload-issue
  @tmuntan1: Please review this pull request using our guidelines:

lib/rmt/downloader.rb
```diff
@@ -85,24 +99,35 @@ def make_request(remote_file, local_file, request_fiber, checksum_type = nil, ch
 
     downloaded_file = Tempfile.new('rmt', Dir.tmpdir, mode: File::BINARY, encoding: 'ascii-8bit')
 
+    headers = {}
+    headers['If-Modified-Since'] = if_modified_since if if_modified_since
```
thutterer: if if if if? Would read nicer to me if the variable name wouldn't start with "if_" ... still voting for renaming it to something more nice :name_badge:

ikapelyukhin: :rofl: Yes, that turned out a bit weird.

lib/rmt/downloader.rb
```diff
@@ -4,34 +4,47 @@
...
     request_fiber = Fiber.new do
-      make_request(remote_file, local_file, request_fiber, checksum_type, checksum_value)
+      request = make_request(remote_file, request_fiber, if_modified_since)
+      finalize_download(request, local_file, checksum_type, checksum_value)
     end
 
     request_fiber.resume.run
 
+    local_file
+  rescue NotModifiedException
```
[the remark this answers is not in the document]
ikapelyukhin: I guess it depends on how to look at it.

Normally we expect a response code `200` with a body that contains the file that have we requested. But now we'd get in some cases (not always) a response code `304` with an empty body, which we cannot process as we normally would further.

But it should be possible to move this logic into `#make_request`, between:
```
request.receive_headers
request.receive_body
```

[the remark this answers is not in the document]
thutterer: Up to you! You definitely have given that more thought than I have. So whatever feels better to you :+1:

lib/rmt/downloader.rb
```diff
@@ -4,34 +4,47 @@
...
     @repository_url = repository_url
     @local_path = local_path
     @concurrency = 4
     @auth_token = auth_token
     @logger = logger || Logger.new('/dev/null')
+    @cache_path = cache_path
   end
 
-  def download(remote_file, checksum_type = nil, checksum_value = nil)
+  def download(remote_file, checksum_type: nil, checksum_value: nil, use_cache: false)
```
thutterer: Why not make `use_cache: true` the default behavior? That would save you from explicitly setting it with each call to this method.

lib/rmt/downloader.rb
```diff
@@ -27,24 +27,25 @@ def initialize(repository_url:, local_path:, auth_token: nil, logger: nil, cache
   def download(remote_file, checksum_type: nil, checksum_value: nil, use_cache: false)
     local_file = self.class.make_local_path(@local_path, remote_file)
 
-    if_modified_since = nil
+    cache_timestamp = nil
     if use_cache
       raise 'Cache path not set!' unless @cache_path
       cache_file = File.join(@cache_path, remote_file)
-      if_modified_since = File.mtime(cache_file).utc.rfc2822 if File.exist?(cache_file)
+      cache_timestamp = File.mtime(cache_file).utc.rfc2822 if File.exist?(cache_file)
```
thutterer: Thanks! Reads so much better now! :+1:

thutterer: I found out why it does not work for me. Your time format resultet in "Thu, 25 Jan 2018 09:34:05 -0000" on my machine.
When I replace that with `File.mtime(cache_file).gmtime.strftime('%a, %d %b %Y %T GMT')` it works.
I haven't found a method or %-shortcut for that. :confused:

spec/lib/rmt/downloader_spec.rb
```diff
@@ -103,8 +103,72 @@
...
 
+    describe '#download with If-Modified-Since' do
+      let(:cache_dir) { Dir.mktmpdir }
+      let(:repo_dir) { Dir.mktmpdir }
+      let(:downloader) { described_class.new(repository_url: 'http://example.com', local_path: repo_dir, cache_path: cache_dir) }
+      let(:time) { Time.utc(2018, 1, 1, 10, 10, 0) }
+      let(:if_modified_headers) do
+        {
+          'User-Agent' => "RMT/#{RMT::VERSION}",
+          'If-Modified-Since' => 'Mon, 01 Jan 2018 10:10:00 -0000'
```
thutterer: This needs to be updated with my proposed change of strftime above.

spec/lib/rmt/downloader_spec.rb
```diff
@@ -103,8 +103,72 @@
...
+        }
+      end
+      let(:filename) { 'repomd.xml' }
+      let(:downloaded_file) { downloader.download("/#{filename}", use_cache: true) }
+      let(:cached_content) { 'cached_content' }
+      let(:fresh_content) { 'fresh_content' }
+
+      before do
+        cache_dir
```
thutterer: You can use `let!` instead.
https://relishapp.com/rspec/rspec-core/v/3-4/docs/helper-methods/let-and-let

thutterer: Actually, in this case, you do not need this before block at all. You reference cache_dir and repo_dir in another let above, so they get evaluated before the test runs. Please remove the before block.

lib/rmt/downloader.rb
ikapelyukhin: [There's also](https://ruby-doc.org/stdlib-2.2.2/libdoc/time/rdoc/Time.html#method-i-httpdate) `httpdate` method, it produces the date in the correct format.

spec/lib/rmt/downloader_spec.rb
ikapelyukhin: Yeah, but I guess that's why there's a Rubocop rule against `let!` -- with lazy loading it's hard to say what gets initialized when in the code :D

lib/rmt/downloader.rb
ikapelyukhin: I was thinking that it would make it explicit that cache is being used. I could just enable checking cache whenever `cache_path` is set.

thutterer: Up to you!

ikapelyukhin: But yeah, at the moment caching should be used for all `downloader.download` calls, so I've just enabled it by default.

ikapelyukhin:
OK, now we should have specs to make sure that cached metadata is being used during mirroring.

lib/rmt/downloader.rb
```diff
@@ -117,7 +120,7 @@ def make_request(remote_file, request_fiber, cache_timestamp = nil)
   end
 
   def copy_from_cache(cache_file, local_file)
-    FileUtils.cp(cache_file, local_file)
+    FileUtils.cp(cache_file, local_file) unless (cache_file == local_file)
```
thutterer: Have you checked if `cache_file == local_file` has any performance issues on larger files? Wouldn't checksum be good enough? There is also [File.identical?](http://ruby-doc.org/core-2.2.0/File.html#method-c-identical-3F) :confused:

ikapelyukhin: Both are actually filenames (strings). I should probably rename the variables to `cache_filename` and `local_filename`.

ikapelyukhin:
@thutterer The licenses are now downloaded into a temp directory first, same as repo metadata.

PR #85, opened 2018-01-25
Custom Repos Pt. 2 / Card 4721

tmuntan1:


suse-tests-pass:
@ikapelyukhin will review your pull request :dancers: check https://trello.com/c/B3g1bJtM/4721-8-rmt-custom-repos
  @ikapelyukhin: Please review this pull request using our guidelines:

lib/rmt/cli/repos_custom.rb
```diff
@@ -85,14 +85,14 @@ def products(id)
 
   desc 'attach ID PRODUCT_ID', 'Attach an existing custom repository to a product'
   def attach(id, product_id)
-    product, repository = attach_or_detach(id, product_id)
+    product, repository = attach_or_detach(id, product_id) || return
```
ikapelyukhin: Why not to use `find_by!` and catch the exception?

ikapelyukhin:
1. What about the feedback from the previous pull request?
2. How would the commands to enable/disable a custom repository look like?
3. It is likely that we'd have to implement a command to mirror a single repository. How this command look like?

tmuntan1:
1) Most of your feedback has already implemented. I will not make another db change without reason, but you are free to submit feedback on this pr.
2) It looks like this was missed, but I will see to adding it.
3) If you mean to mirror without adding to the db, this is outside the scope of this card and would most likely need to be groomed to be done correctly or to see if it has any business value.

lib/rmt/cli/export.rb
```diff
@@ -10,7 +10,8 @@ def data(path)
   desc 'settings PATH', 'Store repository settings at given path'
   def settings(path)
     needs_path(path) do
-      File.write(File.join(path, 'repos.json'), Repository.only_mirrored.pluck(:id).to_json)
+      data = Repository.only_mirrored.inject([]) { |data, repo| data << { url: repo.external_url, auth_token: repo.auth_token.to_s } }
```
ikapelyukhin: That doesn't seem to do what it needs to.

For SCC repos dumping just the ids was fine, because all the data about those repos is contained in the `scc-data` JSON dump.

For custom repos, however, there's no info about the repos in the JSON dump -- and the current code wouldn't preserve ids, names and associations with products.

lib/rmt/cli/import.rb
```diff
@@ -10,15 +10,22 @@ def data(path)
...
       end
 
-      repos.each do |repo|
-        repo.external_url = 'file://' + path + Repository.make_local_path(repo.external_url)
-        mirror!(repo)
+      repos = JSON.parse(File.read(repos_file))
+      repos.each do |repo_json|
+        repo = Repository.find_by(external_url: repo_json['url'])
+        if repo.nil?
+          warn "repository by url #{repo_json['url']} does not exist in database"
```
ikapelyukhin: That wouldn't work for the custom repos.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  include ::RMT::CLI::ArrayPrintable
+
+  desc 'add URL NAME', 'Creates a custom repository.'
+  def add(url, name)
+    previous_repository = Repository.find_by(external_url: url)
+
+    if previous_repository
+      warn 'A repository by this URL already exists.'
+      return
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

README.md
```diff
@@ -41,19 +41,31 @@ desired location.
...
 * Run `rmt-cli products --all` to see the list of products that are available for your organization
 * Run `rmt-cli repos --all` to see the list of all repositories available
-* Use the `rmt-cli repos enable` command to choose which product repositories to mirror, for example:
+* Use the `rmt-cli products enable` command to choose which product repositories to mirror, for example:
   ```
-  rmt-cli repos enable SLES/12.2/x86_64
+  rmt-cli products enable SLES/12.2/x86_64
   ```
   The above command would select the mandatory (`pool`, `updates`) SLES 12 SP2 repositories to be mirrored.
   Alternatively, you can specify repository ID to choose individual repositories.
```
ikapelyukhin: This line needs to be changed to say that mirroring of individual repositories can be changed by using `rmt-cli repos enable/disable REPO_ID` commands.

README.md
```diff
@@ -41,19 +41,31 @@ desired location.
...
 * Run `rmt-cli mirror` to mirror selected repositories and make their content available through RMT
 * Register client against RMT by running `SUSEConnect --url https://rmt_hostname`
   After successful registration the repositories from RMT will be used by `zypper` on the client machine.
 
+### Custom Repositories
+
+* Run `rmt-cli repos custom add URL NAME PRODUCT_ID` to add a new custom repository to a product, for example:
```
ikapelyukhin: `PRODUCT_ID` is no longer an allowed parameter. The step with attaching the product should be additionally documented.

lib/rmt/cli/export.rb
ikapelyukhin: We've established that this works correctly :+1:

lib/rmt/cli/import.rb
ikapelyukhin: We've established that this works correctly :+1:

lib/rmt/cli/products.rb
```diff
@@ -40,24 +40,28 @@ def disable(target)
...
-
-  def change_product_mirroring_by_id(id, mirroring_enabled)
-    conditions = { mirroring_enabled: !mirroring_enabled } # to only update the repos which need change
-    conditions[:enabled] = true if mirroring_enabled
 
-    repo_count = Product.find(id).change_repositories_mirroring!(conditions, mirroring_enabled)
+    repo_count = repository_service.change_mirroring_by_product!(set_enabled, products)
+    puts "#{repo_count} repo(s) successfully #{set_enabled ? 'enabled' : 'disabled'}."
+  rescue ActiveRecord::RecordNotFound
+    warn "Product by id \"#{product_id}\" not found."
```
ikapelyukhin: `RMT::CLI::Error` should be raised here, for consistency as well as for setting the exit code.
We've got some special handling for it in `RMT::CLI::Base`, which hides the stack trace and sets the exit code to a non-zero value, depending on the type of exception.

lib/rmt/cli/repos.rb
```diff
@@ -77,4 +46,17 @@ def list_repositories(scope: :enabled)
     end
   end
 
+  private
+
+  def change_mirroring(target, set_enabled)
+    repository_service.change_repository_mirroring!(target, set_enabled)
+    puts "Repository successfully #{set_enabled ? 'enabled' : 'disabled'}."
+  rescue RepositoryService::RepositoryNotFound => e
+    warn e.message
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+      repository_service.create_repository!(nil, url, {
+        name: name,
+        mirroring_enabled: 1,
+        autorefresh: 1,
+        enabled: 0
+      }, custom: true)
+
+      puts 'Successfully added custom repository.'
+    rescue RepositoryService::InvalidExternalUrl => e
+      warn "Invalid URL \"#{e.message}\" provided."
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+    change_mirroring(target, false)
+  end
+
+  desc 'remove ID', 'Remove a custom repository'
+  def remove(id)
+    repository = find_repository(id)
+
+    if repository.nil?
+      warn "Cannot find custom repository by id \"#{id}\"."
+      return
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+      warn "Invalid URL \"#{e.message}\" provided."
+    end
+  end
+
+  desc 'list', 'List all custom repositories'
+  def list
+    repositories = Repository.only_custom
+
+    if repositories.empty?
+      warn 'No custom repositories found.'
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+  end
+  map rm: :remove
+
+  desc 'products ID', 'Shows products attached to a custom repository'
+  def products(id)
+    repository = find_repository(id)
+
+    if repository.nil?
+      warn "Cannot find custom repository by id \"#{id}\"."
+      return
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+    end
+
+    products = repository.products
+
+    if products.empty?
+      warn 'No products attached to repository.'
+      return
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+    puts 'Detached repository from product'
+  end
+
+  private
+
+  def change_mirroring(target, set_enabled)
+    repository_service.change_repository_mirroring!(target, set_enabled, scc_repository: false)
+    puts "Repository successfully #{set_enabled ? 'enabled' : 'disabled'}."
+  rescue RepositoryService::RepositoryNotFound => e
+    warn e.message
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+    @repository_service ||= RepositoryService.new
+  end
+
+  def attach_or_detach(id, product_id)
+    repository = find_repository(id)
+    product = find_product(product_id)
+
+    if repository.nil?
+      warn "Cannot find custom repository by id \"#{id}\"."
+      return false
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,144 @@
...
+    elsif product.nil?
+      warn "Cannot find product by id \"#{product_id}\"."
+      return false
```
ikapelyukhin: That should raise `RMT::CLI::Error`.

spec/support/mirroring_helpers.rb
```diff
@@ -0,0 +1,4 @@
+
+def repo_url_to_local_path(url)
+  'file://' + path + Repository.make_local_path(url)
```
ikapelyukhin: What's `path` here?

lib/rmt/cli/repos_custom.rb
tmuntan1: This has been updated but github doesn't seem to pick up on it.

spec/lib/rmt/cli/repos_custom_spec.rb
```diff
@@ -0,0 +1,359 @@
...
+        expect(Repository.find_by(name: 'foo')).not_to be_nil
+      end
+    end
+
+    context 'invalid URL' do
+      let(:argv) { %w[add http://foo.bar foo] }
+
+      it 'adds the repository to the database' do
+        expect(described_class).to receive(:exit)
+        expect { described_class.start(argv) }.to output("Invalid URL \"http://foo.bar\" provided.\n").to_stderr.and output('').to_stdout
```
ikapelyukhin: Technically that's a valid URL. (Maybe I'm hosting my custom repos somewhere like `http://localhost/`?)

app/services/repository_service.rb
```diff
@@ -0,0 +1,60 @@
+class RepositoryService
+
+  class RepositoryNotFound < RuntimeError
+  end
+
+  def create_repository!(product, url, attributes, custom: false)
+    repository = Repository.find_or_initialize_by(external_url: url)
+
+    # TODO: See if we can clean this up
```
wstephenson: Please clarify the problem here.

lib/rmt/cli/export.rb
```diff
@@ -1,7 +1,7 @@
 class RMT::CLI::Export < RMT::CLI::Base
 
-  desc 'data PATH', 'Store SCC data in files at given path'
-  def data(path)
+  desc 'scc-data PATH', 'Store SCC data in files at given path'
```
wstephenson: What was the thinking in renaming this 'scc-data'? It seems overly dependent on the identity of scc, which might change over the lifespan of rmt. 'upstream-data' instead?

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    puts array_to_table(products, {
+      id: 'Product ID',
+      name: 'Product Name'
+    })
+  end
+
+  desc 'attach ID PRODUCT_ID', 'Attach an existing custom repository to a product'
+  def attach(id, product_id)
+    product, repository = attach_or_detach(id, product_id)
```
wstephenson: It isn't clear from the method name that attach_or_detach looks up a product and a repository from a pair of IDs.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,124 @@
...
+    Product.find_by!(id: id)
+  rescue ActiveRecord::RecordNotFound
+    raise RMT::CLI::Error.new("Cannot find product by id \"#{id}\".")
+  end
+
+  def repository_service
+    @repository_service ||= RepositoryService.new
+  end
+
+  def attach_or_detach(id, product_id)
```
wstephenson: See above.

app/services/repository_service.rb
tmuntan1: This was extracted from older code during the refactoring process, nothing has changed, but I don't like it at all. 

Unfortunately, I haven't see an easy way to clean it up, so I just made a note to clean it up in the future. Also, this was from a previous commit, but the file was renamed so it shows a change when there really is none.

lib/rmt/cli/repos_custom.rb
tmuntan1: This wasn't my code, but you are totally right.

tmuntan1: This wasn't my code, but you are totally right.

README.md
```diff
@@ -41,19 +41,37 @@ desired location.
 * Run `rmt-cli sync` to download available products and repositories data for your organization from SCC
 * Run `rmt-cli products --all` to see the list of products that are available for your organization
```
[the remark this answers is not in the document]
tmuntan1: nice catch

app/services/repository_service.rb
ikapelyukhin: @tmuntan1 Let's remove the TODO then. No point in keeping it in the code forever.

lib/rmt/cli/repos_custom.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    if previous_repository
+      raise RMT::CLI::Error.new('A repository by this URL already exists.')
+    end
+
+    repository_service.create_repository!(nil, url, {
+      name: name,
+      mirroring_enabled: 1,
+      autorefresh: 1,
+      enabled: 0
```
ikapelyukhin: It would be better to use the correct data types: `true`/`false` instead of `0`/`1`.

spec/services/repository_service_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:suse_repository) { create :repository }
+
+  describe '#create_repository' do
+    before do
+      service.create_repository!(product, 'http://foo.bar/repos', {
+        name: 'foo',
+        mirroring_enabled: true,
+        description: 'foo',
+        autorefresh: 1,
+        enabled: 0
```
ikapelyukhin: Also would be better to use the correct data types.

lib/rmt/mirror.rb
```diff
@@ -35,14 +38,17 @@ def mirror
     replace_directory(@temp_licenses_dir, File.join(@repository_dir, '../product.license/'))
   end
 
-  def self.from_repo_model(repository, base_dir = nil)
+  def self.from_uri(uri, auth_token, repository_url: nil, base_dir: nil, to_offline: false)
```
ikapelyukhin: Technically, it works only with URLs (we don't support URNs, which are also URIs).

PR #90, opened 2018-01-31
Contestant 1: ID is a MD5 hash of a URL

ikapelyukhin:
```
$ rmt-cli repos custom add http://example.com/test1 Example1
Successfully added custom repository.
$ rmt-cli repos custom add http://example.com/test2 Example2
Successfully added custom repository.
$ rmt-cli repos custom list
+----------------------------------+----------+--------------------------+------------+---------+---------------+
| ID                               | Name     | URL                      | Mandatory? | Mirror? | Last Mirrored |
+----------------------------------+----------+--------------------------+------------+---------+---------------+
| ffdd055a25599d926ac090f1d883037e | Example1 | http://example.com/test1 | false      | true    |               |
| 582364a2b6e410b78c1ee5128019e2b1 | Example2 | http://example.com/test2 | false      | true    |               |
+----------------------------------+----------+--------------------------+------------+---------+---------------+
```

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -15,6 +15,7 @@ class Repository < ApplicationRecord
   validates :local_path, presence: true
 
   before_destroy :ensure_destroy_possible
+  before_create :set_unique_id
 
   class << self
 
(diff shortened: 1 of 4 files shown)


ikapelyukhin:
But SHA1 would have exactly the same issue? Also, it's even longer.

ikapelyukhin:
SHA1 has the same issue, UUID is basically same as #92, just longer.

PR #96, opened 2018-02-09
Added nginx to the rmt-server package - Card 15

tmuntan1:


package/rmt-server.spec
```diff
@@ -40,6 +40,14 @@ Source0:        %{name}-%{version}.tar.bz2
 Source1:        rmt-server-rpmlintrc
```
lagartoflojo: We should probably bump the version for this as well

lagartoflojo:
Getting a weird error when trying to `osc build` the package.
Will check again next week:

```
[    2s] removing nis flags from /var/tmp/build-root/SLE_15-x86_64/etc/nsswitch.conf...
[    2s] now finalizing build dir...
[    3s] -----------------------------------------------------------------
[    3s] I have the following modifications for rmt-server.spec:
[    3s] -----------------------------------------------------------------
[    3s] ----- building rmt-server.spec (user abuild)
[    3s] -----------------------------------------------------------------
[    3s] -----------------------------------------------------------------
[    3s] + exec rpmbuild -ba --define '_srcdefattr (-,root,root)' --nosignature /home/abuild/rpmbuild/SOURCES/rmt-server.spec
[    3s] rpmbuild: error while loading shared libraries: librpmbuild.so.8: cannot open shared object file: No such file or directory
[    3s] 
[    3s] localhost failed "build rmt-server.spec" at Fri Feb  9 15:57:52 UTC 2018.
```

package/rmt-server-sync.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=RMT Sync timer
+After=rmt.service
+Requires=rmt.service
+
+[Timer]
+OnCalendar=*-*-* 01:00:00
```
lagartoflojo: All RMTs in the same timezone will bomb our servers at the same time! Self-inflicted DDoS FTW!

Maybe we can randomize this.

digitaltom: 🙏 please randomize :-), or at least not at the same time as all SMTs (they also DDoS us at 1:00). 

Using something like `H H * * *` might be good.

tmuntan1: I put in a random delay of 3 hours

package/rmt-server.spec
tmuntan1: version has been bumped to "0.0.3"

lagartoflojo:
[LGTM](http://www.lgtm.in/i/069eAehAn)

PR #108, opened 2018-02-20
Added removal of patches to changelog

tmuntan1:
changelog check on obs failed because there was no reference to removed patches


tmuntan1:
@kalabiyau thanks

PR #114, opened 2018-03-13
Point nginx to /usr/share/rmt/public

### LINKED ISSUE #113: Wrong root in nginx config
Based on the report from QA: [current nginx config](https://github.com/SUSE/rmt/blob/e8303a20e8ae938aad645dfeb749b3425d7d4204/package/rmt-server.conf#L10) uses `/var/lib/rmt/public` as a root directory.

`rmt-cli` commands download the files into `/usr/share/rmt/public`, which is a symlink to `/var/lib/rmt/public`.

If the user needs to download files somewhere else, they would change the `/usr/share/rmt/public` symlink to point somewhere else -- nginx, however, will still serve files from `/var/lib/rmt/public`.

ikapelyukhin:
Fixes #113.

### DIFF
--- a/package/rmt-server.conf
+++ b/package/rmt-server.conf
@@ -7,7 +7,7 @@ server {
     server_name rmt;
     access_log  /var/log/nginx/rmt_access.log;
     error_log   /var/log/nginx/rmt_error.log;
-    root        /var/lib/rmt/public;
+    root        /usr/share/rmt/public;
 
     location / {
         try_files $uri/index.html $uri.html $uri @rmt_app;


lagartoflojo:
Would this need to change as well?

 https://github.com/SUSE/rmt/blob/e8303a20e8ae938aad645dfeb749b3425d7d4204/package/rmt-server.spec#L25

ikapelyukhin:
@lagartoflojo No, it is actually needed to stay as is to create the symlinks, as you can see here:
https://github.com/SUSE/rmt/blob/e8303a20e8ae938aad645dfeb749b3425d7d4204/package/rmt-server.spec#L114

PR #120, opened 2018-03-23
Update rails html sanitizer & cut version 0.0.4

lagartoflojo:


package/rmt-server.changes
```diff
@@ -1,3 +1,14 @@
+-------------------------------------------------------------------
+Fri Mar 23 10:51:14 UTC 2018 - hschmidt@suse.com
+
+- Version 0.0.4
+- Fix nginx config to point to correct root location
```
digitaltom: was there a bug# that we could reference here?

PR #121, opened 2018-03-23
Offline migration endpoint

fschueller:
~~Please note that after running the migrations and syncing, it might break on a `foreign key constraint fails`. 
This happens because apparently we have migration paths where the predecessor of the product has a numerically higher ID than the product itself. This constraint was introduced in https://github.com/SUSE/rmt/commit/10e57dd4fb12ba93d58b8993e80bc60a3a742e1f, which is not part of this PR.~~

EDIT: this should be fixed with f9041e0.


lagartoflojo:
I was just thinking about this and I realized we can solve the foreign constraint problem in yet way. We can move the association creation out of `#create_product`, and create the associations AFTER we've imported all products instead!

app/models/migration_engine.rb
```diff
@@ -45,23 +56,76 @@ def base_product
...
     end
   end
 
+  # SLE 15 needs to have all of its available modules enabled, because there is
+  # no direct mapping from SLE12 modules to SLE15 modules.
+  # This method modifies the given array.
+  def add_dependencies(migrations)
+    migrations.map do |migration|
+      base = migration.first
+      migration.concat Product.available_or_recommended_modules(base.id) if base.version.split('.').first == '15'
```
tmuntan1: Do we even need to enter this code SLES 12?

Also, I wouldn't just look if the base version is 15, but see if it's below 15.

lib/rmt/scc.rb
```diff
@@ -80,36 +70,53 @@ def import(path)
...
+  def cleanup_database
+    @logger.info('Cleaning up the database')
+    Subscription.delete_all
+  end
+
+  def update_repositories(data)
     @logger.info('Updating repositories')
-    data = JSON.parse(File.read(File.join(path, 'organizations_repositories.json')), symbolize_names: true)
-    data.each do |item|
+    repos = data
```
tmuntan1: Is there a reason for this?

lib/rmt/scc.rb
```diff
@@ -80,36 +70,53 @@ def import(path)
...
     end
+  end
 
-    Repository.remove_suse_repos_without_tokens!
-
+  def update_subscriptions(data)
     @logger.info('Updating subscriptions')
-    data = JSON.parse(File.read(File.join(path, 'organizations_subscriptions.json')), symbolize_names: true)
-    data.each do |item|
+    subscriptions = data
```
tmuntan1: Is there a reason for this?

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -24,19 +24,24 @@ def show
...
-          ActiveModelSerializers::SerializableResource.new(
-            item,
-            each_serializer: ::V3::UpgradePathItemSerializer
-          )
-        end
-      ).to_json
+      offline_upgrade_paths = MigrationEngine.new(@system, installed_products)
+        .offline_migrations(product_from_hash(params[:target_base_product]))
+
+      render_migration_path offline_upgrade_paths
```
tmuntan1: a bit nick-picky, I would have `render_migration_path` only create the json and keep the `render` in the controller action.

This isn't a blocking comment.

spec/lib/rmt/scc_spec.rb
```diff
@@ -31,6 +31,21 @@
       end
     end
 
+    it 'creates the correct predecessor association when predecessor exists' do
+      products.each do |product|
+        extension = product[:extensions][0]
+
+        [product, extension].each do |product|
```
tmuntan1: The variable `product` is already declared.

This isn't a blocking comment.

app/models/migration_engine.rb
lagartoflojo: If you're upgrading a system from SLE12SP3 to SLE15, this code adds all the SLE15 modules to the migration.
This should ONLY happen when you are upgrading to SLE15.

I'm not sure if that addresses your comment.

lagartoflojo:
Oh, the dangling comment is probably invalid (the thing works now) so I'll remove it.

app/models/migration_engine.rb
ikapelyukhin: It seems like it doesn't account for whether the extension is mirrored or not.
I think the migration would fail should you have even one of the suggested modules not mirrored. :thinking:

fschueller: I suppose this calls for another card.

PR #122, opened 2018-03-27
Use SSL by default

ikapelyukhin:
Now RMT would serve `clientSetup4RMT.sh` script at http://rmt-hostname/tools/clientSetup4RMT.sh
CA certificate will be available at https://rmt-hostname/rmt.crt


package/nginx-https.conf
```diff
@@ -10,7 +10,7 @@ server {
     error_log   /var/log/nginx/rmt_https_error.log;
     root        /usr/share/rmt/public;
 
-    ssl_certificate     /usr/share/rmt/ssl/rmt-server.pem;
+    ssl_certificate     /usr/share/rmt/ssl/rmt-server.crt;
```
kalabiyau: @ikapelyukhin we are suggesting users of nginx to switch to TLS1.2 only. See [line 99](https://build.opensuse.org/package/view_file/server:http/nginx/nginx-1.6.1-default_config.patch?expand=1)

Would that make sense to align RMT config?

ikapelyukhin: I'm not sure if TLSv1.2 would work with openssl version that is on SLES12 installation media, but for all modern use cases -- sure, it makes sense.

package/nginx-https.conf
```diff
@@ -12,6 +12,7 @@ server {
 
     ssl_certificate     /usr/share/rmt/ssl/rmt-server.crt;
     ssl_certificate_key /usr/share/rmt/ssl/rmt-server.key;
+    ssl_protocols       TLSv1.2 TLSv1.3;
```
ikapelyukhin: @kalabiyau I've thrown in TLSv1.3 as well, just for the future. Should the cipher suite also be modified?

kalabiyau: @ikapelyukhin 

I think default still makes sense and there was no specific advise from security team on that bit. So, I suggest we keep it this way for now.

```bash
# ssl_ciphers  HIGH:!aNULL:!MD5;
```

ikapelyukhin: OK! :+1:

public/tools/clientSetup4RMT.sh
```diff
@@ -0,0 +1,306 @@
...
+if [ -d "/etc/pki/trust/anchors/" ] && [ -x "/usr/sbin/update-ca-certificates" ]; then
+    CA_TRUSTSTORE="/etc/pki/trust/anchors/"
+    CA_GEN_TRUSTSTORE_CMD="/usr/sbin/update-ca-certificates"
+elif [ -d $CA_TRUSTSTORE ] && [ -x $CA_GEN_TRUSTSTORE_CMD ]; then
+    CA_GEN_TRUSTSTORE_CMD="$CA_GEN_TRUSTSTORE_CMD $CA_TRUSTSTORE"
+fi
+
+function usage()
+{
+    if [ -n "$1" ] ; then
```
felixsch: What is this? Is this required?

public/tools/clientSetup4RMT.sh
```diff
@@ -0,0 +1,306 @@
...
+    exit 1
+fi
+
+
+if [ ! -x $OPENSSL ]; then
+    echo "openssl command not found. Abort."
+    exit 1
+fi
+
+if [ ! -x $CP ]; then
```
felixsch: Are the checks for `cp` and `cat` really required? Since they are in `coreutils` and I can not imagine a generic system without this package

public/tools/clientSetup4RMT.sh
```diff
@@ -0,0 +1,306 @@
...
+if [ ! -x $GREP ]; then
+    if [ -x "/bin/grep" ]; then
+        GREP=/bin/grep
+    else
+        echo "grep command not found. Abort."
+        exit 1
+    fi
+fi
+
+if [ ! -x $RM ]; then
```
felixsch: same here; `rm` and `chmod` both also in `coreutils`

package/nginx-http.conf
```diff
@@ -0,0 +1,11 @@
+server {
+    listen 80   default;
+    server_name rmt;
+    access_log  /var/log/nginx/rmt_http_access.log;
+    error_log   /var/log/nginx/rmt_http_error.log;
+    root        /usr/share/rmt/public;
+
```
felixsch: Why not white list the `tools/clientSetup4RMT.sh` only and redirect everything else to https?

public/tools/clientSetup4RMT.sh
ikapelyukhin: Should print the first argument to STDERR if one is present, but yeah -- it seems like it isn't needed, considering that no arguments are allowed.

public/tools/clientSetup4RMT.sh
ikapelyukhin: It's a bit too paranoid for my taste, but this script is borrowed from SMT package with minimal changes (older SLE11 openssl stuff and `suse_register` references removed) -- I don't see a reason to change this.
We are considering either moving this script to SUSEConnect package or implementing a feature in SUSEConnect itself to allow for importing certificates.

public/tools/clientSetup4RMT.sh
ikapelyukhin: Same as above.

package/nginx-http.conf
ikapelyukhin: @felixsch What do you mean specifically? It doesn't do a redirect at the moment, but otherwise that's what it does.

@digitaltom That's up for discussion. SMT does allow access to RPMs over plain HTTP. If you want to learn how many users do that instead of using SUSECoonect -- we could disable access over HTTP and see how many users complain :grimacing:

package/nginx-https.conf
```diff
@@ -0,0 +1,42 @@
...
+    error_log   /var/log/nginx/rmt_https_error.log;
+    root        /usr/share/rmt/public;
+
+    ssl_certificate     /usr/share/rmt/ssl/rmt-server.crt;
+    ssl_certificate_key /usr/share/rmt/ssl/rmt-server.key;
+    ssl_protocols       TLSv1.2 TLSv1.3;
+
+    location / {
+        try_files $uri/index.html $uri.html $uri @rmt_app;
+        autoindex off;
```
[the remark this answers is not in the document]
ikapelyukhin: Actually, I guess it should be off for plain HTTP version. It should list only repos and maybe tools directories.

public/tools/clientSetup4RMT.sh
felixsch: Yes true. Ok than. But its somewhat funny `cp` is checked but `mktemp` is ok too use without checking even its also in `coreutils`. Anyway a connect implementation is preferable I guess.

package/nginx-http.conf
felixsch: @digitaltom actually asked the question behind my statement. If no access to HTTP should be allowed, why not redirect to HTTPS everything except the `clientSetup4RMT.sh`.

ikapelyukhin: @kalabiyau Do you have any thoughts on this? Should the RPMs be accessible over plain HTTP?
SMT serves RPMs over HTTP, at the moment RMT does the same -- only the API is restricted to HTTPS.

package/nginx-https.conf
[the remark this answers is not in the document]
ikapelyukhin: OK, I've disabled autoindex for plain HTTP instead.

package/nginx-http.conf
kalabiyau: @ikapelyukhin as you described to me that atm we have:

registration would by default offer https:// urls
and plain access is allowed to the content via http://

I believe it is a sane setup.

Issue #118, opened 2018-03-21
RMT doesn't clean up temporary directories


ikapelyukhin:
After running RMT for some time:

```
ls -ld /tmp/d20180* | wc -l
7147
```

PR #126, opened 2018-04-04
Don't create a local licenses directory if the repo has no licenses

lagartoflojo:


lib/rmt/mirror.rb
```diff
@@ -128,10 +127,10 @@ def mirror_license
         @downloader.download(filename)
       end
     rescue RMT::Downloader::Exception => e
-      FileUtils.remove_entry(@temp_licenses_dir)
-      @temp_licenses_dir = nil
       raise RMT::Mirror::Exception.new("Error during mirroring metadata: #{e.message}")
     end
+
+    replace_directory(@temp_licenses_dir, File.join(@repository_dir, '../product.license/'))
```
ikapelyukhin: Previously the replacing happened at the very end -- to make sure that you have consistency between metadata and the repo contents.
With this change, when an error occurs during mirroring of RPMs you'll end up with old repo metadata, old RPM and the new product license (which might not be too terrible by itself, but this behavior is inconsistent).

lagartoflojo: I figured it was the easiest way to solve the issue, but you're right that it might create inconsistencies.

My other solution was to do something like

```ruby
def mirror
  create_directories
  license_mirrored = mirror_license # mirror_license returns true or false depending on success

  # (...)
  replace_directory(@temp_licenses_dir, File.join(@repository_dir, '../product.license/')) if license_mirrored
  replace_directory(File.join(@temp_metadata_dir, 'repodata'), File.join(@repository_dir, 'repodata'))
ensure
  remove_tmp_directories
end
```

What do you think?

ikapelyukhin: To address the issue at hand, which is the empty `product.license` directory, I would suggest to make changes here to remove the tempdir: https://github.com/SUSE/rmt/blob/9ba286fb871014db934f67559a7160e952166223/lib/rmt/mirror.rb#L118-L121

And then to skip `replace_directory` if there is no temp directory for licenses.

But refactoring to consolidate all of the temp directory cleanups under `mirror` method is a good idea.

ikapelyukhin:
@lagartoflojo Please also take a look at #118 -- I think the changes that you are making will also resolve this issue.

lib/rmt/mirror.rb
lagartoflojo: Yup that would also work.

lagartoflojo: Updated.
I think my changes also fix #118.

.gitignore
```diff
@@ -33,3 +33,6 @@ config/secrets.yml.enc
 config/rmt.local.yml
 
 .env
+
+public/repo
```
tmuntan1: By doing this, the folder `public/repo` will not be on your system with a `git pull`. Is this what you want?

lagartoflojo: Yeah, I wasn't sure. I wanted to clear all the local mirrored files so I ran `rm -rf public/repo`. And then after re-mirroring, git started showing all the files inside as "untracked".
And you're right, we want to create this dir when pulling. I'll revert this.

tmuntan1: Thanks, I did this because without the folder being there, you start getting errors during mirroring and letting git take care of everything is better than us writing more code.

public/repo/.gitignore
```diff
@@ -1,2 +0,0 @@
-*
-!.gitignore
```
tmuntan1: The gitignore is still deleted here

lagartoflojo: oh i see

PR #129, opened 2018-04-04
Use the recommended systemd config

lagartoflojo:
Puma specifies a recommended way of starting it with systemd at:
https://github.com/puma/puma/blob/5fb2df9144803f0f80ec1d9fb19d20d433706fef/docs/systemd.md#service-configuration

Type=simple can be used, and then there's not need to --daemon the Rails
server.
Additionally, we set the RAILS_LOG_TO_STDOUT variable, so that Rails
outputs its log to stdout. This allows for the logs to be saved and
visible in journalctl.

The PrivateTmp option "sets up a new file system namespace for the
executed processes and mounts private /tmp and /var/tmp directories
inside it that is not shared by processes outside of the namespace.
(...) all temporary files created by a service in these directories will
be removed after the service is stopped."

Since RMT creates some temporary directories while mirroring, with this
option we isolate them and make sure they get deleted when RMT is
stopped.

### DIFF
--- a/package/rmt.service
+++ b/package/rmt.service
@@ -5,12 +5,12 @@ Requires=rmt-migration.service
 After=rmt-migration.service
 
 [Service]
-Type=forking
+Type=simple
 User=_rmt
+PrivateTmp=yes
+Environment=RAILS_LOG_TO_STDOUT=1
 WorkingDirectory=/usr/share/rmt
-ExecStart=/usr/share/rmt/bin/rails server -e production --daemon
-ExecStop=/usr/bin/kill -15 $MAINPID
-PIDFile=/usr/share/rmt/tmp/pids/server.pid
+ExecStart=/usr/share/rmt/bin/rails server -e production
 Restart=always
 
 [Install]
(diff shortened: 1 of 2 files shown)


package/rmt.service
```diff
@@ -5,12 +5,12 @@ Requires=rmt-migration.service
 After=rmt-migration.service
 
 [Service]
-Type=forking
+Type=simple
 User=_rmt
+PrivateTmp=yes
+Environment=RAILS_LOG_TO_STDOUT=1
```
ikapelyukhin: Looks good, but I'd suggest making changes to the RPM package as well -- currently it creates an empty `log` directory to store file logs. It can be dropped now.

lagartoflojo: Does https://github.com/SUSE/rmt/pull/129/commits/923488dd56cdbe7ae462969eff3b3cff1c1c2278 look lood?

ikapelyukhin: Yep! :+1:

lagartoflojo: Cool. I can't merge until we merge the `parser` gem fix xD

PR #132, opened 2018-04-09
Only enable free modules for sles 15 migrations - Card 144

tmuntan1:


spec/models/migration_engine_spec.rb
```diff
@@ -326,6 +326,40 @@
           end
         end
 
+        context 'complex case with SLE 15 module dependencies' do
```
lagartoflojo: Context name is duplicated from https://github.com/SUSE/rmt/pull/132/files#diff-b8fe386eae7950826d27f1d741fe4562L329 no?

app/models/product.rb
```diff
@@ -86,7 +86,7 @@ def available_for
   end
 
   def self.available_or_recommended_modules(root_product_ids)
-    joins(:product_extensions_associations).where(free: true, products_extensions: { root_product_id: root_product_ids }).or(
+    joins(:product_extensions_associations).where(free: true, product_type: :module, products_extensions: { root_product_id: root_product_ids }).or(
```
lagartoflojo: same comment as in https://github.com/SUSE/happy-customer/pull/3705/files#r180096774

PR #134, opened 2018-04-12
Initial contrib directory

felixsch:


ikapelyukhin:
Perhaps it would be better to do it in Ruby instead of bash, especially considering it might be desirable to reuse the SSL certificate generation code?

kalabiyau:
@ikapelyukhin @felixsch @tmuntaner also if done in Ruby it opens up a possibility to actually unit-test the scripts which would be a pain with bash.

contrib/smt-export.sh
```diff
@@ -0,0 +1,180 @@
...
+    warn "Your server SSL certificates have been exported. Make sure to keep the tarball safe!"
+    warn "You can disable the export of SSL certificates by adding --no-ssl-export"
+    warn "to this command."
+
+    # copy ssl files
+    mkdir $temp_ssl_path
+
+    cp $SSL_PATH/openssl.cnf $temp_ssl_path
+    cp $SSL_PATH/ca-bundle.pem $temp_ssl_path
+    cp -r $SSL_PATH/servercerts $temp_ssl_path
```
ikapelyukhin: None of the `openssl.cnf`, `ca-bundle.pem`, `servercerts` should be needed.

contrib/smt-export.sh
```diff
@@ -0,0 +1,180 @@
...
+
+    if [ -f $SSL_PATH/certs/$yast_migration_pem ]; then
+      cp $SSL_PATH/certs/$yast_migration_pem $temp_ssl_path
```
ikapelyukhin: Is that where the Yast CA cert if actually stored? What about the private key?

contrib/smt-to-rmt-helpers/rmt-data-import
```diff
@@ -0,0 +1,91 @@
...
+relative_load_paths = %w[lib lib/rmt app/models app/services].map { |dir| File.join(rmt_path, dir) }
+ActiveSupport::Dependencies.autoload_paths += relative_load_paths
+
+db_config = RMT::Config.db_config
+ActiveRecord::Base.establish_connection(db_config)
+
+#####
+
+# FIXME: need to check that products table isn't empty
+# FIXME wrap everything into a gigantic transaction?
```
fschueller: I prefer the logic broken down into methods tbh. The bash script actually looks cleaner to me :see_no_evil:

fschueller: I prefer when the logic is broken down into methods tbh, currently it's rather hard to read. The bash script actually looks cleaner to me :/

ikapelyukhin: Yep. It also needs test coverage. But it's only PoC, everything will need to be refactored later and the documentation will have to be written by the docu team.

contrib/smt-to-rmt-helpers/README.md
```diff
@@ -0,0 +1,82 @@
...
+    ```
+4. If all went well, the command should output where the tarball has been stored (e.g. `The exported configuration has
+been saved to smt-export.xxxxxx.tar.gz`). Move that file to the RMT
+server or note its path if the same server will be used.
+
+## Importing SMT data to RMT
+
+1. Upload the tarball containing SMT data and `rmt-data-import` script to the RMT server
+1. Unpack the tarball containing SMT data to some directory, e.g. `/root/smt-data`
+1. Decrypt the SMT CA private key to `/usr/share/rmt/ssl/`:
```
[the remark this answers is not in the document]
ikapelyukhin: Added a phrase about it.

contrib/smt-to-rmt-helpers/README.md
```diff
@@ -0,0 +1,82 @@
...
+1. If you chose to export SMT's SSL certificates, copy the SMT CA certificate to `/usr/share/rmt/ssl/`:
+    ```
+    cp /root/smt-data/ssl/cacert.pem /usr/share/rmt/ssl/rmt-ca.crt
+    ```
+1. Run YaST RMT configuration module from YaST command center or by running `yast2 rmt` on the command line;
+1. Go through setup steps of the YaST module. On the SSL setup page it is possible to add alternative common names (e.g., the hostname of the SMT server in case it is desirable to perform a switch) 
+1. Run `rmt-cli sync` to get the products and repositories data from SCC;
+1. Run the `rmt-data-import` to import SMT data:
+    ```
+    ./rmt-data-import /root/smt-data-export/
```
fschueller: This is really just a nitpick: on line 63 the directory we unpacked to was called differently :)

ikapelyukhin: I didn't expect our users to be so attentive!

tmuntaner:
@fschueller I have a feeling that this will change in the future. This was just a proof of concept.

PR #139, opened 2018-04-17
Add RMT header to requests to connect API

Sergeykot:


lib/suse/connect/api.rb
```diff
@@ -47,6 +47,7 @@ def make_request(method, url, options)
         options[:userpwd] = "#{@username}:#{@password}" unless options[:userpwd]
         options[:method] = method
         options[:accept_encoding] = 'gzip, deflate'
+        options[:headers] = { 'RMT' => @username }
```
ikapelyukhin: Aren't those organization credentials? The machine credentials should be read from `/etc/zypp/credentials.d/SCCcredentials`.

Sergeykot: yep, that was exactly what I meant on stand up. many thanks for help @ikapelyukhin :+1:

spec/rails_helper.rb
```diff
@@ -26,6 +26,8 @@
 # If you are not using ActiveRecord, you can remove this line.
 ActiveRecord::Migration.maintain_test_schema!
 
+File.delete(Rails.root.join('config', 'system_uuid')) if File.exist?(Rails.root.join('config', 'system_uuid'))
```
ikapelyukhin: I don't think it is a good idea to manipulate files in your working copy on each test run.
You should mock the file system calls instead.

spec/suse/connect/api_spec.rb
```diff
@@ -77,7 +77,8 @@
   let(:expected_request_headers) do
     {
       'Authorization' => 'Basic ' + Base64.encode64("#{username}:#{password}").strip,
-      'User-Agent' => "RMT/#{RMT::VERSION}"
+      'User-Agent' => "RMT/#{RMT::VERSION}",
+      'RMT' => api_client.send(:system_uuid)
```
ikapelyukhin: Declare some test value in a let block, e.g. `let(:uuid) { 'test-uuid' }` and expect it here: `RMT => uuid`.

spec/suse/connect/api_spec.rb
```diff
@@ -86,41 +87,48 @@
     subject { api_client.send(:make_single_request, 'GET', 'http://example.org/api_method') }
 
     it { is_expected.to eq(response_data) }
+    it { expect(File.read(Rails.root.join('config', 'system_uuid'))).to eq(api_client.send(:system_uuid)) }
```
ikapelyukhin: Mock the `api_client.system_uuid` method call to return `uuid` here:
```
it { expect(api_client).to receive(:system_uuid).and_return(uuid) }
```

lib/suse/connect/api.rb
```diff
@@ -79,6 +83,22 @@ def make_paginated_request(method, url, options = {})
         @entities
       end
 
+      private
+
+      def system_uuid
+        @system_uuid ||= fetch_system_uuid
```
ikapelyukhin: It would be better to merge this method with `fetch_system_uuid`, there shouldn't be a need for a 1-line method.

Sergeykot:
@ikapelyukhin please take a look if we should go with those specs. also regarding testing the 'system_uuid' - it is a private method which is usually not tested. also I think I covered it in the shared examples: https://github.com/SUSE/rmt/pull/139/files#diff-d6fa49050fca4a3f8d1385c1547fae59R90
what do you think?

ikapelyukhin:
@Sergeykot 

1. The project has 100% test coverage -- everything should be covered, including private methods :grimacing: (tests under `describe #system_uuid` are needed)
2. Right now you are executing the same shared example test over and over again in `describe #make_paginated_request` with the same parameters.
    
    In those tests you supposed to test `#make_paginated_request` and not `#system_uuid`. This is why in `describe #make_paginated_request` the `system_uuid` method should be mocked.
    
    Also calling the same test on `#system_uuid` multiple times doesn't really improve testing -- on the contrary, it will just make the test suite slower to run.
3. Shared examples serve a different purpose -- they exist to run same examples on different methods/classes/controllers/etc. Using them to run the same test on the same method doesn't seem like a good idea to me.

lib/suse/connect/api.rb
```diff
@@ -7,30 +7,34 @@ module Connect
     class Api
 
       class InvalidCredentialsError < StandardError; end
+      SYSTEM_UUID_LOCATION = 'config/system_uuid'.freeze
+      ROOT_PATH = Pathname.new(File.dirname(__FILE__)).join('../', '../', '../').freeze
+      CONNECT_API_URL = 'https://scc.suse.com/connect'.freeze
+      UUID_FILE_LOCATION = ROOT_PATH.join(SYSTEM_UUID_LOCATION).freeze
```
thutterer: Why create an extra file and all the boilerplate? We could just store it in the existing config file without any of this.
This would also allow the customer to set it to something more meaningful than a randomly generated string.

And why do we even need to generate a new identifier?
We could just use the fqdn, couldn't we? Since we have the SSL functionality, each RMT has one setup anyway, right? @ikapelyukhin

ikapelyukhin: @thutterer 
1. What's the relation between FQDN and SSL?
2. In regards to the value of RMT identifier, it depends on whether we want the user to change it or not. I was under the assumption that we want it to be as static as possible -- a separate file would survive DNS hostname changes and RPM package uninstallation/reinstallation.
    
    Making it changeable presents another issue -- once you change the ID, the old IDs will remain in SCC. Perhaps it would be better to assign a custom label to a uniquely-identified RMT in SCC, instead of changing the unique ID on RMT side?

thutterer: Don't you have to provide a hostname when generating a certificate for an RMT? So by design, each RMT has a hostname, that was set by its administrator. And if I put myself in that persons shoes, I would like to find that same RMT under that same name in my web interface SCC.

Because if you, let's say, setup 3 RMTs for your organization and sync them all with SCC at roughly the same time, you would see 3 new RMTs there, all with a unique Id, but you would still have no idea, which is which, because these Ids do not mean anything to you, and you have no good way to understand or change them.

ikapelyukhin: The FQDN can be obtained from the host system just by running `hostname`/`hostname --long`. FQDN doesn't have anything to do with SSL. Every host can have multiple hostnames.

Anyway, I don't have any strong preference of what the identifier should be.

digitaltom: But if the RMT is registered itself (it should be in a supported scenario), the `SCC_` id will lead to its system + hwinfo which already holds the hostname.

ikapelyukhin: @digitaltom `SCCcredentials` isn't readable by `_rmt` user. Not without some sort of service startup hack, anyway.

Sergeykot: FYI: the identifier has to be an UUID or an UUID with `SCC_` prefix. see
https://github.com/SUSE/happy-customer/blob/f756108955e53f6fe546ddf1b0de78ebdc24ec8c/glue/app/controllers/api/connect/v3/organizations/base_controller.rb#L8
https://github.com/SUSE/happy-customer/blob/f756108955e53f6fe546ddf1b0de78ebdc24ec8c/glue/app/models/registration_proxy.rb#L29

Sergeykot: otherwise we will probably have to make changes (possibly harmful (?)) to the code which handles the header in 'glue'

Sergeykot: however what about adding a field 'fdqn' to `RegistrationProxy` in glue? and we could pass it in a separate header and show it to customer @ikapelyukhin @digitaltom @thutterer 
could probably be in a separate ticket thoough..

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,166 @@
...
-    subject { api_client.list_orders }
-
-    it { is_expected.to eq([ { endpoint: 'organizations/orders' } ]) }
+  let(:uuid) { 'test-uuid' }
+
+  describe '#system_uuid' do
+    subject(:method_call) { api_client.send(:system_uuid) }
+
+    before do
+      allow(SecureRandom).to receive(:uuid).and_return(uuid)
```
ikapelyukhin: Why is it always expected? It shouldn't be called if the file already exists, the spec should be failing.

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,166 @@
...
+
+    context 'when system_uuid file exists' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(true)
+      end
+
+      it 'reads a file' do
+        expect(File).not_to receive(:write).with(described_class::UUID_FILE_LOCATION, uuid)
+        expect(File).to receive(:read).with(described_class::UUID_FILE_LOCATION).and_return(uuid)
+        expect_any_instance_of(described_class).to receive(:system_uuid).and_call_original
```
ikapelyukhin: Hmm? Why does the spec for `system_uuid` expect that `system_uuid` will be called? Also, why call the original implementation?

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,166 @@
...
+
+    context 'when system_uuid file does not exist' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(false)
+      end
+
+      it 'creates a file' do
+        expect(File).to receive(:write).with(described_class::UUID_FILE_LOCATION, uuid).exactly(1).times
+        expect(File).not_to receive(:read).with(described_class::UUID_FILE_LOCATION)
+        expect_any_instance_of(described_class).to receive(:system_uuid).and_call_original
```
ikapelyukhin: Same as above.

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,161 @@
...
-
-    it { is_expected.to eq([ { endpoint: 'organizations/orders' } ]) }
+  let(:uuid) { 'test-uuid' }
+
+  describe '#system_uuid' do
+    subject(:method_call) { api_client.send(:system_uuid) }
+
+    context 'when system_uuid file exists' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(true)
```
ikapelyukhin: There's only one example in the context, the `before` block can be merged with the example.

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,161 @@
...
+        expect(File).not_to receive(:write).with(described_class::UUID_FILE_LOCATION, uuid)
+        expect(File).to receive(:read).with(described_class::UUID_FILE_LOCATION).and_return(uuid)
+        method_call
+      end
+    end
+
+    context 'when system_uuid file does not exist' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(false)
+        allow(SecureRandom).to receive(:uuid).and_return(uuid)
```
ikapelyukhin: There's only one example in the context, the `before` block can be merged with the example.

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,161 @@
...
+
+    context 'when system_uuid file exists' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(true)
+      end
+
+      it 'reads a file' do
+        expect(File).not_to receive(:write).with(described_class::UUID_FILE_LOCATION, uuid)
+        expect(File).to receive(:read).with(described_class::UUID_FILE_LOCATION).and_return(uuid)
+        method_call
```
ikapelyukhin: `expect(method_call).to be(uuid)`, to make sure that the method actually returns the UUID.

spec/suse/connect/api_spec.rb
```diff
@@ -2,125 +2,161 @@
...
+    context 'when system_uuid file does not exist' do
+      before do
+        allow(File).to receive(:exist?).with(described_class::UUID_FILE_LOCATION).and_return(false)
+        allow(SecureRandom).to receive(:uuid).and_return(uuid)
+      end
+
+      it 'creates a file' do
+        expect(File).to receive(:write).with(described_class::UUID_FILE_LOCATION, uuid).exactly(1).times
+        expect(File).not_to receive(:read).with(described_class::UUID_FILE_LOCATION)
+        method_call
```
ikapelyukhin: `expect(method_call).to be(uuid)`, to make sure that the method actually returns the UUID.

Sergeykot:
@ikapelyukhin I addressed the latest comments. please have a look. will squash smaller commits after approval - in order to not to mess up discussion threads.

lib/suse/connect/api.rb
ikapelyukhin: `UUID_FILE_LOCATION = File.expand_path('../../../config/system_uuid', __dir__)`.
`SYSTEM_UUID_LOCATION` and `ROOT_PATH` aren't used anywhere and shouldn't be needed.

PR #141, opened 2018-04-20
Enable Recommended Products - Card 170

tmuntaner:
**Who:** RMT users
**What:** need to have mirroring enabled for recommended modules when enabling mirroring for the base product
**Why**: currently if you could do the following:

* enable `sles/15/x86_64` 
* do not enable `basesystem` or any other module
* register your client machine
* have only 3 RPM packages base product repos
* have no extensions will be listed in the output of `SUSEConnect --list-extensions`

### DIFF
--- a/spec/lib/rmt/cli/products_spec.rb
+++ b/spec/lib/rmt/cli/products_spec.rb
@@ -139,7 +139,7 @@ RSpec.describe RMT::CLI::Products do
           "#{repo_count} repo(s) successfully enabled.\n"
         end
 
-        it 'enables the mandatory product repositories' do
+        it 'enables product and recommended products repositories' do
           products.flat_map(&:repositories).each do |repository|
             expect(repository.mirroring_enabled).to eq(repository.enabled)
           end
@@ -203,7 +203,7 @@ RSpec.describe RMT::CLI::Products do
         let(:product) { create :product, :with_mirrored_repositories, :with_recommended_extensions }
         let(:extensions) { Product.recommended_extensions(product).to_a }
 
-        it 'disabled the mandatory product repositories' do
+        it 'does not disable extension repositories' do
           product.repositories.each do |repository|
             expect(repository.mirroring_enabled).to eq(false)
           end
--- a/spec/lib/rmt/cli/products_spec.rb
+++ b/spec/lib/rmt/cli/products_spec.rb
@@ -129,7 +129,7 @@ RSpec.describe RMT::CLI::Products do
         end
       end
 
-      context 'with recommended products' do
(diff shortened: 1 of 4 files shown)


ikapelyukhin:
Should there a more informative message be displayed, explaining that not only the product that you have requested, but also the modules that are required for this product to work?

(Also, linking to the private stuff in public project :confused:)

tmuntaner:
@ikapelyukhin I updated the branch with your requested changes.

ikapelyukhin:
```
rmt-cli products disable sles/15/x86_64
The following required extensions for SLES/15/x86_64 have been enabled: Basesystem Module, Server Applications Module.
6 repo(s) successfully disabled.
```

Uh-oh. It probably shouldn't be disabling the extra modules.

tmuntaner:
@ikapelyukhin good catch. I pushed a fix and added a better test to check disabling with recommended extensions.

spec/lib/rmt/cli/products_spec.rb
```diff
@@ -198,6 +198,22 @@
           expect(repository.mirroring_enabled).to eq(false)
         end
       end
+
+      context 'with recommended extensions' do
+        let(:product) { create :product, :with_mirrored_repositories, :with_recommended_extensions }
+        let(:extensions) { Product.recommended_extensions(product).to_a }
+
+        it 'disabled the mandatory product repositories' do
```
ikapelyukhin: I think it would disable all of the product repos (not just the mandatory ones). I see that the example that we've previously had has the same text, but it's misleading.

spec/lib/rmt/cli/products_spec.rb
```diff
@@ -176,6 +198,22 @@
...
+        let(:product) { create :product, :with_mirrored_repositories, :with_recommended_extensions }
+        let(:extensions) { Product.recommended_extensions(product).to_a }
+
+        it 'does not disable extension repositories' do
+          product.repositories.each do |repository|
+            expect(repository.mirroring_enabled).to eq(false)
+          end
+          extensions.each do |extension|
+            extension.repositories.each do |repository|
+              expect(repository.mirroring_enabled).to eq(false)
```
ikapelyukhin: Although wait, what does this check?

ikapelyukhin: The `with_recommended_extensions` creates extensions with `with_not_mirrored_repositories` -- so mirroring is always disabled for those repos. I'm not sure what this example tries to check.

tmuntaner:
@ikapelyukhin thanks

PR #142, opened 2018-04-20
Remove uniqueness for HwInfo#uuid

Sergeykot:
The unique index on 'HwInfo#uuid' breaks the behaviour in case of cleanup and attempt to register again. There should be no unique index

The solution is to update the migration and fix it on ref host manually

### DIFF
--- a/app/models/hw_info.rb
+++ b/app/models/hw_info.rb
@@ -5,7 +5,7 @@ class HwInfo < ApplicationRecord
   before_validation :make_invalid_uuid_nil
 
   # We store UUID as a downcased string. Please take that in account in finders
-  validates :uuid, uuid_format: true, uniqueness: { allow_nil: true }
+  validates :uuid, uuid_format: true
   validates :system, uniqueness: true, presence: true
 
   before_save -> { uuid.try(:downcase!) }
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180416124217) do
+ActiveRecord::Schema.define(version: 20180420145408) do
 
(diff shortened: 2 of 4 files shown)


db/schema.rb
```diff
@@ -41,7 +41,7 @@
     t.datetime "updated_at", null: false
     t.index ["hypervisor"], name: "index_hw_infos_on_hypervisor"
     t.index ["system_id"], name: "index_hw_infos_on_system_id", unique: true
-    t.index ["uuid"], name: "index_hw_infos_on_uuid", unique: true
+    t.index ["uuid"], name: "index_hw_infos_on_uuid"
```
Sergeykot: I am thinking of doing this in a separate migration.. :thinking:

db/migrate/20180416124217_create_hw_infos.rb
```diff
@@ -12,6 +12,6 @@ def change
     end
 
     add_index :hw_infos, :system_id, unique: true
-    add_index :hw_infos, :uuid, unique: true
+    add_index :hw_infos, :uuid
```
ikapelyukhin: I don't think we need the index at all (nobody's going to query the DB with UUID condition).

db/schema.rb
ikapelyukhin: OK, drop the index in a separate migration then.

Sergeykot: should be better like that. done

app/models/hw_info.rb
```diff
@@ -5,7 +5,7 @@ class HwInfo < ApplicationRecord
   before_validation :make_invalid_uuid_nil
 
   # We store UUID as a downcased string. Please take that in account in finders
-  validates :uuid, uuid_format: true, uniqueness: { allow_nil: true }
+  validates :uuid, uuid_format: true
```
ikapelyukhin: This migration should remain unchanged, then.

ikapelyukhin: Ah, this is not the migration. Neverime! :grimacing:

PR #145, opened 2018-04-23
Change log formatting

Sergeykot:


wstephenson:
Something is not working with eliding the rails logger timestamp for journald logging on the mirror service:  
```
-- Logs begin at Wed 2018-04-25 06:36:27 CEST, end at Thu 2018-04-26 07:31:42 CEST. --
Apr 26 07:29:40 e178 systemd[1]: Started Mirror RMT.
Apr 26 07:29:56 e178 rmt-cli[29450]: Mirroring repository SLE-Product-SLES15-Pool to /usr/share/rmt/public/repo
Apr 26 07:29:56 e178 rmt-cli[29450]: I, [2018-04-26T07:29:42.901271 #29450]  INFO -- : ↓ directory.yast
Apr 26 07:29:56 e178 rmt-cli[29450]: I, [2018-04-26T07:29:43.092716 #29450]  INFO -- : ↓ license.txt
Apr 26 07:29:56 e178 rmt-cli[29450]: I, [2018-04-26T07:29:43.273965 #29450]  INFO -- : ↓ repomd.xml
```

package/rmt-migration.service
```diff
@@ -7,6 +7,7 @@ Before=rmt.service
 [Service]
 Type=oneshot
 User=_rmt
+Environment=RAILS_LOG_TO_JOURNALD=1
```
Sergeykot: should I rename it to something like `LOG_TO_JOURNALD`? - since CLI does not really init rails

lib/rmt/downloader.rb
```diff
@@ -21,6 +21,13 @@ def initialize(repository_url:, destination_dir:, auth_token: nil, logger: nil,
     @concurrency = 4
     @auth_token = auth_token
     @logger = logger || Logger.new('/dev/null')
+    if ENV['RAILS_LOG_TO_JOURNALD'].present?
```
Sergeykot: maybe I could put the code somewhere else to make it DRY-ish ?

Sergeykot:
@wstephenson I made fixes for services. the reason was that the CLI does not really init Rails so that the stuff from 'config/environments/production.rb' is not executed. so that I had to change this in the 'lib' folder.

FYI: I updated the previous commit to not make too many of them in the history

package/rmt-migration.service
wstephenson: Good idea

package/rmt-server-mirror.service
```diff
@@ -6,6 +6,7 @@ Requires=rmt.service
 [Service]
 Type=simple
 Restart=no
+Environment=RAILS_LOG_TO_JOURNALD=1
 ExecStart=/usr/share/rmt/bin/rmt-cli mirror
```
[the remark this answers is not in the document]
wstephenson: Ignore, I didn't post this comment when I made it this morning, but you caught the cause anyway :).

lib/rmt/downloader.rb
wstephenson: Add an accessor in RMT::Config? Doesn't need to be written to the config file though.

config/environments/production.rb
```diff
@@ -38,14 +38,15 @@
...
 
   # Use a different logger for distributed setups.
   # require 'syslog/logger'
   # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
 
-  if ENV['RAILS_LOG_TO_STDOUT'].present?
+  if ENV['RAILS_LOG_TO_JOURNALD'].present?
     logger           = ActiveSupport::Logger.new(STDOUT)
     logger.formatter = config.log_formatter
     config.logger    = ActiveSupport::TaggedLogging.new(logger)
```
ikapelyukhin: Perhaps create a new class (like `RMT::Logger`) and move all this boilerplate logger initialization code (ENV var check, format setting, etc.) into a constructor, and then use this class everywhere?

package/rmt-migration.service
tmuntaner: @Sergeykot `LOG_TO_JOURNALD` doesn't work if you consider that you can run these commands in the CLI.

tmuntaner: Ignore my previous comment

lib/rmt/downloader.rb
Sergeykot: it should be more clean to make a logger class by advice from @ikapelyukhin

config/environments/production.rb
Sergeykot: I think since we don't have `ActiveSupport` in the CLI (do we?) I have to make it only in context of the CLI code, right? @ikapelyukhin

ikapelyukhin: We have `ActiveSupport`: https://github.com/SUSE/rmt/blob/54170ef0d7ce5c969de8dabbf899b672db95ad70/bin/rmt-cli#L9

But you'll have to see if it is possible to just replace default Ruby logger with `ActiveSupport::Logger`.
I think it would be better to have the same logger for everything.

Sergeykot:
@tmuntaner I fixed logging for `mirror` command. please take a look

spec/lib/rmt/logger_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    context 'without env variable' do
+      before do
+        ENV['LOG_TO_JOURNALD'] = nil
+      end
+
+      it do
+        Timecop.freeze(Time.current) do
+          logger.info('42')
+          log.rewind
+          expect(log.read).to eq("I, [#{Time.now.getlocal('+02:00').strftime('%Y-%m-%dT%H:%M:%S.%6N')} ##{Process.pid}]  INFO -- : 42\n")
```
Sergeykot: I did not find a better way other that hardcode the TZ here.. :neutral_face:

tmuntaner:
@Sergeykot your tests aren't working ^_^

PR #146, opened 2018-04-24
Supportconfig plugin

wstephenson:


support/rmt
```diff
@@ -0,0 +1,75 @@
...
+    echo "package $THISRPM is not installed"
+    echo "Status: Skipped"
+  fi
+  echo
+}
+
+RPMLIST="
+rmt-server
+nginx
+mariadb
```
ikapelyukhin: I would suggest adding `openssl`, `libssl` and `libcurl` -- sometimes the customers mess with those and that causes various obscure issues.

Hmm, and maybe getting vendor information somehow would be good too, in case the package is installed from 3rd-party repos.

support/rmt
```diff
@@ -0,0 +1,75 @@
...
+"
+SERVICES="rmt rmt-server-sync rmt-server-mirror mariadb nginx"
```
ikapelyukhin: Shouldn't this check for timers instead? `rmt-server-sync.timer` and `rmt-server-mirror.timer`. You could actually get statuses of both timers and services.

support/rmt
```diff
@@ -0,0 +1,75 @@
...
+plugin_tag "Packages"
+for THISRPM in $RPMLIST; do
+  validate_rpm_if_installed "$THISRPM"
+done
+
+plugin_tag "Configuration"
+pconf_files "$CONF_FILES"
+
+plugin_tag "SSL Configuration"
+plugin_command "echo | openssl s_client -showcerts -servername localhost -connect localhost:443 2>/dev/null | openssl x509 -inform pem -noout -text"
```
ikapelyukhin: On a second though, this assumes that nginx is running. Maybe listing `/usr/share/nginx/ssl/` and getting the CA and server certs from there would be good too.

support/rmt
```diff
@@ -0,0 +1,75 @@
...
+done
+
+plugin_tag "Mirroring Status"
+plugin_command "rmt-cli products list"
+plugin_command "rmt-cli repos list"
+plugin_command "rmt-cli repos custom list"
+
+plugin_tag "Service Logs"
+for service in $SERVICES; do
+  plugin_command "journalctl -u $service"
```
ikapelyukhin: Perhaps we should set some limit. I think SMT supportconfig modules grabs only 500 last lines from the log (I guess due to size limitations).

support/rmt
felixsch: Maybe `rpm -q --whatprovides ssl`?

support/rmt
```diff
@@ -0,0 +1,75 @@
...
+      echo "Status: WARNING"
+    fi
+  else
+    echo "package $THISRPM is not installed"
+    echo "Status: Skipped"
+  fi
+  echo
+}
+
+RPMLIST="
```
felixsch: Why not putting them in one line or even more modern using array notation to make clear this is an array and not a list. Wondering why this is working anyway since IFS should be set to whitespace.

support/rmt
```diff
@@ -0,0 +1,75 @@
...
+mariadb
+"
+SERVICES="rmt rmt-server-sync rmt-server-mirror mariadb nginx"
+
+CONF_FILES="/etc/rmt.conf /etc/nginx/vhosts.d/rmt-server-http.conf /etc/nginx/vhosts.d/rmt-server-https.conf"
+
+section_header "Supportconfig Plugin for RMT, v${SVER}"
+if ! rpm -q rmt-server &>/dev/null; then
+  echo "ERROR: RMT rpm package ('rmt-server') not installed"
+  echo
```
felixsch: I don't know if this is confusing for people. I always try to avoid echo without arguments. Using `echo ""` or even `echo -e "..\n".

support/rmt
wstephenson: Ok, I just copied this behaviour from susemanager,

support/rmt
wstephenson: Noted

support/rmt
wstephenson: @ikapelyukhin rpm version and vendor information for all rpms is already included by the default supportconfig queries.
```#==[ Command ]======================================#
# rpm -qa --queryformat "%-35{NAME} %-35{DISTRIBUTION} %{VERSION}-%{RELEASE}\n" | sort -k 1,2 -t " " -i
NAME                                DISTRIBUTION                        VERSION
...
libopenssl1_0_0                     SUSE Linux Enterprise 15            1.0.2n-1.19
libopenssl1_0_0-32bit               SUSE Linux Enterprise 12            1.0.2j-60.20.2
libopenssl1_1                       SUSE Linux Enterprise 15            1.1.0g-1.5
libopenssl1_1-32bit                 SUSE Linux Enterprise 15            1.1.0g-1.5
```
as well as the installed products and their services and repositories.

wstephenson: The real danger would be people dumping unpackaged openssl binaries somewhere, but supportconfig should already catch those by dumping the output of /usr/bin/env to env.txt.

support/rmt
wstephenson: Ah, I forgot we discussed that before standup.

support/rmt
ikapelyukhin: Maybe there's some limit? I'm not sure how journald takes care of the old logs.

support/rmt
ikapelyukhin: The real danger would be people building libcurl/libssl from source without touching the libs installed from RPMs and messing with `LD_LIBRARY_PATH` :grimacing:
But that's beyond the point. Verifying files in the RPMs and having vendor info should be good enough for most cases :+1:

wstephenson: You wouldn't believe it but one of our tricky SUSEConnect bugs was due to a Novell groupwise server installation script futzing with the system openssl bugs in precisely that way. The support engineer didn't look for it as craziness in groupwise packages was 'beyond the point'.

tmuntaner:
@wstephenson Could you please add a changelog entry before you merge?

PR #151, opened 2018-04-26
Rename organization_products export to organizations_products_unscoped

felixsch:


lib/rmt/scc.rb
```diff
@@ -42,7 +42,7 @@ def export(path)
     @logger.info('Exporting products')
     File.write(File.join(path, 'organizations_products_scoped.json'), scc_api_client.list_products.to_json)
```
ikapelyukhin: I think it was named `organizations_products.json` originally -- I guess we can rename it back to reflect the API endpoint name.

spec/lib/rmt/scc_spec.rb
```diff
@@ -309,7 +309,7 @@ def scc
       before do
         FakeFS.with_fresh do
           FileUtils.mkdir_p(path)
-          File.write(File.join(path, 'organizations_products_scoped.json'), products.to_json)
+          File.write(File.join(path, 'organizations_products.json'), products.to_json)
```
tmuntaner: Do you mind adding `organizations_products_unscoped.json` and  `organizations_orders.json` here as well?

PR #153, opened 2018-04-27
Add lockfile for mirror, sync & import commands

Sergeykot:

PR #157, opened 2018-05-07
Improve lockfile feature

Sergeykot:
Addressing comments on #153 

* move catching`Execution locked` error into `RMT::CLI::Base#handle_exceptions`
* raise `RMT::CLI::Error` to set proper exit code
* print location of the lockfile in the output
* catch all exceptions inherited from `Exception` and removing lockfile

This makes lockfile to be removed on interruption with Ctrl+C

Also regarding using `File#flock`:

1. The function is not supported in some filesystems. For example; NFS.
2. The function is not supported in `FakeFS` which we use for testing. It would be tricky to rebuild the specs.


lib/rmt/cli/base.rb
```diff
@@ -67,6 +67,14 @@ def handle_exceptions
         "The SCC credentials are not configured correctly in '/etc/rmt.conf'. You can obtain them from https://scc.suse.com/organization",
         RMT::CLI::Error::ERROR_SCC
       )
+    rescue RMT::Lockfile::ExecutionLockedError
+      raise RMT::CLI::Error.new(
+        "Process is locked. Please check lockfile at #{RMT::Lockfile::LOCKFILE_LOCATION}\n",
+        RMT::CLI::Error::ERROR_OTHER
+      )
+    rescue Exception # rubocop:disable RSpec/NestedGroups
+      RMT::Lockfile.remove_file
```
wstephenson: Is this path covered by the specs?

ikapelyukhin: Why are you catching all exceptions here? This doesn't seem right.

Sergeykot: @wstephenson this is covered here https://github.com/SUSE/rmt/blob/78ceb1db9ea029247fc4a048a5d76de8bd0375cc/spec/lib/rmt/cli/main_spec.rb#L92

@ikapelyukhin if we don't catch all exceptions and remove lockfile it will stay if we have any exception that we don't catch. also I re-raise the original exception again after removing lockfile

ikapelyukhin: @Sergeykot see the previous PR and use the `ensure` block for removing the lock files.

Sergeykot: @ikapelyukhin in the 'ensure' block we unfortunately can not check for the exception class so that the lockfile would be removed even if we hit `RMT::Lockfile::ExecutionBlockedError`

ikapelyukhin: In the `ensure` block you don't need to check anything.

This `remove_file` call needs to be moved to `ensure` block around `locked_method` method  -- this will ensure that the lockfile is always removed, even when there's an exception raised:

https://github.com/SUSE/rmt/blob/0a7cd6660bccc728983c0b2cedce4c6343d26f2e/lib/rmt/cli/base.rb#L88-L91

Any other `remove_file` calls should be removed, like this one:

https://github.com/SUSE/rmt/blob/0a7cd6660bccc728983c0b2cedce4c6343d26f2e/lib/rmt/cli/base.rb#L94-L97

Or this one:

https://github.com/SUSE/rmt/blob/78ceb1db9ea029247fc4a048a5d76de8bd0375cc/lib/rmt/cli/base.rb#L75-L77

lib/rmt/cli/base.rb
```diff
@@ -67,6 +67,11 @@ def handle_exceptions
         "The SCC credentials are not configured correctly in '/etc/rmt.conf'. You can obtain them from https://scc.suse.com/organization",
         RMT::CLI::Error::ERROR_SCC
       )
+    rescue RMT::Lockfile::ExecutionLockedError
+      raise RMT::CLI::Error.new(
+        "Process is locked. Please check lockfile at #{RMT::Lockfile::LOCKFILE_LOCATION}\n",
+        RMT::CLI::Error::ERROR_OTHER
```
lagartoflojo: This is not a full review, just a question about this message.
It says "check lockfile at (...)". What is a user expected to do with the lock file? How is it helpful to check that file?

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,13 @@ class RMT::Lockfile
...
-      else
-        dirname = File.dirname(LOCKFILE_LOCATION)
-        FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
-        File.open(LOCKFILE_LOCATION, 'w') { |f| f.write(Process.pid) }
-      end
+      dirname = File.dirname(LOCKFILE_LOCATION)
+      FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
+
+      f = File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT)
+      Timeout.timeout(0.001) { f.flock(File::LOCK_EX) }
```
ikapelyukhin: Timeout is not needed. See the previous PR comments -- use `File::LOCK_EX | File::LOCK_NB` mode.
From the docu:
```
Returns false if File::LOCK_NB is specified and the operation would otherwise have blocked.
```

lib/rmt/cli/base.rb
lagartoflojo: If that lock file exists and there's no RMT command running, we have big trouble...
When zypper is locked, it looks like this, which I think is more useful:

```
> sudo zypper in xfce
System management is locked by the application with pid 6096 (/usr/bin/ruby).
Close this application before trying again.
```

We could say "Close this application or wait for it to finish before trying again" instead?

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,11 @@ class RMT::Lockfile
...
-        raise RMT::Lockfile::ExecutionLockedError
-      else
-        dirname = File.dirname(LOCKFILE_LOCATION)
-        FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
-        File.open(LOCKFILE_LOCATION, 'w') { |f| f.write(Process.pid) }
-      end
+      dirname = File.dirname(LOCKFILE_LOCATION)
+      FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
+
+      f = File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT)
```
ikapelyukhin: Once the method returns, `f` isn't referenced anymore and gets freed by the garbage collector, meaning that the file is closed and the lock is released.

lib/rmt/cli/base.rb
Sergeykot: :+1: updated. new messages looks like:
'Process is locked by the application with pid 19032. Close this application or wait for it to finish before trying again'

Sergeykot:
@ikapelyukhin I hope I fixed the GC-related issue. please check it because I was not able to reproduce

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,13 @@ class RMT::Lockfile
...
   class << self
     def create_file
-      if File.exist?(LOCKFILE_LOCATION)
-        raise RMT::Lockfile::ExecutionLockedError
-      else
-        dirname = File.dirname(LOCKFILE_LOCATION)
-        FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
-        File.open(LOCKFILE_LOCATION, 'w') { |f| f.write(Process.pid) }
-      end
+      dirname = File.dirname(LOCKFILE_LOCATION)
```
thutterer: Why so complicated?
Why a relative path in `LOCKFILE_LOCATION`? Lockfiles have to go to `/var/lock` which nowadays in most distributions (including openSUSE and SLES) is just a link to `/run/lock`. Please double-check if I'm right. Where does for example zypper put its lockfile? But I'm sure you can make the code simpler without the `../../`

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,13 @@ class RMT::Lockfile
...
+      FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
+
+      # has to be a class instance variable otherwise garbage-collected and lock is removed
+      @f = File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT)
```
thutterer: Can you please provide a comment or link that explains this lines. It's far from obvious to me.

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,13 @@ class RMT::Lockfile
...
+      @f.write(Process.pid.to_s) if @f.read.empty? # write PID if file is new. this has to be done before locking
```
thutterer: Here is some twisted logic I don't get:
Why check if the file was empty?
Afaik, you are using the flock mechanism to determine if another running instance has the lock on this file.

So why not just try to grab the lock, and if you get it, you write the pid into that file.
If the file is locked already, the exception is raised.

Because now, if you for some unexpected reason, already have a lock file with some random number in it, but no other flock on that file, when you start the mirror command, it will work and start mirroring.
If you start another instance of mirror, it will stop, because of the locked file, but it will tell you to kill the process with that random number, not the actual pid of the first mirror.

... I just tried it and ran into the same problem than what you commented on ("this has to be done before locking") ... Do you know why this is?

Is it maybe related to the not so nice work-around with the instance variable??
Dunno, but I don't like that construct too much anyway.

Why not do this more OO? You could have a RMT::Lockfile `instanciate` method which returns the file object. This object you can store in a variable (when calling it from the `locked_method` wrapper method). And when you give that class an instance method `lock!`, it would all read so much nicer imo: `RMT::Lockfile.new.lock!`

Of course these Lockfile instances also could have a `remove!` method.

Maybe by refactoring it like this, you can then write into that file while it is locked??
In it's current form, I couldn't figure out, why it is not working :thinking:

spec/lib/rmt/cli/import_spec.rb
```diff
@@ -18,13 +19,19 @@
     end
 
     context 'with existing lockfile' do
-      before { allow(RMT::Lockfile).to receive(:create_file).and_raise(RMT::Lockfile::ExecutionLockedError) }
+      before do
+        allow(RMT::Lockfile).to receive(:create_file).and_raise(RMT::Lockfile::ExecutionLockedError)
```
thutterer: Is it possible to really create a lockfile? So that it actually executes the real check if a file is locked or not?
Just raising this execption skippes most of the code that we should test.

spec/lib/rmt/cli/import_spec.rb
```diff
@@ -18,13 +19,19 @@
...
+      end
 
       it 'handles lockfile exception' do
         FakeFS.with_fresh do
           FileUtils.mkdir_p path
 
-          expect { command }.to output(/Process is locked/).to_stdout.and output('').to_stderr
+          expect(described_class).to receive(:exit)
+          expect(File).to receive(:read).with(RMT::Lockfile::LOCKFILE_LOCATION).and_return(pid)
```
thutterer: Same as with the other spec: If possible, create a real lockfile and test with that!

spec/lib/rmt/cli/import_spec.rb
```diff
@@ -117,14 +124,20 @@
 
 
     context 'with existing lockfile' do
-      before { allow(RMT::Lockfile).to receive(:create_file).and_raise(RMT::Lockfile::ExecutionLockedError) }
+      before do
+        allow(RMT::Lockfile).to receive(:create_file).and_raise(RMT::Lockfile::ExecutionLockedError)
```
thutterer: Same as with the other spec: If possible, create a real lockfile and test with that!

spec/lib/rmt/cli/import_spec.rb
```diff
@@ -117,14 +124,20 @@
...
+      end
 
       it 'handles lockfile exception' do
         FakeFS.with_fresh do
           FileUtils.mkdir_p path
           File.write("#{path}/repos.json", repo_settings.to_json)
 
-          expect { command }.to output(/Process is locked/).to_stdout.and output('').to_stderr
+          expect(described_class).to receive(:exit)
+          expect(File).to receive(:read).with(RMT::Lockfile::LOCKFILE_LOCATION).and_return(pid)
```
thutterer: You have a lot of duplicated or very similar lines in these new specs. Please try to join them in an shared_example, if possible.

spec/lib/rmt/cli/main_spec.rb
```diff
@@ -76,10 +84,14 @@
...
           allow(RMT::Lockfile).to receive(:create_file).and_raise(RMT::Lockfile::ExecutionLockedError)
+          allow(File).to receive(:read).with(RMT::Lockfile::LOCKFILE_LOCATION).and_return(pid)
         end
 
         it do
-          expect { command }.to output(/Process is locked\n/).to_stdout
+          expect(described_class).to receive(:exit)
+          expect { command }.to output(
+            "Process is locked by the application with pid #{pid}. Close this application or wait for it to finish before trying again\n"
+          ).to_stderr
```
thutterer: this it block is a 1:1 duplicate with the one above :arrow_right: shared example

lib/rmt/lockfile.rb
lagartoflojo: I have a similar concern regarding the usage of the lockfile.
Right now there's this `locked_method` call in `lib/rmt/cli/base.rb`, which in turn calls `RMT::Lockfile.create_file` and `RMT::Lockfile.remove_file`.

Why not get rid of that, and instead move it INTO `RMT::Lockfile`?

Then in `lib/rmt/cli/main.rb` and `lib/rmt/cli/import.rb` you could instead do:

```ruby
  desc 'data PATH', 'Read SCC data from given path'
  def data(path)
    RMT::Lockfile.lock do # <===
      needs_path(path) do
        RMT::SCC.new(options).import(path)
      end
    end
  end
```

Then, within `RMT::Lockfile` you can also simply use `File.open do ...`, so the the file closes automatically at the end.

lib/rmt/lockfile.rb
```diff
@@ -4,13 +4,13 @@ class RMT::Lockfile
...
+      raise ExecutionLockedError unless @f.flock(File::LOCK_EX | File::LOCK_NB)
     end
 
     def remove_file
```
lagartoflojo: Is removing the file enough? AFAIK, once you lock a file, you need to unlock it with `f.flock(File::LOCK_UN)`. See for example https://www.safaribooksonline.com/library/view/ruby-cookbook/0596523696/ch06s13.html

lib/rmt/cli/base.rb
```diff
@@ -67,6 +67,12 @@ def handle_exceptions
         "The SCC credentials are not configured correctly in '/etc/rmt.conf'. You can obtain them from https://scc.suse.com/organization",
         RMT::CLI::Error::ERROR_SCC
       )
+    rescue RMT::Lockfile::ExecutionLockedError
+      pid = File.read(RMT::Lockfile::LOCKFILE_LOCATION)
+      raise RMT::CLI::Error.new(
+        "Process is locked by the application with pid #{pid}. Close this application or wait for it to finish before trying again\n",
```
lagartoflojo: How about moving the entire logic of reading the lockfile pid into `RMT::Lockfile` and then creating the `RMT::Lockfile::ExecutionLockedError` with the message? Then `CLI::Base` doesn't have to worry about lockfile locations and things, it just outputs the message in contained in `RMT::Lockfile::ExecutionLockedError`.

lib/rmt/lockfile.rb
lagartoflojo: Maybe a link to https://ruby-doc.org/core-2.3.1/File.html#method-i-flock would suffice. The code is pulled straight from the example there.

lib/rmt/lockfile.rb
Sergeykot: we can either delete or unlock the file. as I checked it now - both options work

lib/rmt/lockfile.rb
Sergeykot: @lagartoflojo I tried your suggestion, however maybe I understood it wrong - but it does not look any better IMHO. also another thing I changed - I unlock the file and write it to prevent he unexpected behavior described but @thutterer . so it looks like this:

ruby```
def lock(&_block)
      dirname = File.dirname(LOCKFILE_LOCATION)
      FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
      # https://ruby-doc.org/core-2.5.0/File.html#method-i-flock
      File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT) do |f|
        if f.flock(File::LOCK_EX | File::LOCK_NB)
          f.flock(File::LOCK_UN) # unlock for writing
          f.write(Process.pid.to_s)
          f.flock(File::LOCK_EX) # lock again
        else
          pid = File.read(RMT::Lockfile::LOCKFILE_LOCATION)
          raise ExecutionLockedError.new(
            "Process is locked by the application with pid #{pid}. Close this application or wait for it to finish before trying again\n"
          )
        end

        begin
          yield
        ensure
          f.close
          File.delete(LOCKFILE_LOCATION) if File.exist?(LOCKFILE_LOCATION)
        end
      end
    end
```

File: spec/lib/rmt/cli/import_spec.rb (hunk 8 above)
Comment (Sergeykot) [reply spec/lib/rmt/cli/import_spec.rb]: so I changed it stubbing the the `flock` method. if you mean using real filesystem though I would disagree to do so. also in other specs we use `FakeFS`

File: spec/lib/rmt/cli/import_spec.rb (hunk 9 above)
Comment (Sergeykot) [reply spec/lib/rmt/cli/import_spec.rb]: changed to stubbing the `flock` method

File: spec/lib/rmt/cli/import_spec.rb (hunk 10 above)
Comment (Sergeykot) [reply spec/lib/rmt/cli/import_spec.rb]: changed to stubbing the `flock` method

File: lib/rmt/lockfile.rb (hunk 7 above)
Comment (Sergeykot) [reply lib/rmt/lockfile.rb]: FYI: writing to file after locking it does not really writes to file however it even returns the number of bytes written. so that we can not know PID if we do so. I guess unlocking and locking it again fixes the problem

**Comment by thutterer** [comment]
@Sergeykot Have a look at a1e51b7
wfm :)

Ofc, needs update of specs now. But these should also get way simpler now.
Plus, you have to create the `/run/lock/rmt/` folder in the RPM specfile and chown it to the `_rmt` user.

**Comment by lagartoflojo** [comment]
@thutterer 's approach is also what I had in mind. I would still move the exception message creation into RMT::Lockfile :)

File: lib/rmt/lockfile.rb [hunk 15]
```diff
@@ -1,20 +1,24 @@
...
-        dirname = File.dirname(LOCKFILE_LOCATION)
-        FileUtils.mkdir_p(dirname) unless File.directory?(dirname)
-        File.open(LOCKFILE_LOCATION, 'w') { |f| f.write(Process.pid) }
-      end
-    end
+    def lock
+      # https://ruby-doc.org/core-2.5.0/File.html#method-i-flock
+      File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT) do |f|
+        if f.flock(File::LOCK_EX | File::LOCK_NB)
+          f.flock(File::LOCK_UN) # unlock for writing the PID
```
ikapelyukhin: You don't need to unlock the file to write it!

spec/lib/rmt/cli/main_spec.rb
```diff
@@ -2,47 +2,36 @@
...
+
       context 'without repositories marked for mirroring' do
         before { create :repository, :with_products, mirroring_enabled: false }
 
-        it 'outputs a warning' do
-          FakeFS.with_fresh do
-            expect_any_instance_of(RMT::Mirror).not_to receive(:mirror)
-            expect { command }.to output("There are no repositories marked for mirroring.\n").to_stderr.and output('').to_stdout
-          end
+        it 'outputs a warning', :with_fakefs do
```
lagartoflojo: Missed this -- `, :with_fakefs` is not necessary here

lib/rmt/lockfile.rb
Sergeykot: fixed

package/rmt-server.spec
```diff
@@ -180,6 +180,10 @@ rm -f %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/extensions/*/*/*/gem_make.out
 rm -f %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/extensions/*/*/*/mkmf.log
 find %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/gems/yard*/ -type f -exec chmod 644 -- {} +
 
+# lockfile location
+mkdir -p /run/lock/rmt/
+chown -R _rmt /run/lock/rmt/
```
lagartoflojo: Actually, `/run/lock` is owned by `root:lock`, so what we should do is add `_rmt` to the `lock` group.
Then we don't need to `chown` this, and I would make the LOCKFILE_LOCATION simply `/run/lock/rmt.lock` (no `rmt/` subdir)

.rubocop.yml
```diff
@@ -37,4 +37,4 @@ Naming/UncommunicativeMethodParamName:
   - id
 
 RSpec/NestedGroups:
-  MaxNesting: 4
```
lagartoflojo: wut really? I added that line and rubocop didn't complain about it!

lagartoflojo: I thought that in the rubocop docs I had read it was `MaxNesting`, not `Max`.. oh well :)

Sergeykot: `MaxNesting` is deprecated

package/rmt-server.spec
lagartoflojo: Other than this, this PR LGTM!! :shipit:

Sergeykot: fixed

package/rmt-server.spec
```diff
@@ -181,8 +181,8 @@ rm -f %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/extensions/*/*/*/mkmf.log
 find %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/gems/yard*/ -type f -exec chmod 644 -- {} +
 
 # lockfile location
-mkdir -p /run/lock/rmt/
-chown -R _rmt /run/lock/rmt/
+mkdir -p /run/lock/rmt.lock
```
lagartoflojo: you don;t need this anymore, since it's not a `dir` :)

Sergeykot: oh.. exactly. will have to recheck it while installing package

Sergeykot: fixed

thutterer:
@Sergeykot @lagartoflojo I vote for squashing commits here. Too many _cleanup_s without a meaningful commit message.

Sergeykot:
squashed commits

package/rmt-server.spec
```diff
@@ -180,6 +180,9 @@ rm -f %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/extensions/*/*/*/gem_make.out
 rm -f %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/extensions/*/*/*/mkmf.log
 find %{buildroot}%{lib_dir}/vendor/bundle/ruby/*/gems/yard*/ -type f -exec chmod 644 -- {} +
 
+# give _rmt user access to /run/lock
+usermod -a -G lock _rmt
```
lagartoflojo: when building:

```
[   22s] + usermod -a -G lock _rmt
[   22s] /var/tmp/rpm-tmp.Ugw9P7: line 104: usermod: command not found
```

I'm still trying to find the proper way to do this...

lagartoflojo: Ah, ok, I think we should add it here, when creating the user

https://github.com/SUSE/rmt/blob/936a0a37e0756a3ebec962a6e0c241428f932b75/package/rmt-server.spec#L219

lagartoflojo: Something like:

```
%{_sbindir}/useradd -g %{rmt_group} -G lock -s /bin/false -r \
```

I'm testing this right now.

lagartoflojo: Not working for me =/

It didn't add the user to the `lock` group.

And then I even added the user manually (with `usermod`), and still no cigar. Strange!

image

ikapelyukhin: Give it a try with `%files`?

lagartoflojo: The problem is not logging in and out, but that `usermod` is not usable within the RPM spec =/
And it's not possible to "log in" as `_rmt`, that's just the user that the `rmt-cli` process runs as. So even if you're root, `rmt-cli` will run as `_rmt`...

lagartoflojo: okay, so actually I think it doesn't make sense to create `/var/run/lock/rmt` or `/run/lock/rmt` or `/var/lock/rmt` in the RPM.

```
> ll /var/
lrwxrwxrwx  1 root root    9 Apr 13 15:28 lock -> /run/lock/
lrwxrwxrwx  1 root root    4 Apr 13 15:28 run -> /run/
```

and 

```
> mount | grep tmpfs
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
```

Basically, `/run` is a `tmpfs` so anything that gets created there will be deleted upon system restart. It makes no sense for the RPM to create something there.

So I think that adding `_rmt` to the `lock` group and just writing to `/var/lock` should work. I just don't know how to make that work :cry: 
This attempt doesn't work: https://github.com/SUSE/rmt/commit/1057783e54d0c8a7d62b9e83b585a85d75bb5fa6

`_rmt` does not get added to the `lock` group.
And even adding it manually with `usermod`, it still complains about permission denied.

Strangely, adding my own user to `lock` did let me write to `/var/lock`.

ikapelyukhin: @lagartoflojo I thing the issue might be the way we switch the user here:

https://github.com/SUSE/rmt/blob/17c2366e9a48af93d0f1e01118c8a9e9364f9398/bin/rmt-cli#L15-L20

That sets the group only to `nginx`, the process will not have `lock` group. I suggest you investigate `supplementary groups` and also this: https://ruby-doc.org/core-2.1.1/Process.html#method-c-groups

lagartoflojo: I see... Okay, @Sergeykot you can investigate that, this is your PR after all hehe..
But I'm thinking that it would be easier to just write the lock file to `Rails.root + '/tmp'` as it was before.

ikapelyukhin: `Rails.root + '/tmp'` might be problematic -- I think some users might be mounting `/usr` in read-only mode, and this is why the whole convention exists about putting stuff that the application generates into `/var`. `/tmp` should work, though.

PR #158, opened 2018-05-07
RMT package should bind to 127.0.0.1, not 0.0.0.0

tmuntaner:


package/rmt.service
```diff
@@ -11,7 +11,7 @@ User=_rmt
 PrivateTmp=yes
 Environment=LOG_TO_JOURNALD=1
 WorkingDirectory=/usr/share/rmt
-ExecStart=/usr/share/rmt/bin/rails server -e production
+ExecStart=/usr/share/rmt/bin/rails server -e production -b 127.0.0.1
```
lagartoflojo: Any idea why it's even binding to 0.0.0.0? AFAIK Rails doesn't do that by default.

lagartoflojo: First things first -- are we sure that it's binding to `0.0.0.0`?

ikapelyukhin: @lagartoflojo 

On my SLES15 machine:

```
f193:~ # ps aux | grep rmt
_rmt     12577  0.0  4.8 777724 92040 ?        Ssl  Apr24   2:30 puma 3.10.0 (tcp://0.0.0.0:4224) [rmt]
root     17798  0.0  0.0   8688   812 pts/0    S+   12:04   0:00 grep --color=auto rmt
f193:~ # rpm -qi rmt-server
Name        : rmt-server
Version     : 0.0.6
Release     : 42.1
```

So it is listening on `tcp://0.0.0.0:4224` with the default config.

ikapelyukhin:
```
f193:~ # rpm -V rmt-server
S.5....T.  c /etc/rmt.conf
f193:~ #
```

lagartoflojo: Ok, it seems it's Puma's default config to listen on 0.0.0.0.

https://github.com/puma/puma/blob/master/examples/config.rb#L69-L77

Why not change our [puma.rb](https://github.com/SUSE/rmt/blob/master/config/puma.rb) file instead of hard-coding it in the service file?

tmuntaner: @lagartoflojo As far as I know, we don't package a `puma.rb` file. If we create one now, that is one extra file to manage for little benefit.

tmuntaner: I like the service file more, it explicitly states that we're binding to '127.0.0.1', it's rails server independent, and will be the most likely place for anyone unfamiliar with puma to look.

tmuntaner: Thanks @fschueller

tmuntaner:
Thanks @fschueller

PR #168, opened 2018-05-11
Skip repos check when deactivating a module/extension

lagartoflojo:
Needed for the use case described here https://github.com/SUSE/connect/pull/365#issuecomment-387789520

> 1. Register base product, basesystem module and server applications module on the client;
> 1. Disable server applications on RMT;
> 1. Try to de-register server applications -- I think it would also break.
> - @ikapelyukhin

Spoiler alert: It does break:

```
# bin/SUSEConnect -d -p sle-module-server-applications/15/x86_64
Error: Registration server returned 'Not all mandatory repositories are mirrored for product Server Applications Module 15 x86_64 (BETA)' (422)
```

### DIFF
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -2,6 +2,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   before_action :authenticate_system
   before_action :require_product, only: %i[show activate upgrade destroy]
+  before_action :check_product_service_and_repositories, only: %i[show activate upgrade]
   before_action :check_base_product_dependencies, only: %i[activate upgrade show]
 
   def activate
@@ -85,7 +86,6 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
     unless @product
       raise ActionController::TranslatedError.new(N_('No product found'))
     end
-    check_product_service_and_repositories
   end
 
   def check_product_service_and_repositories
@@ -93,8 +93,8 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
       fail ActionController::TranslatedError.new(N_('No repositories found for product: %s'), @product.friendly_name)
     end
 
-    mandatory_repos = @product.repositories.where({ enabled: true })
-    mirrored_repos = @product.repositories.where({ enabled: true, mirroring_enabled: true })
+    mandatory_repos = @product.repositories.only_enabled
+    mirrored_repos = @product.repositories.only_enabled.only_mirrored
 
     unless (mandatory_repos.size == mirrored_repos.size)
       fail ActionController::TranslatedError.new(N_('Not all mandatory repositories are mirrored for product %s'), @product.friendly_name)
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -2,7 +2,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   before_action :authenticate_system
   before_action :require_product, only: %i[show activate upgrade destroy]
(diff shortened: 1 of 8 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

spec/models/repository_spec.rb
```diff
@@ -16,6 +16,61 @@
   it { is_expected.to have_db_column(:name).of_type(:string).with_options(null: false) }
   it { is_expected.to have_db_column(:external_url).of_type(:string).with_options(null: false) }
 
+  describe 'scopes' do
+    describe '.only_mirrored' do
+      it 'returns only mirrored repos' do
+        mirrored = create :repository, mirroring_enabled: true
```
Sergeykot: maybe move things like this to `let()` or `let!()` blocks
also I would use `()`-brackets while calling `create` method  https://github.com/bbatsov/ruby-style-guide#method-invocation-parens bit it's more the matter of taste :wink: 
both applies to similar code below

spec/models/repository_spec.rb
```diff
@@ -36,23 +91,41 @@
...
     end
-  end
 
-  describe '#remove_repository' do
-    let(:custom_repository) { create :repository, :custom }
-    let(:suse_repository) { create :repository }
+    context 'with no subpath and no trailing slash' do
+      let(:url) { 'http://localhost.com' }
+
+      it { is_expected.to eq '/' }
```
Sergeykot: also I would use `()`-brackets

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -2,6 +2,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   before_action :authenticate_system
   before_action :require_product, only: %i[show activate upgrade destroy]
+  before_action :check_product_service_and_repositories, only: %i[show activate upgrade]
```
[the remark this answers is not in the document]
lagartoflojo: Added in my last commit.

app/models/repository.rb
```diff
@@ -7,6 +7,7 @@ class Repository < ApplicationRecord
 
   scope :only_installer_updates, -> { unscope(where: :installer_updates).where(installer_updates: true) }
```
lagartoflojo: @ikapelyukhin any idea why we `unscope` here? is it really necessary?

spec/models/repository_spec.rb
```diff
@@ -18,56 +18,59 @@
 
   describe 'scopes' do
     describe '.only_mirrored' do
-      it 'returns only mirrored repos' do
-        mirrored = create :repository, mirroring_enabled: true
-        create :repository, mirroring_enabled: false
+      subject { Repository.only_mirrored }
```
ikapelyukhin: `Repository` -> `described_class` :grimacing:

app/models/repository.rb
ikapelyukhin: Yeah, unscoping and then scoping looks weird.
My best guess is that it is a weird artifact after porting the code from glue (there it has a default scope, which has to be cleared out by default).
Let's try removing it :-)

spec/models/repository_spec.rb
lagartoflojo: :man_facepalming:

lagartoflojo: Rubocop didn't catch that, weird.

ikapelyukhin: Yeah, I though we had a rule for that.

spec/models/repository_spec.rb
lagartoflojo: We have code with and without parentheses all over the place, I wouldn't fret too much about it. If Rubocop complains, then we fix it, otherwise...

lagartoflojo:
@ikapelyukhin should be ready to go now :)

PR #169, opened 2018-05-14
Fix eula_url in product serializer

tmuntaner:
This fixes issues with the installation of SLES 15 against RMT.

The problem is that `object.eula_url` can be either nil or an empty string. The if statement did not catch this and gave connect a broken url like `https://rmt.scc.suse.de:443/repo/`.


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/serializers/v3/product_serializer.rb
```diff
@@ -25,7 +25,7 @@ def arch
   end
 
   def eula_url
-    if object.eula_url
+    if object.eula_url.to_s != ''
```
lagartoflojo: `object.eula_url.present?` should do it.

spec/serializers/v3/product_serializer_spec.rb
```diff
@@ -11,6 +11,26 @@
   end
   let(:base_url) { 'http://example.com' }
 
+  describe 'eula_url' do
+    it 'handles empty string' do
+      product = create(:product, :extension, base_products: [sles15], recommended: true, eula_url: '')
```
lagartoflojo: any reason you are creating an extension & recommended product? i don't think that `#eula_url` cares about that.

spec/serializers/v3/product_serializer_spec.rb
```diff
@@ -11,6 +11,26 @@
...
+      serializer = described_class.new(product, root_product: sles15, base_url: base_url)
+      expect(serializer.eula_url).to eq('')
+    end
+
+    it 'handles url' do
+      product = create(:product, :extension, base_products: [sles15], recommended: true, eula_url: 'http://example.com/eula')
+      serializer = described_class.new(product, root_product: sles15, base_url: base_url)
+      expect(serializer.eula_url).to eq('http://example.com/repo/eula')
+    end
+  end
```
lagartoflojo: Better Specs FTW

```ruby
  describe '#eula_url' do
    subject { described_class.new(product, base_url: base_url).eula_url }

    let(:product) { create :product, eula_url: eula_url }

    context "when the product's eula_url is an empty string" do
      let(:eula_url) { '' }

      it { is_expected.to eq '' }
    end

    context "when the product's eula_url is nil" do
      let(:eula_url) { nil }

      it { is_expected.to eq '' }
    end

    context "when the product's eula_url is a URL" do
      let(:eula_url) { 'http://example.com/eula' }

      it { is_expected.to eq 'http://example.com/repo/eula' }
    end
  end
```

felixsch:
LGTM :+1:

app/serializers/v3/product_serializer.rb
tmuntaner: You're right, I pushed a fix

tmuntaner:
Thanks @felixsch and @lagartoflojo

kalabiyau:
🥇

PR #173, opened 2018-05-16
Fix upgrade endpoint

tmuntaner:
* don't check activations on upgrade
* delete old activations that are successors/predecessors
* activate new product

### DIFF
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -48,17 +48,14 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
   end
 
   def upgrade
-    activation = @system.activations.joins(:service).find_by('services.product_id': [@product.id, @product.predecessor_ids, @product.successor_ids].flatten)
+    obsoleted_product_ids = ([@product] + @product.predecessors + @product.successors).map(&:id)
+    obsoleted_service = @system.services.find_by(product_id: obsoleted_product_ids)
+    @obsoleted_service_name = obsoleted_service.name if obsoleted_service
 
-    unless activation
-      raise ActionController::TranslatedError.new(
-        N_("No activation with product '%s' was found."),
-        @product.friendly_name
-      )
+    ActiveRecord::Base.transaction do
+      remove_previous_product_activations(obsoleted_product_ids)
+      create_product_activation
     end
-
-    activation.service = @product.service
-    activation.save!
 
     render_service
   end
@@ -103,6 +100,10 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   def create_product_activation
     @system.activations.where(service_id: @product.service.id).first_or_create
+  end
+
+  def remove_previous_product_activations(product_ids)
+    @system.activations.includes(:product).where('products.id' => product_ids).destroy_all
   end
 
   # Check if extension base product is already activated
--- a/spec/requests/api/connect/v3/systems/products_controller_spec.rb
+++ b/spec/requests/api/connect/v3/systems/products_controller_spec.rb
@@ -207,90 +207,78 @@ RSpec.describe Api::Connect::V3::Systems::ProductsController do
   end
 
   describe '#upgrade' do
+    subject { response }
+
+    let(:request) { put url, headers: headers, params: payload }
+    let(:new_product) { FactoryGirl.create(:product, :with_mirrored_repositories) }
+    let(:payload) do
+      {
+        identifier: new_product.identifier,
+        version: new_product.version,
+        arch: new_product.arch
+      }
+    end
+    let(:serialized_json) do
+      V3::ServiceSerializer.new(
+        new_product.service,
+        base_url: URI::HTTP.build({ scheme: response.request.scheme, host: response.request.host }).to_s
+      ).to_json
+    end
+
     it_behaves_like 'products controller action' do
       let(:verb) { 'put' }
     end
 
-    context 'with not activated product' do
-      before { put url, headers: headers, params: payload }
-      subject { response }
-
-      let(:product) { FactoryGirl.create(:product, :with_mirrored_repositories) }
-      let(:payload) do
-        {
-          identifier: product.identifier,
-          version: product.version,
-          arch: product.arch
-        }
-      end
-
-      let(:error_response) do
-        {
-          type: 'error',
-          error: "No activation with product '#{product.friendly_name}' was found.",
-          localized_error: "No activation with product '#{product.friendly_name}' was found."
-        }
-      end
-
-
-      its(:code) { is_expected.to eq('422') }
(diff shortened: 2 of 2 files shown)


app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -105,6 +102,10 @@ def create_product_activation
     @system.activations.where(service_id: @product.service.id).first_or_create
   end
 
+  def remove_previous_product_activation(product_ids)
```
lagartoflojo: activationS

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -48,18 +48,15 @@ def offline_migrations
   end
 
   def upgrade
-    activation = @system.activations.joins(:service).find_by('services.product_id': [@product.id, @product.predecessor_ids, @product.successor_ids].flatten)
+    obsoleted_product_ids = ([@product] + @product.predecessors + @product.successors).map(&:id)
+    obsoleted_service = @system.services.find_by(product_id: obsoleted_product_ids)
+    @obsolted_service_name = obsoleted_service.name if obsoleted_service
```
lagartoflojo: typo in `@obsoleted_...`

lagartoflojo: How are tests passing with such a typo? I fear we're not testing something

tmuntaner: That doesn't seem to have been tested ever but was coded in `render_service`, I will see if I can throw one in.

lagartoflojo:
yeow conflic

lagartoflojo:
tested with the real SLE 15 ISO and it works :+1:

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,82 @@
...
+      subject { response }
 
-      let(:error_response) do
-        {
-          type: 'error',
-          error: "No activation with product '#{product.friendly_name}' was found.",
-          localized_error: "No activation with product '#{product.friendly_name}' was found."
-        }
-      end
+      specify { expect(system.activations.count).to eq(1) }
```
ikapelyukhin: 1. Didn't know that `specify` was a thing, TIL. :-) It is used in very few places, maybe we should just use `it` instead?
2. The example should have a name, otherwise in the `--format=documentation` it looks like this:
```
#upgrade
  new product
    should eq 1
```

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,82 @@
...
-      let(:product) { FactoryGirl.create(:product, :with_mirrored_repositories) }
-      let(:payload) do
-        {
-          identifier: product.identifier,
-          version: product.version,
-          arch: product.arch
-        }
-      end
+    context 'new product' do
+      subject { response }
```
ikapelyukhin: `subject` in all of those examples is the same and can be moved out to `describe` block.

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,82 @@
...
-      its(:body) { is_expected.to eq(error_response.to_json) }
+      it 'activates new product' do
+        expect(system.activations.first.reload.service_id).to equal(new_product.service.id)
+      end
     end
 
-    context 'with activated product' do
-      let(:request) { put url, headers: headers, params: payload }
+    context 'with activated previous product' do
+      subject { response }
```
ikapelyukhin: `subject` in all of those examples is the same and can be moved out to describe block.

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,82 @@
...
-        let(:payload) do
-          {
-            identifier: new_product.identifier,
-            version: '24.0-0',
-            arch: new_product.arch
-          }
-        end
+    context 'with "-" in product version' do
+      before { request }
+      subject { response }
```
ikapelyukhin: `subject` in all of those examples is the same and can be moved out to describe block.

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,82 @@
...
-        let(:new_product) { FactoryGirl.create(:product, :with_mirrored_repositories, version: '24.0', predecessors: [old_product]) }
-        let(:payload) do
-          {
-            identifier: new_product.identifier,
-            version: '24.0-0',
-            arch: new_product.arch
-          }
-        end
+    context 'with "-" in product version' do
+      before { request }
```
ikapelyukhin: There's already a shared `before` block on `describe` level, this isn't needed.

spec/requests/api/connect/v3/systems/products_controller_spec.rb
tmuntaner: I didn't know about `specify` either, it was in the specs there already. I will see about addressing the other issues.

ikapelyukhin: If it was in the specs already doesn't mean it should stay!

ikapelyukhin: Used only 19 times in glue. Looks like really old stuff :grimacing:

tmuntaner: True, but I didn't have a problem with it other than the documentation issue that you had.

spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -207,90 +207,78 @@
...
 
-        let(:new_product) { FactoryGirl.create(:product, :with_mirrored_repositories, version: '24.0', predecessors: [old_product]) }
-        let(:payload) do
-          {
-            identifier: new_product.identifier,
-            version: '24.0-0',
-            arch: new_product.arch
-          }
-        end
+      it('has one activation') { expect(system.activations.count).to eq(1) }
```
ikapelyukhin: :+1:

spec/requests/api/connect/v3/systems/products_controller_spec.rb
lagartoflojo: @tmuntaner @ikapelyukhin I've been sprinkling `specify`s in our tests, sorry hehe. (It's not "really old stuff). But I can explain.

when you have a `subject` you can then say `it { is_expected ... }`, which looks nice.
But sometimes you need to run the subject within a block, to check for example if it raises an error: `it { expect { subject }.to raise_error }`
But "it expect" looks weird, so in such a case, `specify` is used: `specify { expect { subject }.to raise_error }`

In this particular case, I would agree that `specify` doesn't really need to be used, since there's no mention of `subject`. And probably, a more descriptive `it` block would be better.

PR #175, opened 2018-05-17
Squash migrations

Sergeykot:
Squashed with https://github.com/jalkoby/squasher


Sergeykot:
also: do we need a record in `changelog` for this?

Sergeykot:
also I want to put info about how to squash migrations however not sure if we want to have gem I used for it in the repo. 
where is a good place for that instructions?

PR #177, opened 2018-05-18
Safer configuration

ikapelyukhin:
1. Added default values for database section;
2. Now uses safe navigation when accessing nested parameters.

### DIFF
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -7,4 +7,4 @@ require 'rails/command'
 require 'rails/commands/server/server_command'
 
 Rails::Command::ServerCommand.send(:remove_const, 'DEFAULT_PORT')
-Rails::Command::ServerCommand.const_set('DEFAULT_PORT', Settings&.[](:rails)&.[](:port) || 4224)
+Rails::Command::ServerCommand.const_set('DEFAULT_PORT', 4224)
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -5,14 +5,14 @@ class RMT::HttpRequest < Typhoeus::Request
 
   def set_defaults
     Typhoeus::Config.user_agent = "RMT/#{RMT::VERSION}"
-    Typhoeus::Config.verbose = Settings.http_client.verbose || nil
+    Typhoeus::Config.verbose = Settings.try(:http_client).try(:verbose) || nil
 
     super
 
-    options[:proxy] = Settings.http_client.proxy
-    options[:proxyauth] = Settings.http_client.proxy_auth ? Settings.http_client.proxy_auth.to_sym : :any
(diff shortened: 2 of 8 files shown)


lib/rmt/http_request.rb
```diff
@@ -5,14 +5,14 @@ class RMT::HttpRequest < Typhoeus::Request
 
   def set_defaults
     Typhoeus::Config.user_agent = "RMT/#{RMT::VERSION}"
-    Typhoeus::Config.verbose = Settings.http_client.verbose || nil
+    Typhoeus::Config.verbose = Settings.try(:http_client).try(:verbose) || nil
```
thutterer: The ` || nil` at the end should not be required anymore, right?

ikapelyukhin: Right!

suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lagartoflojo:
As you may remember, I'm not a fan of `try` (nor ruby's safe operator `.&`)..
It basically makes code insecure about itself... is the thing there? or is it not? :scream: if it's not, you have to check for `nil`s all over the place. => Sadness.

How would I prefer to solve it in this case? We could:

- Never use the `Settings` class directly, only from `lib/rmt/config.rb`. Why do we have some settings accessed from `Settings` and yet others from `RMT::Config`?
- [Prepend a default config](https://github.com/railsconfig/config#adding-sources-at-runtime) to the `config` gem, which is included somewhere in the RMT sources, ie. it's not user-modifiable.

I'm leaving this here as a comment to open the discussion, not as a ship-stopping review :)

config/boot.rb
```diff
@@ -7,4 +7,4 @@
 require 'rails/commands/server/server_command'
 
 Rails::Command::ServerCommand.send(:remove_const, 'DEFAULT_PORT')
-Rails::Command::ServerCommand.const_set('DEFAULT_PORT', Settings&.[](:rails)&.[](:port) || 4224)
```
ikapelyukhin: `Settings&.[](:rails)&.[](:port)` -- so pretty :grimacing:

ikapelyukhin:
@lagartoflojo Yeah, the config needs some love. It would be actually cool to use schema validation for the config, it is supported by `railsconfig` gem in some form.

I suggest we add the task to the backlog and properly plan it later.

lagartoflojo:
> It would be actually cool to use schema validation for the config,

Oh yeah, that's cool! https://github.com/railsconfig/config#validation

tmuntaner:
But I agree that it's worth it to clean up the config properly in the future. It can be a good candidate for the maintenance column.

PR #184, opened 2018-06-01
Use curl instead of wget

wstephenson:


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

public/tools/rmt-client-setup
```diff
@@ -40,6 +40,99 @@ EOT
 exit 1
 }
 
+function http_get_signing_keys_index_html()
+{
+    local index_html_url=`echo "$REGURL" | awk -F/ '{print "https://" $3 "/repo/keys/"}'`
+
+    $CURL --insecure --fail --output $TMPDIR/index.html $index_html_url
```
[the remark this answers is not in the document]
ikapelyukhin: Why should be there an `index.html`, specifically? (It's a directory listing, not a static HTML file)

public/tools/rmt-client-setup
```diff
@@ -40,6 +40,99 @@ EOT
...
+
+    $CURL --silent --fail --output "$TMPDIR/$1" $key_file_url
+
+    return $?
+}
+
+function import_key
+{
+    if [ "$1" == "res-signingkeys.key" -a $ISRES -eq 0 ]; then
+        # this is no RES system, so we do not need this key
```
ikapelyukhin: This definitely should be dropped. :grimacing:

wstephenson: Is there a chance that customers may copy custom keys into their RMT repo/keys when upgrading from SMT, bringing this key with them and then adding it to their new systems. If so wouldn't it make sense to leave this filter in?

ikapelyukhin: Ah, it's skipping the keys.
Yes, you are right -- it is possible to have those keys after migrating from SLE11.
But in case of this script, `$ISRES` is always `0`, might make sense to remove it and do a check just based on the file name, then.

wstephenson: Ok

PR #185, opened 2018-06-04
Log RMT version on startup

ikapelyukhin:


config/application.rb
```diff
@@ -1,16 +1,12 @@
...
+
 require 'active_model/railtie'
 require 'active_record/railtie'
 
 require 'action_controller/railtie'
 require 'action_view/railtie'
-# require 'action_mailer/railtie'
-# require 'active_job/railtie'
-# require 'action_cable/engine'
-# require 'sprockets/railtie'
```
ikapelyukhin: Small clean-up.

suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

config/initializers/log_version.rb
```diff
@@ -0,0 +1,2 @@
+# Output version to log on startup
+Rails.logger.info "Starting RMT server version #{RMT::VERSION}"
```
lagartoflojo: Should I be seeing this message when running `rails s`? Because I don't see it.
The file IS getting loaded though. Putting a `puts` there, I see it.

ikapelyukhin: You are not seeing it, because in development mode logs are written to a file.
Run in production mode and you'll see it on `STDOUT`.

lagartoflojo: Huh, okay. I found it strange because, at least locally, I also see all logging in the console.

ikapelyukhin: `Rails.logger` writes to a file in dev mode, but other logger instances (`rmt-cli` an such) might always log to `STDOUT`.
Maybe things like `rmt-cli mirror` and `sync` should report their versions as well?

ikapelyukhin:
The logs aren't looking very nice:

* Missing severity stamps on some of the lines (`Mirroring repository`)
* The version line doesn't come first
* The version line is repeated for each repo (seems like the logger is being instantiated for every repo)

```
Jun 06 11:38:41 f193 rmt-cli[21833]: Mirroring repository SLE-Product-SLES15-Updates to /usr/share/rmt/public/repo
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: RMT version 1.0.0
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: No product license found
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml.key
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml.asc
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ 401dc19bda88c82c403423fb835844d64345f7e95f5b9835888189c03834cc93-filelists.xml.gz
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ 6bf9672d0862e8ef8b8ff05a2fd0208a922b1f5978e6589d87944c88259cb670-other.xml.gz
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ 4372c8f207c64d1837e706043fa0a3d439af7673ff71e084855634d5e641e47e-primary.xml.gz
Jun 06 11:38:41 f193 rmt-cli[21833]: Mirroring repository SLE-Product-SLES15-Pool to /usr/share/rmt/public/repo
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: RMT version 1.0.0
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ directory.yast
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ license.txt
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml.key
Jun 06 11:38:41 f193 rmt-cli[21833]: INFO: ↓ repomd.xml.asc
```

lagartoflojo:
the "mirroring repository" line uses `puts` instead of the logger: https://github.com/SUSE/rmt/blob/ba1d0ec94215081e80b6f34d78faf0af82976fdd/lib/rmt/cli/export.rb#L37

ikapelyukhin:
@lagartoflojo Yes, obviously. I'm looking into that.

ikapelyukhin:
@lagartoflojo I've added hacks to have the logger for `RMT::Mirror` and created #189 -- the code there isn't looking good ATM, but refactoring isn't strictly related to this feature.

ikapelyukhin:
Now the output should be consistent:

```
INFO: RMT version 1.0.0
INFO: Mirroring repository SLE-Product-SLES15-Debuginfo-Updates to /home/ivan/work/rmt/public/repo
INFO: No product license found
WARN: Repodata download failed: repodata/repomd.xml - HTTP request failed with code 410
INFO: Mirroring repository SLE15-Installer-Updates to /home/ivan/work/rmt/public/repo
INFO: No product license found
INFO: → repomd.xml
INFO: → repomd.xml.key
INFO: → repomd.xml.asc
INFO: → 401dc19bda88c82c403423fb835844d64345f7e95f5b9835888189c03834cc93-filelists.xml.gz
INFO: → 6bf9672d0862e8ef8b8ff05a2fd0208a922b1f5978e6589d87944c88259cb670-other.xml.gz
INFO: → 4372c8f207c64d1837e706043fa0a3d439af7673ff71e084855634d5e641e47e-primary.xml.gz
INFO: Mirroring repository SLE-Product-SLES15-Updates to /home/ivan/work/rmt/public/repo
INFO: No product license found
INFO: → repomd.xml
```

lagartoflojo:
Hmm, as you've well said, passing the logger instance around feel likes a hack... How about something like what's suggested here? https://stackoverflow.com/a/23508304 By simply including a module, we'll have access to the logger were we need it.

ikapelyukhin:
@lagartoflojo well, the logger has to be accessible outside of `RMT::Mirror` as well:

https://github.com/SUSE/rmt/blob/8c3e0d2223416a6a8380a73a4f4bcaec3b61f6e5/lib/rmt/cli/base.rb#L100

Encapsulating the logger into `RMT::Mirror` and then calling it from outside of the class would be also a hack.

There's nothing wrong with dependency injection of the logger, the problem is that the methods around `RMT::Mirror` constructors are hacky -- fixing #189 would allow to pass it just once into the constructor instead of passing it around 10 times like now.

lagartoflojo:
Right, that's why I mean that having a Logging module, like the one in the SO answer, would help. It would just be included everywhere it's needed (CLI::Base in your example).

ikapelyukhin:
Ah, you mean to make the logger a singleton. Well, we could do that -- will require hacks in the specs to redirect it to `/dev/null`, though.

lagartoflojo:
Yes, exactly. It's already a singleton in Rails, you always have access to `logger`, everywhere, so it's counter-intuitive that in the CLI code, it has to be explicitly passed-in.

In the specs, in the `spec_helper` all we'd need to do is say `Logging.logger = Logger.new('/dev/null')`.

ikapelyukhin:
It's still a hack though (e.g., [this](https://stackoverflow.com/questions/12755539/why-is-singleton-considered-an-anti-pattern)).

lagartoflojo:
Yep, I know singletons are an anti-patter, but Rails already has it so we can just embrace it :D 
Anyway, it's an idea. If you prefer your current implementation, that's also fine by me.

ikapelyukhin:
It is not a singleton in Rails. Calling the constructor of the logger would give you a new instance of the logger. `Rails.logger` is just an accessor.

package/rmt-server.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Jun 15 13:12:09 UTC 2018 - ikapelyukhin@suse.com
+
+- Log version on startup
```
thutterer: Is it maybe worth mentioning that it does this only when ran as a service?
Now you could think it would also log it when started manually on the command line.

ikapelyukhin: OK!

PR #187, opened 2018-06-05
Log RMT version in journald logs

ikapelyukhin:


ikapelyukhin:
@lagartoflojo hmm, how about this? It would log the version only in `journald` logs, not just for the Rails, but for mirroring and SCC sync timers as well.

PR #197, opened 2018-06-27
Fix duplicate migration paths

ikapelyukhin:


app/models/migration_engine.rb
```diff
@@ -103,6 +103,7 @@ def sort_migrations(migrations)
       .map { |migration| sort_migration(migration) }
       .sort_by { |migration| migration.map(&:version) }
       .reverse!
+      .uniq
```
Sergeykot: after adding this `.uniq` method call - do we really need to call it above?
https://github.com/SUSE/rmt/blob/aa15e6e8411f107b0c2cc0c2f9373e4a087fbad7/app/models/migration_engine.rb#L102

ikapelyukhin: We do. The first call removes duplicate products within one migration path (e.g., `[ [sle15, sle15, sle15], [sle12] ]` -> `[ [sle15], [sle12] ]`).
The second one removes duplicate migration paths (e.g., `[ [sle15], [sle15] ]` -> `[ [sle15] ]`).

PR #206, opened 2018-07-27
order list items for output

thutterer:
- products
- repos
- custom reps

### DIFF
--- a/lib/rmt/cli/repos.rb
+++ b/lib/rmt/cli/repos.rb
@@ -28,7 +28,7 @@ class RMT::CLI::Repos < RMT::CLI::Base
   protected
 
   def list_repositories(scope: :enabled)
-    repositories = (scope == :all) ? Repository.only_scc : Repository.only_scc.only_mirrored
+    repositories = ((scope == :all) ? Repository.only_scc : Repository.only_scc.only_mirrored).order(:name, :description)
 
     if repositories.empty?
       if options.all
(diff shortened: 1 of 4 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lib/rmt/cli/repos.rb
```diff
@@ -28,7 +28,7 @@ def disable(id)
   protected
 
   def list_repositories(scope: :enabled)
-    repositories = (scope == :all) ? Repository.only_scc : Repository.only_scc.only_mirrored
+    repositories = ((scope == :all) ? Repository.only_scc : Repository.only_scc.only_mirrored).order(:name, :description)
```
ikapelyukhin: Why `:description`? Maybe there's something more suitable for sorting, arch or friendly name maybe?

thutterer: Well, it only makes sense to order by stuff that is also visible in the output, which would be:
https://github.com/SUSE/rmt/blob/3f45944df9ed56e37d052fda544e967860b563a7/lib/rmt/cli/repos.rb#L40-L47

ikapelyukhin: Well, description is basically `#{name} for #{arch}`, so sorting by description is equivalent to sorting by arch when names are the same.

thutterer:
> Please mention the bug number on the changes file.

@ikapelyukhin Done. I amended the previous commit with the changelog entry.

PR #209, opened 2018-08-08
CLI integration tests

Sergeykot:
commands for running locally:

`docker-compose -f docker-compose-test.yml build`
`docker-compose -f docker-compose-test.yml run --rm rmt_test ./integration.sh`


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Dockerfile.test
```diff
@@ -0,0 +1,60 @@
...
+    mariadb-client ruby2.5-rubygem-bundler make build sudo ruby-devel
+
+
+ENV DOCKERIZE_VERSION v0.6.0
+RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+
+ARG OBS_USER
+ARG OBS_PASSWORD
```
lagartoflojo: Might be cleaner to put all `ARG`s at the top, so they are easier to see.
(This is not a full review ;-) )

features/lockfile_spec.rb
```diff
@@ -0,0 +1,17 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
+
+describe 'rmt-cli' do
+  before(:example) { system '/usr/bin/rmt-cli sync > /dev/null &' }
+  # kill running process to let further specs pass
+  after(:example) { system 'kill -9 $(cat /tmp/rmt.lock)' }
```
lagartoflojo: the default scope of `before` and `after` hooks is `:example` (or `:each`, which is the same) so you don't need to specify it

Dockerfile.test
Sergeykot: fixed

features/lockfile_spec.rb
Sergeykot: fixed

lib/tasks/package.rake
```diff
@@ -0,0 +1,17 @@
+namespace :package do
```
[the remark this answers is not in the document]
Sergeykot: it is being used for integrations tests only

features/sync_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+describe 'rmt-cli' do
+  describe 'sync' do
+    command '/usr/bin/rmt-cli sync'
+    its(:exitstatus) { is_expected.to eq 0 }
+    its(:stdout) { is_expected.to include("INFO -- : Cleaning up the database\n") }
+    its(:stdout) { is_expected.to include("INFO -- : Downloading data from SCC\n") }
+    its(:stdout) { is_expected.to include("INFO -- : Updating products\n") }
+    its(:stdout) { is_expected.to include("INFO -- : Updating repositories\n") }
+    its(:stdout) { is_expected.to include("INFO -- : Updating subscriptions\n") }
```
ikapelyukhin: I don't think we should care about the output of the command in the integration test at all:
 * If we want to check the output of the command, it would be much easier to test this in unit tests (no need to have this super complicated setup with docker).
 * If the output is changed by a PR, the tests in the PR would be green and would break in the Jenkins pipeline after the PR is merged to master.

What we actually should verify is that SCC sync works correctly, which means that we can get products, repositories, etc. downloaded to the local database.

features/repos_spec.rb
```diff
@@ -0,0 +1,8 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
+
+[3114, 3115, 3116, 2705, 2707].each do |repo_id|
+  describe 'enable repos' do
+    command "/usr/bin/rmt-cli repos enable #{repo_id}"
+    its(:stdout) { is_expected.to eq("Repository successfully enabled.\n") }
```
ikapelyukhin: This should be already taken care of by the unit tests.

features/mirror_spec.rb
```diff
@@ -0,0 +1,8 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
+
+describe 'mirror' do
+  before { system '/usr/bin/rmt-cli repos enable 3114' }
+
+  command '/usr/bin/rmt-cli mirror'
+  its(:exitstatus) { is_expected.to eq 0 }
```
ikapelyukhin: Mirroring process would exit with code zero if mirroring of some repositories failed. We aren't really checking if we managed to download any RPM files.

Sergeykot:
@fschueller I updated the spec for mirroring so that we have a repo for mirroring and it should work again

features/mirror_spec.rb
Sergeykot: good point! fixed

features/sync_spec.rb
Sergeykot: fixed

lib/tasks/package.rake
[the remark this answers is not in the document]
Sergeykot: yes, sorry, now I see that I removed usage of this task from `integration.sh` some time ago. I think I will delete it

package/rmt-server.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Wed Aug 15 13:41:49 UTC 2018 - skotov@suse.com
+
+- Change file premissions for rmt.conf
```
fschueller: You have a small typo here.

Sergeykot: fixed

Sergeykot:
@fschueller I fixed it. the reason was deleting `/var/lib/rmt/public/repo` repo after mirror command spec
now I changed it in the way that it deletes only files for one repo, not the whole dir.
also included #213 here

Sergeykot:
@fschueller you may need to rebuild the image with `--no-cache` or create the directory `/var/lib/rmt/public/repo` inside container

Dockerfile.test
```diff
@@ -0,0 +1,62 @@
...
+RUN mkdir /etc/rmt
+RUN printf "database:\n\
+  host: $MYSQL_HOST\n\
+  username: $MYSQL_USER\n\
+  password: $MYSQL_PASSWORD\n\
+  database: $MYSQL_DATABASE\n\
+scc:\n\
+  username: $SCC_USERNAME\n\
+  password: $SCC_PASSWORD\n\
+" >> /etc/rmt/rmt.conf
```
ikapelyukhin: That was changed in another PR.

features/mirror_spec.rb
```diff
@@ -0,0 +1,23 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
+
+describe 'mirror' do
+  let(:files) { %w'repomd.xml repomd.xml.asc repomd.xml.key *-primary.xml.gz *-filelists.xml.gz *-other.xml.gz' }
```
ikapelyukhin: No RPM files listed here -- will not fail if RPMs fail to mirror.

features/mirror_spec.rb
```diff
@@ -0,0 +1,23 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
```
ikapelyukhin: Not actually used in this file -- see my previous comment about using `%x` and `command`.

features/mirror_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  after do
+    `/usr/bin/rmt-cli repos disable 3114`
+
+    # cleanup files
+    FileUtils.rm_r('/var/lib/rmt/public/repo/SUSE/Updates/SLE-Product-SLES/15')
+  end
+
+  it do
+    files.each do |filename_pattern|
+      expect(`find /var/lib/rmt/public/ -name \'#{filename_pattern}\'`).to include(filename_pattern.gsub('*', ''))
```
ikapelyukhin: 1. This is still `bash`. (Ruby is `Dir.glob('/var/lib/rmt/public/**/*')`)
2. One expectation per example would be nicer:
```
files.each do |filename_pattern|
  it "mirrors #{filename_pattern} file" do
    expect(`find /var/lib/rmt/public/ -name \'#{filename_pattern}\'`).to include(filename_pattern.gsub('*', ''))
  end
end
```

features/repos_spec.rb
```diff
@@ -0,0 +1,22 @@
+require File.expand_path('../support/command_rspec_helper', __FILE__)
```
ikapelyukhin: Also not used here.

PR #211, opened 2018-08-15
Change file permissions for rmt.conf

Sergeykot:


package/rmt-server.spec
```diff
@@ -247,6 +247,9 @@ if [ $1 -eq 2 ]; then
   fi
 fi
 
+chmod 640 %{conf_dir}/rmt.conf
+chown _rmt %{conf_dir}/rmt.conf
```
thutterer: Couldn't you do the same with `%attr`
See https://stackoverflow.com/a/16951028/3449673 and http://ftp.rpm.org/max-rpm/s1-rpm-specref-files-list-directives.html (or better spend some more time googling this than I did :grin:)

Not an expert here, but this seems cleaner than doing this manually.

Sergeykot:
@thutterer I made more clean solution by mixing `%config` and `%attr`. Seems to work for package update as well. Would be nice to have another verification of the functionality though

thutterer:
> Seems to work for package update as well.

Not for me when I upgraded from the release version (1.0.0) on a fresh SLE15 to your 1.0.7
`/etc/rmt/rmt.conf` was still 644 afterwards.

When I install 1.0.7 on a fresh system (that had no previous version installed), the file permissions are correct (640).

Please coordinate with @ikapelyukhin if that is good enough.

ikapelyukhin:
@thutterer I think it might be caused by the changes @felixsch made to move the config files into a new location: `/etc/rmt/rmt.conf` doesn't exist, it's installed from the package with `0640` permissions, but then the old `/etc/rmt.conf` file is copied on top of it.

ikapelyukhin:
We could move `/etc/rmt/rmt.conf` back into `/etc/rmt.conf` :D

thutterer:
> I think it might be caused by the changes @felixsch made to move the config files into a new location

Oh yes, that is probably the reason.
Well, we could work around that by just doing a `chmod` on the `rmt.conf` in that if-block after it got moved? Not pretty, but then we contain the ugly `chmod` in that block at least.

ikapelyukhin:
> Well, we could work around that by just doing a `chmod` on the `rmt.conf` in that if-block after it got moved? Not pretty, but then we contain the ugly `chmod` in that block at least.

But that would mean more stuff in `%post`, not less :D

Sergeykot:
I thing reverting location of `rmt.conf` will be the best solution. From functionality perspective this does not change anything but saves us a lot of problemns

thutterer:
@Sergeykot I think Ivan was only joking when he suggested moving it back?? 
At least I would find that very inconsistent¹. What is wrong with adding 1 line to an if-block that already exists?
I'm tempted to "request changes" to prevent this revert, but I don't think it is reasonable to involve 4 people in what was supposed to be a one-liner.
So please coordinate with @ikapelyukhin _and_ @felixsch and figure out what works best for now and the future.

¹: If you really wanna go this way, please bare in mind, that this would also mean reverting changes in yast2-rmt and the SUSE documentation! Have a look at this [card](https://trello.com/c/9G37vU91/385-p1bzrmt-certificates-stored-in-wrong-place).

felixsch:
For me this would be really inconsistent to have `/etc/rmt.conf` and a additional directory `/etc/rmt/`. But it's true I forgot to check the file permissions when moving the configuration from the new to the old position.

felixsch:
Another thing why not just add your file permission changes to 1.0.6, since this bump is not released yet and only contains changes which are similar to yours (this also means I don't have to change `yast2-rmt` again)

Issue #112, opened 2018-03-12
Nicer error message when export directory isn't writable


ikapelyukhin:
As reported by @keichwa, when export directory isn't writable export command produced a traceback.
It would be have a nicer error message suggesting to make sure that the export directory is writable by RMT.

```
sudo rmt-cli export data /var/tmp/usb-ex
```

```
I, [2018-03-07T10:20:36.949116 #11559]  INFO -- : Exporting data from SCC to /var/tmp/usb-ex
I, [2018-03-07T10:20:36.949174 #11559]  INFO -- : Exporting products
Traceback (most recent call last):
        20: from /usr/bin/rmt-cli:40:in `<main>'
        19: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor/base.rb:444:in `start'
        18: from /usr/share/rmt/lib/rmt/cli/base.rb:22:in `dispatch'
        17: from /usr/share/rmt/lib/rmt/cli/base.rb:33:in `handle_exceptions'
        16: from /usr/share/rmt/lib/rmt/cli/base.rb:22:in `block in dispatch'
        15: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
        14: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
        13: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
        12: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor.rb:242:in `block in subcommand'
        11: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor/invocation.rb:115:in `invoke'
        10: from /usr/share/rmt/lib/rmt/cli/base.rb:22:in `dispatch'
         9: from /usr/share/rmt/lib/rmt/cli/base.rb:33:in `handle_exceptions'
         8: from /usr/share/rmt/lib/rmt/cli/base.rb:22:in `block in dispatch'
         7: from /usr/lib64/rmt/vendor/bundle/ruby/2.5.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
[...]

PR #226, opened 2018-09-13
Add: Writable option to `need_path`

simson-suse:
* add writable option to `need_path` to catch permission errors,
  which are handled more gracefully
  This fixes https://github.com/SUSE/rmt/issues/112

* for this, add function `process_user_name` to `RMT::CLI::Base` as a
  more descriptive wrapper of the actual call

* refactor `require` statements in `bin/rmt_cli` so they are all in the
  beginning and use new `process_user_name` function

### DIFF
--- a/bin/rmt-cli
+++ b/bin/rmt-cli
@@ -1,29 +1,27 @@
 #!/usr/bin/env ruby
 
-require 'etc'
-
 rmt_path = File.expand_path('..', __dir__)
 require_relative '../config/boot'
 $LOAD_PATH.unshift File.join(rmt_path, 'lib')
 
+require 'etc'
 require 'active_support'
 require 'active_record'
 require 'erb'
 require 'yaml'
 require 'rmt/config'
 
-# don't run as root
-if Process.euid == 0
+relative_load_paths = %w[lib lib/rmt app/models app/services].map { |dir| File.join(rmt_path, dir) }
+ActiveSupport::Dependencies.autoload_paths += relative_load_paths
+
+if RMT::CLI::Base.process_user_name == 'root'
   # set group and then user, otherwise user cannot change group
   Process::Sys.setegid(Etc.getgrnam(RMT::DEFAULT_GROUP).gid)
   Process::Sys.seteuid(Etc.getpwnam(RMT::DEFAULT_USER).uid)
 end
 
-relative_load_paths = %w[lib lib/rmt app/models app/services].map { |dir| File.join(rmt_path, dir) }
-ActiveSupport::Dependencies.autoload_paths += relative_load_paths
-
 if File.exist?(RMT::DEFAULT_MIRROR_DIR) && !File.writable?(RMT::DEFAULT_MIRROR_DIR)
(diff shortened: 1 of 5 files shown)


simson-suse:
This PR is for the CI stuff to work, the original one is https://github.com/SUSE/rmt/pull/225

suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bin/rmt-cli
```diff
@@ -1,29 +1,27 @@
...
+if RMT::CLI::Base.process_user_name == 'root'
   # set group and then user, otherwise user cannot change group
   Process::Sys.setegid(Etc.getgrnam(RMT::DEFAULT_GROUP).gid)
   Process::Sys.seteuid(Etc.getpwnam(RMT::DEFAULT_USER).uid)
 end
 
-relative_load_paths = %w[lib lib/rmt app/models app/services].map { |dir| File.join(rmt_path, dir) }
-ActiveSupport::Dependencies.autoload_paths += relative_load_paths
-
 if File.exist?(RMT::DEFAULT_MIRROR_DIR) && !File.writable?(RMT::DEFAULT_MIRROR_DIR)
```
ikapelyukhin: I still think this can be unified with `needs_path`, but otherwise LGTM.

bin/rmt-cli
```diff
@@ -1,29 +1,27 @@
...
 require 'erb'
 require 'yaml'
 require 'rmt/config'
 
-# don't run as root
-if Process.euid == 0
+relative_load_paths = %w[lib lib/rmt app/models app/services].map { |dir| File.join(rmt_path, dir) }
+ActiveSupport::Dependencies.autoload_paths += relative_load_paths
+
+if RMT::CLI::Base.process_user_name == 'root'
```
ikapelyukhin: I don't see why this change would be required.

bin/rmt-cli
simson-suse: Well, in theory I agree, in practice this doesn't work out so clean:
(1) Since `./bin/rmt-cli` is not inheriting from `RMT::CLI:base`, `needs_path` isn't available here, as it is not inside of `class << self` of `RMT::CLI::Base`. We could change this, but than we need to prefix every `needs_path` with `RMT::CLI::Base`
(2) Even if we did this, we still wouldn't get the nice error handling, since the Exceptions `needs_path` throws, are not caught by the Thor class

bin/rmt-cli
simson-suse: This change is not required, but with `process_user_name` it is now possible to write descriptive code without needing a comment, which I learned is the preferred way,

bin/rmt-cli
ikapelyukhin: (1) (2) Override `RMT::CLI::Main.start` and move into that method. `RMT::CLI::Main` inherits from `RMT::CLI::Base` and handles exceptions in a nice way.

ikapelyukhin: Indeed. But it was a fun thing to try. :grimacing::+1:

PR #265, opened 2018-11-14
Fix auto-selected modules for online migrations

Sergeykot:


package/rmt-server.changes
```diff
@@ -4,6 +4,7 @@ Tue Oct 30 15:35:17 UTC 2018 - skotov@suse.com
 - Version 1.0.7
 - Import migration_extra attribute from SCC API
 - Auto-selected modules feature migration engine changes
+- Restore old behavior for ASMM in online migrations
```
digitaltom: I don't think we need this line. Maybe the line before can be worded more customer friendly: 
`Offline migrations will automatically add additional modules to the client systems depending on the base product`

spec/models/migration_engine_spec.rb
```diff
@@ -435,4 +381,54 @@
...
+        ProductsExtensionsAssociation.find_by(
+          product: sle15,
+          extension: recommended_module,
+          root_product: sle15
+        ).update!(recommended: true)
+
+        recommended_module
+      end
+
+      it { is_expected.to contain_exactly([sle15, sle15_recommended_module]) }
```
digitaltom: the context name (`...modules do _not_ get added...`) and the test do not match

spec/models/migration_engine_spec.rb
```diff
@@ -435,4 +381,54 @@
...
+        ProductsExtensionsAssociation.find_by(
+          product: sle15,
+          extension: extra_module,
+          root_product: sle15
+        ).update!(migration_extra: true)
+
+        extra_module
+      end
+
+      it { is_expected.to contain_exactly([sle15, sle15_extra_module]) }
```
digitaltom: the context name (`...do _not_ get added...`) and the test do not match

Sergeykot: fixed

spec/models/migration_engine_spec.rb
Sergeykot: fixed

package/rmt-server.changes
Sergeykot: fixed

PR #267, opened 2018-11-15
Registration sharing plug-in

ikapelyukhin:
* Adds ActiveRecords after-commit hooks to `System` and `Activation` models
* The web-application creates a file with a timestamp on the disk when a system is modified (with flock)
* The rake task (ran by a systemd service/timer) picks up those files and syncs the current state of the system to the peer RMT server
* The files are removed (with flock) if the sync API call succeeded and haven't changed during sync
* An endpoint for legacy SMT XML-based API calls

Some differences compared to SMT:

* Web-workers don't block during registration API calls due to sync API calls (thus there should be no slow-down during registration)
* The latest system state is read from the DB when the system file is picked up (thus order doesn't matter)

Things that aren't there but that I'm pondering about:

* Delaying picking up freshly creates files by 1 or 2 minutes -- SLE15 registration requires activating multiple products, this would help avoid syncing not full state
* Some kind of check that would prevent overwriting more recent system state with an older state (timestamp check? But that would require correct time on all the servers)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

engines/registration_sharing/app/controllers/registration_sharing/smt_to_rmt_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    rescue SmtToRmtException => e
+      render status: :bad_request, plain: e.to_s
+    end
+
+    protected
+
+    def check_ip
+      allowed_ips = (Settings[:regsharing][:smt_allowed_ips] rescue [] || [])
+      allowed_ips.include?(request.remote_ip)
```
rjschwei: This is going to break in GCE. We moved GCE to a different model since their published IP address data is a mess. In GCE there is a script running on the platform that generates firewall rules. This script is run by the Google team and we do not restrict access on our end. We have to account for this case, either by allowing the check to be skipped via config, or by recognizing an entry such as 0.0.0.0/0

engines/registration_sharing/app/controllers/registration_sharing/smt_to_rmt_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+        captures = el.attribute('value').value.scan(/PRODUCTDATAID=(\d+)/)
+        hash[el.attribute('columnName').value] = captures[0][0] if captures[0]
+      end
+      hash
+    end
+
+    def smt_delete_registration(xml)
+      guid = xml.css('deleteRegistrationData guid')
+
+      raise 'System GUID not found' unless guid
```
rjschwei: I don't think we need an exception in this case. If we try to delete a system that doesn't exist then there is simply nothing to do. We should log a warning.

engines/registration_sharing/lib/registration_sharing.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    protected
+
+    def write_file(peer, system_login)
+      dirname = File.join(config_data_dir, peer)
+      FileUtils.mkpath(dirname)
+
+      filename = File.join(config_data_dir, peer, system_login)
+
```
rjschwei: Is system_login unique for every invocation of write_file? If not we have the same dilemma with flock here as discussed w.r.t. SMT.

rjschwei:
I have a concern with the taken approach in that this becomes very difficult to monitor. Writing with the data to be shared gets us away from slowing down the registration because of a synchronous https request to the sibling servers. However, by writing a file to share the data latter we can no longer use the existence of such a file as an indication that the sibling server is not responding for some reason. The change in logic from a monitoring perspective can certainly be addressed, we just have to keep this in mind.

engines/registration_sharing/app/controllers/registration_sharing/smt_to_rmt_controller.rb
ikapelyukhin: That's the analogue of `acceptRegistrationSharing` in `/etc/smt.conf` -- I see that it is set on GCE SMTs. This parameter holds IPs of SMTs which RMT will accept shared registrations from. For restricting client IP ranges I've added optional nginx config includes as part of `strict_authentication` plugin.

RMT to RMT sharing uses a shared secret instead (due to reverse-lookups and other considerations from the latest SMT CVE).

engines/registration_sharing/app/controllers/registration_sharing/smt_to_rmt_controller.rb
ikapelyukhin: This raises the exception only if the XML is malformed (`guid` tag is missing). If the system with the supplied GUID isn't in the DB -- the API call will succeed.

engines/registration_sharing/lib/registration_sharing.rb
ikapelyukhin: `write_file` dumps the timestamp (with microseconds) into the file. After the registration is shared, the process would lock the file and read it's contents -- if the file was modified during regsharing call, it will not be unlinked and will be synced again in the next iteration. If the file wasn't modified, it is removed.

ikapelyukhin:
@rjschwei Indeed, we'd have to adjust the monitoring. I'm hoping it wouldn't be too hard, though -- with individual files we know exactly how many systems are in the backlog. We could raise alerts when the number of files is greater than, say, 25 (arbitrary number).

engines/registration_sharing/app/controllers/registration_sharing/smt_to_rmt_controller.rb
rjschwei: OK, yes now that you say it it is obvious, must have had something else in my head at the time.

rjschwei:
W.r.t. monitoring we'll have to write something somewhere else, I think. If RMT would notice that the registration sharing requests are not being accepted. Then write a new entry into /var/log/regsharingerrors

That would keep the monitoring approach the same while also allowing the RMT server to use the log-replay logic you are perusing.

engines/registration_sharing/lib/registration_sharing.rb
ikapelyukhin: `system_login` is the login of the client system that had it's registration data modified. So yes, it will create a unique file for each individual system.

It is possible for the system to change while it's being synced (an additional product could be activated). That's why the file isn't locked when API requests are being made (so that the web workers don't block in `flock` call). If the file has been modified during sync, it won't be deleted and will be re-synced in the next iteration.

ikapelyukhin:
@rjschwei Sure, I can implement something on RMT side specifically for monitoring. Should we address the monitoring bits now? I wouldn't mind looking into monitoring setup and deployment automation, I think after doing initial testing some adjustments would be required -- we could address it there and then.

engines/registration_sharing/lib/registration_sharing.rb
rjschwei: OK

engines/registration_sharing/lib/registration_sharing/client.rb
```diff
@@ -0,0 +1,58 @@
...
+        'User-Agent' => "RMT::Regsharing/#{RMT::VERSION}"
+      },
+      body: JSON.dump(params),
+      capath: RegistrationSharing.config_ca_path
+    )
+
+    response = request.run
+
+    unless response.success?
+      raise "Regsharing request failed with code #{response.code}: #{response.body}"
```
rjschwei: Wherever this is caught that's where we need to write some indicator that we can pick up in monitoring.

rjschwei:
I added a comment about the monitoring here: https://github.com/SUSE/rmt/pull/267/files#diff-ef4cb66273717e05b341f48ce09b9e2dR52

I am OK with proceeding with this and then adding code we can use as a monitoring trigger later.

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+module RegistrationSharing
+  class RmtToRmtController < ApplicationController
+
+    before_action :authenticate
+
+    def create
+      System.transaction do
+        system = RegistrationSharing::System.find_or_create_by(login: params[:login])
+
+        %w[password hostname registered_at created_at last_seen_at].each do |attribute|
```
bear454: Why not use [strong parameters](https://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters) here?

engines/registration_sharing/lib/registration_sharing/client.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  protected
+
+  def peer_register_system(system)
+    params = {}
+
+    %w[login password hostname registered_at created_at last_seen_at].each do |attribute|
+      params[attribute] = system.send(attribute)
```
bear454: Consider using a decorator to construct the JSON-ready hash.

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
ikapelyukhin: Yes, good idea.

PR #273, opened 2018-11-22
SLP support

wstephenson:


package/rmt-server.spec
```diff
@@ -50,6 +50,7 @@ Source14:       nginx-https.conf
 Source15:       auth-handler.conf
 Source16:       auth-location.conf
 Source17:       rmt-cli_bash-completion.sh
+Source18:       rmt-server.reg
```
ikapelyukhin: In the hindsight, adding `SourceN` items was perhaps not the best idea. It is easier to just throw those files into the tarball and install from there, rather than maintaining the list of sources.

Thoughts on moving those into the tarball? (I can do that later in a separate PR for some of those sources)

package/rmt-server.spec
```diff
@@ -128,6 +129,7 @@ mkdir -p %{buildroot}%{_bindir}
 ln -s %{app_dir}/bin/rmt-cli %{buildroot}%{_bindir}
 ln -s %{app_dir}/bin/rmt-data-import %{buildroot}%{_bindir}/rmt-data-import
 install -D -m 644 %{_sourcedir}/rmt-cli.8.gz %{buildroot}%{_mandir}/man8/rmt-cli.8.gz
+install -D -m 644 %{_sourcedir}/rmt-cli.8.gz %{buildroot}%{_mandir}/man8/rmt-cli.8.gz
```
ikapelyukhin: @wstephenson Looks like unintentional copy.

package/rmt-server.changes
```diff
@@ -1,3 +1,7 @@
+Thu Nov 22 11:23:18 UTC 2018 - wstephenson@suse.com
+
+- Register to SLP if a server running on this host
```
fschueller: Please append `bsc#1113760`

package/rmt-server.spec
fschueller: I agree, let's do this in a separate PR.

PR #278, opened 2018-11-28
Fix products wrongly showing as enabled for mirroring

simson-suse:
See Bugzilla bug #1102193 for reference.

To avoid confusion: When a repository is enabled or not, we talk about
zypper (RPM) repositories, which ship with a default, either `enabled`
or `disabled`. As an example, typically "debug symbols" are `disabled`
by default.

Before, if there were no `enabled` repositories for a product (which
happens for very old products), `mirror?` would return true even though
there was nothing mirrored. This kinda made sense since we accept a
__product__ as `enabled` when all the __repos__ that are `enabled` by
default are "enabled for mirroring", because that's usually everything
one needs (by definition, since they other repos on the clients are
`disabled` by default anyway).

Now it will first check if there are repositories that are `enabled` by
default. And if that is true, it then checks if these are all "enabled
for mirroring". If there are no repositories that are `enabled` by
default, it will check if there are repositories that are `disabled` by
default, and consider the product `enabled` if all those repositories
are "enabled for mirroring". If there are no repositories at all, it
will just return `false`, since a product without repositories is not
understood by RMT anyway.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -66,16 +66,14 @@ class Product < ApplicationRecord
   end
 
   def mirror?
+    enabled = repositories.select(&:enabled).to_a
+    disabled = repositories.reject(&:enabled).to_a
 
-    enabled_repositories = repositories.where(enabled: true)
-    unless enabled_repositories.empty?
-      return enabled_repositories.where(mirroring_enabled: false).empty?
-    end
+    # If we have enabled repositories, return true if any are enabled to be mirrored.
+    return enabled.reject(&:mirroring_enabled).empty? unless enabled.empty?
 
-    disabled_repositories = repositories.where(enabled: false)
-    unless disabled_repositories.empty?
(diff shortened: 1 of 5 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

simson-suse:
If we agree on this change, I will add the corresponding tests to rspec.

app/models/product.rb
```diff
@@ -66,16 +66,14 @@ def has_extension?
   end
 
   def mirror?
+    enabled = repositories.select(&:enabled)
```
felixsch: Just nitpicking, how about adding the `.to_a` here and remove it from the other lines?

tmuntaner: You are so very right. Updated!

PR #279, opened 2018-11-30
Sync products without base product

Sergeykot:
Some of the products do not have base one (for example SUMA). However, it should be possible to mirror some of them.

Right now products like that are filtered out by RMT during `sync` (SMT does not).

I suggest that we remove the check which syncs products only when they are base products so that it would be possible to mirror them.

### DIFF
--- a/lib/rmt/scc.rb
+++ b/lib/rmt/scc.rb
@@ -21,9 +21,8 @@ class RMT::SCC
 
     @logger.info(_('Updating products'))
     data = scc_api_client.list_products
-    data.each { |item| create_product(item) if (item[:product_type] == 'base') }
-    # with this loop, we create the migration paths after creating all products from sync, thus avoiding fk_constraint errors
-    data.each { |item| migration_paths(item) if (item[:product_type] == 'base') }
+    data.each { |item| create_product(item) }
+    data.each { |item| migration_paths(item) }
 
     update_repositories(scc_api_client.list_repositories)
 
@@ -144,14 +143,7 @@ class RMT::SCC
     uri = URI(item[:url])
     auth_token = uri.query
 
-    # Sometimes the extension is available, but a base product is not, e.g.:
-    # sle-hae/11.3/s390x available without base product for s390x
-    # In this case no repository data was added in create_product -- can't update those repos.
-    begin
-      Repository.find_by!(scc_id: item[:id]).update! auth_token: auth_token
-    rescue ActiveRecord::RecordNotFound
-      @logger.debug _('Repository %{repo} is not available') % { repo: item[:id] }
-    end
+    Repository.find_by!(scc_id: item[:id]).update! auth_token: auth_token
   end
 
   def create_subscription(item)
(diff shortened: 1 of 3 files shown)


Sergeykot:
@ikapelyukhin please review this suggestion when you have time

lib/rmt/scc.rb
```diff
@@ -129,7 +128,7 @@ def create_product(item, root_product_id = nil, base_product = nil, recommended
       ProductsExtensionsAssociation.where(root_product_id: root_product_id).destroy_all
     end
 
-    item[:extensions].each do |ext_item|
+    item[:extensions]&.each do |ext_item|
```
ikapelyukhin: Why wouldn't a product item have `:extensions` key? It's the same serializer for everything.

lib/rmt/scc.rb
```diff
@@ -21,8 +21,7 @@ def sync
 
     @logger.info(_('Updating products'))
     data = scc_api_client.list_products
-    data.each { |item| create_product(item) if (item[:product_type] == 'base') }
-    # with this loop, we create the migration paths after creating all products from sync, thus avoiding fk_constraint errors
+    data.each { |item| create_product(item) }
     data.each { |item| migration_paths(item) if (item[:product_type] == 'base') }
```
ikapelyukhin: Not sure if old logic was left here intentionally or not. Products without base won't have migrations, but maybe that's OK.

package/rmt-server.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Fri Nov 30 16:50:06 UTC 2018 - skotov@suse.com
+
+- Sync product without base product
```
ikapelyukhin: Needs bug number reference.

lib/rmt/scc.rb
Sergeykot: spec issue. fixed

lib/rmt/scc.rb
Sergeykot: does not apply to any of those product currently, but yeah it was not intentional
some products without base _could_ have migration path. fixed

Sergeykot:
had to rebase from master because of changelog conflict

Sergeykot:
@fschueller I updated the PR description and made changes to the code as well. I think it is ready for review again

lib/rmt/scc.rb
```diff
@@ -171,8 +163,8 @@ def create_subscription(item)
   def migration_paths(item)
     product = get_product(item[:id])
     ProductPredecessorAssociation.where(product_id: product.id).destroy_all
-    create_migration_path(product, item[:online_predecessor_ids], :online)
-    create_migration_path(product, item[:offline_predecessor_ids], :offline)
+    create_migration_path(product, item[:online_predecessor_ids], :online) if item[:online_predecessor_ids]
+    create_migration_path(product, item[:offline_predecessor_ids], :offline) if item[:offline_predecessor_ids]
```
ikapelyukhin: Wouldn't the serialized JSON have an empty array if there are no predecessors, and thus be always truthy?

fschueller: I believe that is the case. We should check for contents of the array, possibly something like `if !item[:offline_predecessor_ids].empty?`

Sergeykot:
does it also even make sense to import migration paths for the products without base? @ikapelyukhin @fschueller 

I think that even if there will be any, the migration engine will never return them

lib/rmt/scc.rb
Sergeykot: issue with the spec, it was `nil` because wrong spec setup. fixed

Sergeykot:
@ikapelyukhin @fschueller I checked the products without a base and currently they all do not have migration targets, however they could have them. so that we will import those if this would be the case. 

also fixed the issue with the spec setup with `offline_predecessor_ids` and `online_predecessor_ids`

PR #284, opened 2018-12-04
Added the ability to enable multiple repositories at the same time

tmuntaner:


spec/lib/rmt/cli/repos_spec.rb
```diff
@@ -10,10 +10,43 @@
...
+      let(:argv) { ['enable', repository.scc_id.to_s, repository_2.scc_id.to_s, repository_3.scc_id.to_s] }
+      let(:expected_output) do
+        <<-OUTPUT
+Repository by id #{repository.scc_id} successfully enabled.
+Repository by id #{repository_2.scc_id} successfully enabled.
+Repository by id #{repository_3.scc_id} successfully enabled.
+OUTPUT
+      end
+
+      it 'ensables repository' do
```
fschueller:
```suggestion
      it 'enables repository' do
```

tmuntaner: This should be fixed now ^_^

PR #291, opened 2018-12-06
Updated the manpage for new features

tmuntaner:


MANUAL.md
```diff
@@ -33,11 +33,33 @@ You can install and run this wizard like this:
 
     Use the `--csv` flag to output the list in CSV format.
 
-  * `rmt-cli products enable <id | string>`:
-    Enables mandatory repositories of a single product by its id or identifier string.
+  * `rmt-cli products enable <id | string>...`:
```
fschueller: Why `...` here but not in L39 "`rmt-cli products disable <id | string>`:" ?

tmuntaner: Added

PR #309, opened 2019-01-18
Fix product version display

Sergeykot:
Fixes product version display:

* in migration endpoints (we already have `friendly_name` there so just had to delete `Product#frinedly_version` method which used `version`)
* added syncing `friendly_version` and changed CLI output accordingly


app/models/product.rb
```diff
@@ -87,8 +87,8 @@ def self.clean_up_version(version)
     [version, version.tr('-', '.').chomp('.0')].uniq
   end
 
-  def friendly_name
-    "#{name} #{version} #{arch}"
+  def safe_friendly_version
+    friendly_version ? friendly_version : version
```
thutterer: `friendly_version || version`  ?

Sergeykot: nice catch! fixed

PR #319, opened 2019-01-31
Fix HTTP vhost config

ikapelyukhin:
- Redirect /repo/repoindex.xml from HTTP to HTTPS
- Include auth subrequest into /repo location

### DIFF
--- a/package/nginx-http.conf
+++ b/package/nginx-http.conf
@@ -5,11 +5,23 @@ server {
     error_log   /var/log/nginx/rmt_http_error.log;
     root        /usr/share/rmt/public;
 
+    include /etc/nginx/rmt-auth*.d/auth-location*.conf;
+
     location / {
         autoindex off;
     }
 
     location /repo {
         autoindex on;
+        include /etc/nginx/rmt-auth*.d/auth-handler*.conf;
+    }
+
+    location = /repo/repoindex.xml {
+        return 301 https://$host$request_uri;
+    }
+
+    # An alias to RMT CA certificate, so that it can be downloaded to client machines.
+    location /rmt.crt {
+        alias /etc/rmt/ssl/rmt-ca.crt;
     }
 }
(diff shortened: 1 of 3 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ikapelyukhin:
@felixsch OK, this one should be good to go.

package/nginx-http.conf
```diff
@@ -5,11 +5,23 @@ server {
...
         autoindex on;
+        include /etc/nginx/rmt-auth*.d/auth-handler*.conf;
+    }
+
+    location = /repo/repoindex.xml {
+        return 301 https://$host$request_uri;
+    }
+
+    # An alias to RMT CA certificate, so that it can be downloaded to client machines.
+    location /rmt.crt {
```
tmuntaner: Do we really need this part?

ikapelyukhin: Robert has asked to make the certificate available over HTTP as well (to avoid handling SSL exceptions on the client side -- `rmt-client-setup` isn't used in the cloud).

tmuntaner: @digitaltom and I talked and this should be fine, as the supported setup is through https only.

ikapelyukhin:
@tmuntaner Using redirect from HTTP to HTTPS resulted in having wrong 'credentials' parameter in repo URLs (with `https` scheme). As the result, zypper is looking for the wrong file in `/etc/zypp/credentials.d`.

I've fixed it to always use HTTP in the service name/`credentials` parameter.

package/nginx-http.conf
ikapelyukhin: OK, cool. Thanks.

PR #325, opened 2019-02-11
Truncate PID in lockfile

Sergeykot:
If the old PID has more digits than the current one user would see the wrong PID because the file is not truncated but only first digits are rewritten. This PR fixes that.

See also https://apidock.com/ruby/File/flock for an example.

The fix can be verified by writing some content into `/tmp/rmt.lock` which should be longer that the PID and then running 2 RMT CLI processes in 2 console tabs. You should see correct PID in the error message.

### DIFF
--- a/lib/rmt/lockfile.rb
+++ b/lib/rmt/lockfile.rb
@@ -8,7 +8,8 @@ class RMT::Lockfile
       File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT) do |f|
         if f.flock(File::LOCK_EX | File::LOCK_NB)
           f.write(Process.pid.to_s)
-          f.fsync
+          f.flush
+          f.truncate(f.pos)
         else
           pid = File.read(RMT::Lockfile::LOCKFILE_LOCATION)
           raise ExecutionLockedError.new(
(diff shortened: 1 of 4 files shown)


lib/rmt/lockfile.rb
```diff
@@ -8,7 +8,8 @@ def lock
       File.open(RMT::Lockfile::LOCKFILE_LOCATION, File::RDWR | File::CREAT) do |f|
         if f.flock(File::LOCK_EX | File::LOCK_NB)
           f.write(Process.pid.to_s)
-          f.fsync
+          f.flush
+          f.truncate(f.pos)
```
digitaltom: Is it possible to test this?

Sergeykot: see PR description :slightly_smiling_face:

Sergeykot: oh if you mean making a spec, I would suggest adding a feature spec for that.

ikapelyukhin:
Just use `File.open` with `File::TRUNC`?
(`File::RDWR | File::CREAT | File::TRUNC`).

lib/rmt/lockfile.rb
Sergeykot: added specs :+1:

Sergeykot:
@ikapelyukhin I tried it, and the thing is that 'File::TRUNC' truncates the file _before_ lock, so that even if `File#flock` returns false the pid is gone and user does not see it in error message.

Sergeykot:
added specs and updated, ready for review

PR #326, opened 2019-02-11
Improve handling of Settings

### LINKED ISSUE #178: Improve handling of Settings
This discussion started here: https://github.com/SUSE/rmt/pull/177

I'm not a fan of `try` (nor ruby's safe operator `.&`)..
It basically makes code insecure about itself... is the thing there? or is it not? :scream: if it's not, you have to check for `nil`s all over the place. => Sadness.
When working with the `Settings` object, our code is littered with `Settings.try(:foo)` to check if the settings are really OK.

A few things that can be done to improve this:

- Never use the `Settings` class directly (which is the low-level construct provided by the [railsconfig/config](https://github.com/railsconfig/config) gem), only from `lib/rmt/config.rb`. We have some settings accessed through `Settings` and yet others through `RMT::Config`, which is inconsistent. We can encapsulate all the `try`s and `nil`s n `RMT::Config`, leaving the rest of the code confident about itself.
- [Prepend a default config](https://github.com/railsconfig/config#adding-sources-at-runtime) to the `config` gem, which is included somewhere in the RMT sources, ie. it's not user-modifiable. By staring with a clean, default config, then the `config` gem will override the settings that are user-provided.
- Use the `config` gem's [validation](https://github.com/railsconfig/config#validation) feature, checking whether the loaded settings are correct.

gionaufal:
Fixes #178, using `RMT::Config` instead of `Settings`, encapsulating the error in `RMT::Config` and using `dry-validation`'s schema.

I also tried to change how the settings are being handled in `engines`, but as the tests were already breaking and I don't know if this is expected, I didn't touch the files and tests in this folder.

### DIFF
--- a/lib/rmt/config.rb
+++ b/lib/rmt/config.rb
@@ -4,10 +4,12 @@ require_relative '../rmt'
 Config.setup do |config|
   config.merge_nil_values = false
 
-  config.schema do
-    required(:scc).schema do
-      required(:username).filled
-      required(:password).filled
+  unless ENV['RAILS_ENV'] == 'test'
+    config.schema do
+      required(:scc).schema do
+        required(:username).filled
+        required(:password).filled
+      end
     end
   end
 end
--- a/lib/rmt/config.rb
+++ b/lib/rmt/config.rb
@@ -25,7 +25,7 @@ module RMT::Config
     end
   end
 
-  def self.respond_to_missing?
+  def self.respond_to_missing?(method, _)
     super
   end
 
--- a/lib/rmt/config.rb
+++ b/lib/rmt/config.rb
@@ -12,6 +12,15 @@ Config.load_and_set_settings(
 )
 
 module RMT::Config
+  def self.method_missing(key)
+    raise NoMethodError if Settings[key].nil?
+    Settings[key]
+  end
+
+  def self.respond_to_missing?(method, *args)
+    method == :scc or super
+  end
+
   def self.db_config(key = 'database')
     {
       'username' => Settings[key].username,
@@ -31,5 +40,4 @@ module RMT::Config
   def self.deduplication_by_hardlink?
     Settings.try(:mirroring).try(:dedup_method).to_s.to_sym != :copy
   end
-
 end
--- a/lib/rmt/config.rb
+++ b/lib/rmt/config.rb
@@ -3,9 +3,17 @@ require_relative '../rmt'
 
 Config.setup do |config|
   config.merge_nil_values = false
+
+  config.schema do
+    required(:scc).schema do
+      required(:username).filled
+      required(:password).filled
+    end
+  end
 end
 
 Config.load_and_set_settings(
+  File.join(__dir__, '../../config/default_rmt.yml'),
   '/etc/rmt.conf',
   File.join(__dir__, '../../config/rmt.yml'),
   File.join(__dir__, '../../config/rmt.local.yml')
--- a/lib/rmt/config.rb
+++ b/lib/rmt/config.rb
@@ -4,13 +4,8 @@ require_relative '../rmt'
 Config.setup do |config|
   config.merge_nil_values = false
 
-  unless ENV['RAILS_ENV'] == 'test'
-    config.schema do
-      required(:scc).schema do
-        required(:username).filled
-        required(:password).filled
-      end
-    end
+  config.schema do
+    required(:scc).schema
   end
 end
 
@@ -23,12 +18,15 @@ Config.load_and_set_settings(
 
 module RMT::Config
   def self.method_missing(key)
-    raise NoMethodError if Settings[key].nil?
-    Settings[key]
+    if Settings[key]
+      Settings[key]
+    else
+      super
+    end
   end
 
-  def self.respond_to_missing?(method, *args)
-    method == :scc or super
+  def self.respond_to_missing?
+    super
   end
 
   def self.db_config(key = 'database')
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -9,11 +9,12 @@ class RMT::HttpRequest < Typhoeus::Request
 
     super
 
-    options[:proxy] = Settings.try(:http_client).try(:proxy)
-    options[:proxyauth] = Settings.try(:http_client).try(:proxy_auth) ? Settings.try(:http_client).try(:proxy_auth).to_sym : :any
+    options[:proxy] = RMT::Config.http_client.proxy
 
-    if Settings.try(:http_client).try(:proxy_user) && Settings.try(:http_client).try(:proxy_password)
-      options[:proxyuserpwd] = "#{Settings.http_client.proxy_user}:#{Settings.http_client.proxy_password}"
+    options[:proxyauth] = RMT::Config.http_client.proxy_auth ? RMT::Config.http_client.proxy_auth.to_sym : :any
+
+    if RMT::Config.http_client.proxy_user && RMT::Config.http_client.proxy_password
+      options[:proxyuserpwd] = "#{RMT::Config.http_client.proxy_user}:#{RMT::Config.http_client.proxy_password}"
     end
   end
 
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -5,7 +5,7 @@ class RMT::HttpRequest < Typhoeus::Request
 
   def set_defaults
     Typhoeus::Config.user_agent = "RMT/#{RMT::VERSION}"
-    Typhoeus::Config.verbose = Settings.try(:http_client).try(:verbose)
+    Typhoeus::Config.verbose = RMT::Config.http_client.verbose
 
     super
 
--- a/config/default_rmt.yml
+++ b/config/default_rmt.yml
@@ -1,0 +1,25 @@
+database:
+  host: localhost
+  database: rmt
+  username: rmt
+  password:
+  adapter: mysql2
+  encoding: utf8
+  timeout: 5000
+  pool: 5
+
+scc:
+  username:
+  password:
+
+mirroring:
+  mirror_src: false
+  verify_rpm_checksums: false
+  dedup_method: hardlink
+
+http_client:
+  verbose: false
+  proxy:
+  proxy_auth:
+  proxy_user:
+  proxy_password:
--- a/lib/rmt/scc.rb
+++ b/lib/rmt/scc.rb
@@ -17,7 +17,7 @@ class RMT::SCC
     cleanup_database
 
     @logger.info(_('Downloading data from SCC'))
-    scc_api_client = SUSE::Connect::Api.new(Settings.scc.username, Settings.scc.password)
+    scc_api_client = SUSE::Connect::Api.new(RMT::Config.scc.username, RMT::Config.scc.password)
 
     @logger.info(_('Updating products'))
     data = scc_api_client.list_products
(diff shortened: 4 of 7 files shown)


lib/rmt/http_request.rb
```diff
@@ -9,11 +9,12 @@ def set_defaults
```
bgeuken: There is one more `try` to get rid of. It's one line above this one:

` Typhoeus::Config.verbose = Settings.try(:http_client).try(:verbose)`

ikapelyukhin:
Tests are dying with:

```
Config::Validation::Error: Config validation failed:
scc.username: must be filled\n  scc.password: must be filled
```

Actually, username and password aren't mandatory in airgap mode.

bgeuken:
Thanks for your collaboration:bouquet:

lib/rmt/http_request.rb
gionaufal: removed :)

gionaufal:
I've add a condition to check for the RAILS_ENV in `RMT::Config`. If feels hackish, but for now it's one way I thought for working. I didn't use `Rails.env.test?` because, for this, I'd have to require `rails` in the file.

The CI is still breaking, it seems because of Danger. Is there anything I can do to fix this?

lib/rmt/config.rb
```diff
@@ -3,15 +3,34 @@
 
 Config.setup do |config|
   config.merge_nil_values = false
+
+  unless ENV['RAILS_ENV'] == 'test'
+    config.schema do
+      required(:scc).schema do
+        required(:username).filled
+        required(:password).filled
```
ikapelyukhin: As I've mentioned in my previous comment, SCC username and password actually aren't always required. Yast RMT module allows skipping setting username and password [as of late](https://github.com/yast/yast-rmt/pull/44), which is required for airgap setups. (Also not having to set SCC username and passwords might come handy for openSUSE use cases)

lib/rmt/config.rb
```diff
@@ -3,15 +3,34 @@
...
 )
 
 module RMT::Config
+  def self.method_missing(key)
+    raise NoMethodError if Settings[key].nil?
+    Settings[key]
+  end
+
+  def self.respond_to_missing?(method, *args)
+    method == :scc or super
```
ikapelyukhin: Seems like this shouldn't be needed in the current implementation -- there's no way `RMT::Config.scc` could be missing: 1) it's marked as required in schema validation; 2) it's declared in the default config.

lib/rmt/config.rb
```diff
@@ -3,15 +3,34 @@
...
+      required(:scc).schema do
+        required(:username).filled
+        required(:password).filled
+      end
+    end
+  end
 end
 
 Config.load_and_set_settings(
+  File.join(__dir__, '../../config/default_rmt.yml'),
```
ikapelyukhin: My concern would be that this file is can be modified by the users.

lib/rmt/config.rb
```diff
@@ -3,15 +3,34 @@
...
   '/etc/rmt.conf',
   File.join(__dir__, '../../config/rmt.yml'),
   File.join(__dir__, '../../config/rmt.local.yml')
 )
 
 module RMT::Config
+  def self.method_missing(key)
+    raise NoMethodError if Settings[key].nil?
```
ikapelyukhin: `try` guaranteed safe navigation, this actually might raise exceptions. Although there's probably no way to address this.

lib/rmt/config.rb
gionaufal: Actually I implemented this because the tests were breaking without it. But I agree, I'll try other ways to mock or extend this class specifically to remove this method here

lib/rmt/config.rb
gionaufal: Understood! So I'll remove the `filled` requirements from the schema.

ikapelyukhin: I think that will fix the tests :-)

lib/rmt/config.rb
gionaufal: Removed, just kept the implementation of the method falling back to `super` because of `rubocop`.
https://github.com/rubocop-hq/ruby-style-guide#no-method-missing

lib/rmt/config.rb
gionaufal: I've changed this implementation a bit because `rubocop` was complaining about the lack of falling back to `super` (https://github.com/rubocop-hq/ruby-style-guide#no-method-missing), but the results are the same.

lib/rmt/config.rb
gionaufal: Could you suggest a better place for this file, or a way that the user couldn't modify? :)

ikapelyukhin: Yeah, I'm not sure -- I'm still thinking about this. Apparently, `dry-validation` doesn't allow specifying default values. However, `dry-initializer` does -- perhaps it would be nicer to have explicitly defined config methods, rather than relying on `method_missing?`. But I'm not sure how well it handles nested config keys (as they are currently defined) and how well that would work in general. :-)

gionaufal:
I've made the changes and the tests are passing, but the build seems to be breaking because the lack of Danger env variables.

```
For your GitHub repo, you need to expose: DANGER_GITHUB_API_TOKEN
You may also need: DANGER_GITHUB_HOST, DANGER_GITHUB_API_BASE_URL, DANGER_OCTOKIT_VERIFY_SSL
```

I looked into other successful builds and in fact my build is not setting the variable, skipping this step:

```
Setting environment variables from repository settings
$ export DANGER_GITHUB_API_TOKEN=[secure]
``` 

It seems like I'm missing something, but I couldn't find out what.

ikapelyukhin:
@gionaufal Hmm, yes, I think Travis CI doesn't expose the token to PRs coming from outside of `SUSE/rmt` repo.

digitaltom:
We changed the `danger` config not to run on pull requests from forks.

ikapelyukhin:
OK, let's compare how `RMT::Config` to how `Settings` behaved previously:

```
2.5.1 :001 > RMT::Config.foofoo
        2: from (irb):1
        1: from lib/rmt/config.rb:24:in `method_missing'
NoMethodError (undefined method `foofoo' for RMT::Config:Module)
2.5.1 :002 > Settings.foofoo
 => nil 
```

OK, so now `RMT::Config` raises an exception for missing keys -- it didn't before.
Previously for optional parameters you could have written `do_stuff() if Settings.foofoo` -- with `RMT::Config` you can't do that anymore.

``` 
2.5.1 :003 > RMT::Config.foofoo.barbar
        2: from (irb):3
        1: from lib/rmt/config.rb:24:in `method_missing'
NoMethodError (undefined method `foofoo' for RMT::Config:Module)
2.5.1 :004 > Settings.foofoo.barbar
        1: from (irb):4
NoMethodError (undefined method `barbar' for nil:NilClass)
```

So accessing nested options still raises exceptions, which was the whole reason we've used `try` and safe navigation operator -- nothing changed here.

```
2.5.1 :005 > RMT::Config[:http_client]
        2: from (irb):5
        1: from lib/rmt/config.rb:20:in `method_missing'
ArgumentError (wrong number of arguments (given 2, expected 1))
2.5.1 :006 > Settings[:http_client]
 => #<Config::Options verbose=false, proxy=nil, proxy_auth=nil, proxy_user=nil, proxy_password=nil> 
```

Accessing settings with `[]` doesn't work anymore, but OK, that's not that important.

So basically the only difference is that now it raises an exception on missing keys. We can achieve exactly the same behavior with `Settings` by setting `fail_on_missing` option to `true`:

```
Config.setup do |config|
  config.merge_nil_values = false
  config.fail_on_missing = true
end
```

```
2.5.1 :001 > Settings.foofoo
        2: from (irb):1
        1: from config (1.7.1) lib/config/options.rb:173:in `method_missing'
KeyError (key not found: :foofoo)
```

felixsch:
I think the method chain (`Config.foo.bar`) should be lazy evaluated (only at the end of the chain). This would allow to check the chain and the values on evaluation. But this also requires something like a end of chain method to mark the end of an actual method chain. For me would look like something like: `RMT::Config.some.settings.foo.value` where `value` is the end of chain method.

This would almost preserve the initial idea and allows for method chaining and checking the values.

What do you think @gionaufal @ikapelyukhin ?

ikapelyukhin:
@felixsch `RMT::Config.some.settings.foo.value` -- we could. Or we could just use the safe navigation operator (no additional coding required!), I think both of those options are on the same level of prettiness/hackiness :-)

I would actually prefer to have some stricter way of defining the schema and default values -- currently you just put some values into the config file, but there's no real list of all possible config values. The issue with `default_rmt.yml` is that everybody must not forget to put default values there as new config options are added. That can be somewhat addressed by enabling `fail_on_missing` (although that can backfire, because you can define a variable in your local config -- and that would suppress exceptions on your local workstation only).

So in my ideal world there would be some DSL (or just some classes + methods) for defining config schema and default values, e.g.:

```ruby
RMT::Config.schema do
  http_client {
    verbose false # defaults to false
    proxy # defaults to nil
    potato_mode {
      enable "yes" # defaults to "yes"
    }
  }
end
```

And then the Yaml config files kick in on top of that, but keys that are missing in that schema should always raise exceptions (even if some additional keys are defined in the Yaml config) -- this would enforce keeping this schema up-to-date. This would get us a nice strict schema, the default values and also make navigating this config safe and secure.

(However actually implementing this probably requires creating a new gem -- so what we actually want to do is up for discussion :D)

gionaufal:
@ikapelyukhin 
> So in my ideal world there would be some DSL (or just some classes + methods) for defining config schema and default values...

That would be great. It looks like there's been some conversation about adding default values to a schema in `dry-validation`, but it seems that it hasn't been implemented yet.

I believe issue's initial intention (please correct me if I'm wrong) was to create a standard way of accessing the settings and remove the use of the safe navigation operator. 

To try to solve the issue with people forgetting to add new config options to the default file I'm thinking in adding first only `default_rmt.yml` (or any other default file) and, after the validation, adding the other ones. I'm not really sure this is possible, but I'll test it. Do you think this would be a good approach?

ikapelyukhin:
@gionaufal 

> To try to solve the issue with people forgetting to add new config options to the default file I'm thinking in adding first only `default_rmt.yml` (or any other default file) and, after the validation, adding the other ones.

I'm not sure if that would address the concerns I've described above. If you add `foofoo: 1` to your `rmt.yml` and nowhere else, then `default_rmt.yml` will validate just fine, and then `foofoo` will get merged into `Settings` -- and `Settings.foofoo` will work just fine in the code.

@digitaltom @lagartoflojo please feel free to chime in with your thoughts on the matter.

digitaltom:
> My concern would be that this file is can be modified by the users.

How about moving the default file somewhere else to indicate that this is an immutable file? 

> OK, so now RMT::Config raises an exception for missing keys -- it didn't before.
> Previously for optional parameters you could have written do_stuff() if Settings.foofoo -- with 
> RMT::Config you can't do that anymore.

> @felixsch RMT::Config.some.settings.foo.value -- we could. Or we could just use the safe navigation 
> operator (no additional coding required!), I think both of those options are on the same level of 
> prettiness/hackiness :-)

Isn't it more a feature than a problem that our code raises an exception now on accessing config values that are not defined in `default_rmt.yml`? 

> I'm not sure if that would address the concerns I've described above. If you add foofoo: 1 to your 
> rmt.yml and nowhere else, then default_rmt.yml will validate just fine, and then foofoo will get merged 
> into Settings -- and Settings.foofoo will work just fine in the code. 

But not in the tests on Travis which run with only with the `default_rmt.yml` I think. We could also hardcode the test env to only use `default_rmt.yml`.

ngetahun:
This is not useful to have anymore.

Issue #334, opened 2019-02-20
`rmt-cli mirror` doesn't set exit code on errors, unit doesn't enter failed state


ikapelyukhin:
Currently `rmt-cli mirror` continues to mirror repositories even if errors are encountered. However, it doesn't set exit code if errors were encountered, as the result `systemd` unit doesn't enter failed state. It is hard to tell if mirroring was successful or not, there are a lot of info messages displayed and it's hard to find error messages.

SMT displayed error summary at the very end with the list of repositories that weren't fully mirrored.

```
● rmt-server-mirror.service - Mirror RMT
   Loaded: loaded (/usr/lib/systemd/system/rmt-server-mirror.service; disabled; vendor preset: disabled)
   Active: inactive (dead) since Tue 2019-02-12 18:46:38 CET; 3s ago
  Process: 454 ExecStart=/usr/share/rmt/bin/rmt-cli mirror (code=exited, status=0/SUCCESS)
 Main PID: 454 (code=exited, status=0/SUCCESS)

Feb 12 18:46:37 e127 systemd[1]: Started Mirror RMT.
Feb 12 18:46:38 e127 rmt-cli[454]: INFO: RMT version 1.1.1
Feb 12 18:46:38 e127 rmt-cli[454]: INFO: Mirroring repository SLE-Module-Toolchain12-Pool to /usr/share/rmt/public/repo/SUSE/Products/SLE-Module-Toolchain/12/x86_64/product/
Feb 12 18:46:38 e127 rmt-cli[454]: WARN: Error while mirroring metadata: repodata/repomd.xml - return code ssl_cacert
Feb 12 18:46:38 e127 rmt-cli[454]: INFO: Mirroring repository SLE-Module-Toolchain12-Updates to /usr/share/rmt/public/repo/SUSE/Updates/SLE-Module-Toolchain/12/x86_64/update/
Feb 12 18:46:38 e127 rmt-cli[454]: WARN: Error while mirroring metadata: repodata/repomd.xml - return code ssl_cacert
```

felixsch:
@ikapelyukhin Do you know if something has changed in this regard? How is this handled now?

ikapelyukhin:
@felixsch no changes, handled in the same manner.

PR #344, opened 2019-03-11
Suma product tree

tmuntaner:


lib/rmt/mirror.rb
```diff
@@ -20,6 +20,26 @@ def initialize(mirroring_base_dir: RMT::DEFAULT_MIRROR_DIR, logger:, mirror_src:
...
+    base_dir = File.expand_path(File.join(@mirroring_base_dir, '/../')) if @mirroring_base_dir == RMT::DEFAULT_MIRROR_DIR
+
+    @repository_dir = File.join(base_dir, '/suma/')
+    @downloader.repository_url = URI.join(repository_url)
+    @downloader.destination_dir = @repository_dir
+
+    @logger.info _('Mirroring SUMA product tree to %{dir}') % { dir: @repository_dir }
+    @downloader.download('product_tree.json')
+  rescue RMT::Downloader::Exception => e
+    raise RMT::Mirror::Exception.new(_('Could not mirror suma product tree with error: %{error}') % { error: e.message })
```
felixsch: Can we not just log the error here and remove all try/catch statements in the other files?

package/nginx-http.conf
```diff
@@ -20,5 +20,9 @@ server {
         return 301 https://$host$request_uri;
     }
 
+    location /suma {
```
felixsch: Is the data required in plain `http`? I thought SM is also consuming other endpoints which are all only available via `https`?

public/suma/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
felixsch: I like this! Instead of have a `.keep` file around add a `.gitignore` :+1:

package/nginx-http.conf
tmuntaner: It is a request that RMT works more with http and that https is just the better workflow.

See: https://trello.com/c/fGuUG7O1/3105-rmt-allow-registration-over-http

felixsch: Okay didn't know about this problem. Than I'm fine with this (not with with autoyast being unable to import certs...)

lib/rmt/mirror.rb
tmuntaner: I guess I feel not having the function a black box and that we let the cli itself handle the failure state.

Also, it should make solving issues like [this](https://github.com/SUSE/rmt/issues/334) easier. I think that we actually want to implement a summary of failures these cli commands.

felixsch:
While building the package I noticed that a version bump is missing

tmuntaner:
Shouldn't we only bump the version when we do a submit request?

Issue #320, opened 2019-02-01
Provide checksum of `rmt-client-setup`


tmuntaner:
Clients download the `rmt-client-setup` script over http. We should provide a checksum file that they can verify the downloaded file and avoid a man in the middle.

**Ideas**:
* Build this with `Make`
* @thutterer This checksum could be on the static RMT page.

ikapelyukhin:
Include the code to accept certificates in SUSEConnect (there's already code for handling SSL-related exceptions there) and make `rmt-client-setup` obsolete! :grimacing:

tmuntaner:
@ikapelyukhin you're right, but we'll probably have to support `rmt-client-setup` on older systems, as they would have no way to update `SUSEConnect`.

ikapelyukhin:
@tmuntaner well, sure. But the sooner it's done in SUSEConnect, the sooner the need for supporting old systems will disappear :-)

tmuntaner:
@lagartoflojo Do you know if we're considering this feature for SUSEConnect in the future?

lagartoflojo:
> @lagartoflojo Do you know if we're considering this feature for SUSEConnect in the future?

if that's a desired feature, please create a card for it.

PR #345, opened 2019-03-11
Add checksums for rmt-client-setup

tmuntaner:
Currently, we don't have a way to verify that the downloaded
rmt-client-setup was safely downloaded.

If we provide the checksums, we can give the user a way to verify
downloaded files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ dist: clean man
 	@mkdir $(NAME)-$(VERSION)/vendor
 	@mkdir -p $(NAME)-$(VERSION)/public/repo/
 	@cp -r public/tools $(NAME)-$(VERSION)/public/
+	cd public/tools && md5sum * > ../../$(NAME)-$(VERSION)/public/tools/MD5SUMS
+	cd public/tools && sha1sum * > ../../$(NAME)-$(VERSION)/public/tools/SHA1SUMS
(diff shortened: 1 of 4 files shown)


tmuntaner:
This should solve  #320

suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ikapelyukhin:
But how does that make things more secure? In order to get onto HTTPS, the client instance needs to download `rmt-client-setup` over plain HTTP.  Let's assume that `rmt-client-setup` was modified en-route. So to check this you are supposed to download checksums... also over plain HTTP? What would prevent the potential attacker from modifying both the script and the checksums?

Also, even if the user is running the non-modified version of `rmt-client-setup`, the user still has to know CA certificate fingerprint, which they have to transmit over some other secure channel -- e.g., the sysadmin has look it up on RMT and remember it :-). They user might calculate the checksum for rmt-client-setup themselves -- there's no use in publishing it via HTTP.

If the goal is to save the user from hassle of verifying `rmt-client-setup`, why not just package it with SUSEConnect?

PR #352, opened 2019-03-13
Supply log severity to journald

### LINKED ISSUE #336: Supply message severity level to journald
There's already a special format for binaries that log to `journald` (enabled with `LOG_TO_JOURNALD` env variable). We should adjust the format to specify message severity level: the message needs to be prepended with numeric severity level in diamond brackets (`<severity_number>message`). This would enable filtering the logs (`journalctl -p`).

The following levels are supported by `journald`:

```
<7>This is a DEBUG level message
<6>This is an INFO level message
<5>This is a NOTICE level message
<4>This is a WARNING level message
<3>This is an ERR level message
<2>This is a CRIT level message
<1>This is an ALERT level message
<0>This is an EMERG level message
```

ikapelyukhin:
Fixes #336. Got too tired of seeing all the debug stuff in my logs :-)

### DIFF
--- a/lib/rmt/logger.rb
+++ b/lib/rmt/logger.rb
@@ -1,10 +1,34 @@
 class RMT::Logger < ActiveSupport::Logger
+
+  # For syslog log levels, see: man syslog(2)
+  # For Ruby log levels, see: https://github.com/ruby/ruby/blob/trunk/lib/logger.rb
+  RUBY_TO_SYSLOG_MAPPING = {
+    ::Logger::DEBUG   => 7,
+    ::Logger::INFO    => 6,
+    ::Logger::WARN    => 4,
+    ::Logger::ERROR   => 3,
+    ::Logger::FATAL   => 2,
+    ::Logger::UNKNOWN => 6
+  }.freeze
+
+  LABEL_TO_SEVERITY_MAPPING = {
+    'DEBUG'   => ::Logger::DEBUG,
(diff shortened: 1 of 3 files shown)


lib/rmt/logger.rb
```diff
@@ -1,10 +1,34 @@
...
+  }.freeze
+
   def initialize(dest)
     super(dest)
 
     # LOG_TO_JOURNALD is set by the systemd timers in order to remove duplicate timestamps in the logs
     if ENV['LOG_TO_JOURNALD'].present?
-      self.formatter = proc { |severity, _time, _progname, msg| "#{severity}: #{msg}\n" }
+      self.formatter = proc do |severity, _time, _progname, msg|
+        syslog_severity = RUBY_TO_SYSLOG_MAPPING[LABEL_TO_SEVERITY_MAPPING[severity]]
```
tmuntaner: Maybe we could better handle the nil case here?

```ruby
syslog_severity = RUBY_TO_SYSLOG_MAPPING[LABEL_TO_SEVERITY_MAPPING[severity]] || RUBY_TO_SYSLOG_MAPPING[::Logger::UNKNOWN]
```

What do you think?

ikapelyukhin: There shouldn't be a `nil` severity: https://github.com/ruby/ruby/blob/trunk/lib/logger.rb#L455-L457

PR #381, opened 2019-03-29
Rake test should fail on error

tmuntaner:
Currently, if rspec fails, we continue on and give an exit code of 0, that shouldn't happen.

I split the code coverage into their respective code suites: core and engine. Core will fail if we're missing 100% code coverage in normal RMT and engine will fail if the engine files are not 100% tested.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -14,6 +14,7 @@ Bundler.require(*Rails.groups)
 
 
 # Engine loading mechanism
+# :nocov:
 if (Rails.env.production? || ENV['RMT_LOAD_ENGINES'])
   Dir.glob("#{__dir__}/../engines/*").select { |i| File.directory?(i) }.each do |dir|
     engine_name = File.basename(dir)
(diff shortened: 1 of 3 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

config/application.rb
```diff
@@ -14,13 +14,15 @@
 
 
 # Engine loading mechanism
+# :nocov:
```
tmuntaner: For the reviewer:

`application.rb` will fail with code coverage with the core suite unless we ignore these lines. We can only check coverage with the engines suite.

ikapelyukhin: Oh yeah, that's probably what `fail_on_error = false` was for.

tmuntaner:
@ikapelyukhin thanks

PR #392, opened 2019-04-08
Allow registration on RMT over http

tmuntaner:


package/files/nginx/nginx-http.conf
```diff
@@ -17,12 +18,28 @@ server {
     }
 
     location = /repo/repoindex.xml {
-        return 301 https://$host$request_uri;
```
ikapelyukhin: In the cloud that would have upgraded the existing instances nicely from HTTP to HTTPS, and now they'd be stuck forever on HTTP. I guess now we need a separate config for the cloud :thinking:

tmuntaner:
@thutterer thanks

thutterer:
> _cough-cough_ Secure defaults, please?

Also, would that even fly? :chicken: 

I mean, we would ask the user to manually edit a file that comes from the package ... wouldn't that break with conflicts on an updated package version that comes with changes to that file?

ikapelyukhin:
@thutterer sure, it'd fly. The config files are marked by the `%config` directive in the RPM spec file -- this way when there is an update to the config file, the new config will be saved as `.rpmnew`.

Reducing copypasta and using includes would help with this, BTW (because the stuff that's being included would change, and the config where you do the include wouldn't).

thutterer:
> sure, it'd fly. The config files are marked by the `%config` directive in the RPM spec file -- this way when there is an update to the config file, the new config will be saved as `.rpmnew`.

And then we except the user to notice that and manually move that file to the real location, resolving conflicts manually??

> Reducing copypasta and using includes would help with this, BTW (because the stuff that's being included would change, and the config where you do the include wouldn't).

So you can see into the future and tell me that we would never ever make changes to `nginx-http.conf` again? :crystal_ball:

ikapelyukhin:
> And then we except the user to notice that and manually move that file to the real location, resolving conflicts manually??

Pretty much. But that's how every single RPM package that has a configuration works. That would affect only the subset of users who choose to do this. IMO, still better than unsecure defaults.

> So you can see into the future and tell me that we would never ever make changes to `nginx-http.conf` again? crystal_ball

If includes are done correctly -- most likely this will be the case. If there are no includes -- there's copypasta in http/https conf and both have to be updated simultaneously every time a change is needed. Also, with insecure defaults I'll package secure cloud configs into `rmt-server-publcoud`, so there will be even more copypasta. :-)

tmuntaner:
@ikapelyukhin The team decided to allow HTTP by default.

* The documentation for RMT shows HTTPS, so customers will use HTTPS by default. We can add more to make this a stronger recommendation.
* Users have to choose between HTTP or HTTPS when they configure a server, if they care about HTTPS, they will use HTTPS then.
* Customers expect RMT to work like SMT, which allows registration over HTTP.
* Customers trust their networks and don't want to worry about SSL certificates.
* CaaSP requires RMT to allow registration over HTTP.
* We added comments to help a user disable HTTP if they so wish.
* We only need to worry about one configuration on our reference RMT server.

I'm sorry if this causes any issues for you in the public cloud. Your use case is slightly different and requires something more hardened.

You can also see the comment from Hernan at https://trello.com/c/fGuUG7O1.

ikapelyukhin:
@tmuntaner OK, I'll have to package the hardened configs into the subpackage later, then.

PR #396, opened 2019-04-09
Added the ability to configure registration over http

tmuntaner:


tmuntaner:
Yeah, with if, we have to tread very carefully (https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/).

Tomorrow, I'll see if I can make improvements and present it to the team as an alternative.

It's mostly a requirement from customers that we enable http by default. They trust their environments and don't want the complications of https. Also, CaaSP has a http only requirement for RMT.

I can only hope that, if customers follow the documentation, they'll use RMT with https.

PR #409, opened 2019-05-08
Make service IDs consistent across different RMT instances

ikapelyukhin:
Same product might have different service URLs on different RMT instances, as `service.id` field is auto-increment. This breaks HA setups.

Service ID isn't exposed in sync API, so I've made `service.id = product.id` for instances that haven't synced yet (existing instances will keep random IDs).


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          3 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><table>
  <tbody>
    <tr>
      <td>app/models/product.rb</td>
      <td>133</td>
      <td>Style/ConditionalAssignment: Use the return of the conditional for variable assignment and comparison.</td>
    </tr>
  </tbody>
</table>
</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><table>
  <tbody>
    <tr>
      <td>app/models/product.rb</td>
      <td>133</td>
      <td>Style/UnlessElse: Do not use <code>unless</code> with <code>else</code>. Rewrite these with the positive case first. (https://github.com/bbatsov/ruby-style-guide#no-else-with-unless)</td>
    </tr>
  </tbody>
</table>
</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><table>
  <tbody>
    <tr>
      <td>app/models/product.rb</td>
      <td>133</td>
      <td>Style/RedundantParentheses: Don’t use parentheses around a method call.</td>
    </tr>
  </tbody>
</table>
</td>
    </tr>
  </tbody>
</table>

</p>

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/2020/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged fd38d1f732138475cd23ba8273d1fe05f11c1d1b into 92c43f74076447f9eb4ae2a091b05061e013cfe0)

app/models/product.rb
```diff
@@ -127,7 +127,17 @@ def self.recommended_extensions(root_product_ids)
   end
 
   def service
-    Service.find_or_create_by(product_id: id)
+    service = Service.find_by(product_id: id)
```
digitaltom: Is this line needed? Because we have `has_one :service`.

digitaltom: Oh, I see :eyes:

ikapelyukhin: It used to be `find_or_create_by` -- this line handles the `find` part, rest of the method handles creation of the new service. This is `def service`, so writing `self.service` here would result in recursion -- so I think it is needed.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/2626/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged fd38d1f732138475cd23ba8273d1fe05f11c1d1b into 3429cdbbac99702421b3bc3bf1b95be8cb18d538)

suse-tests-fail:
Prophet reports failure.
https://ci.suse.de/job/scc-RMT-integration-tests/2627/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment.
Failed fast in project rmt on step docker-compose-run-tests. (Merged 778413be2b5ed0f87b167f4b11b4d10a1b1968c5 into 3429cdbbac99702421b3bc3bf1b95be8cb18d538)

app/models/product.rb
```diff
@@ -127,7 +127,17 @@ def self.recommended_extensions(root_product_ids)
...
+
+    service ||= if Service.find_by(id: id)
+                  # for backward-compatibility: create a service with autoincrement ID, if product.id is already occupied
+                  Service.create!(product_id: id)
+                else
+                  # try to create a service with service.id = product.id for keeping consistent service URLs across RMT instances
+                  Service.create!(id: id, product_id: id)
```
digitaltom: There is the chance of a race condition here. Could it happen that on a system this method gets called in parallel?

app/models/product.rb
digitaltom: I don't like that `service()` implicitely creates the service if it is not there. We should know at which point we need to create the service and when we can expect it to be there.

ikapelyukhin: @digitaltom I don't like `Service` model at all. :-) It doesn't do anything and isn't exposed by the SCC API -- could just use products instead.

There was duplication of service creation code in `RepositoryService` service and `Product` model. Should I move service creation code to `RepositoryService`? Not sure if that will that break, though.

app/models/product.rb
ikapelyukhin: Indeed, there is a chance of that. I'll move this into `RepositoryService` so that it's called from `rmt-cli sync` only and wrap the whole product creation method into a transaction.

ikapelyukhin:
@digitaltom @Sergeykot seems like Prophet is flickering. It was green half an hour ago.

app/models/product.rb
ikapelyukhin: @digitaltom OK, I've basically renamed `service()` into `create_service!`, so it's not automagically created on trying to access service anymore but rather explicitly called during SCC sync.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/2648/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged ba51bc5d87e433c0a449ab2b8f46c9e6075f8618 into 3429cdbbac99702421b3bc3bf1b95be8cb18d538)

digitaltom:
Ok, go ahead :ship:

ikapelyukhin:
:shipit:

PR #411, opened 2019-05-13
Mirror repositories that were enabled during mirroring

ikapelyukhin:


suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/3463/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged bf8772b95c7fd1f7649aba93965c8f55d48e22b1 into 80598413be938f78d8a2b856f0c91f21071ea867)

wstephenson:
I think that if we are going to change the mirroring logic to include all repos that are enabled during the mirror run, for consistency, we should also not mirror any repos which are disabled during the mirror run. 
The current logic gets all the unmirrored enabled repos and iterates over those, mirroring them. If any of these are disabled during this loop, they will be mirrored regardless. Use case: admin activates a bunch of architectures that they do not want and notices this during the mirroring run, and disables all but x86_64.

Could the loop be changed to check for the first unmirrored enabled Repository each time around and mirror that one only, thereby reducing the 'redundant mirroring window' to a single repo?

wstephenson:
Please also document the change in MANUAL.md

ikapelyukhin:
@wstephenson disabled repos aren't a concern, at least not for our use cases. If the admin has enabled extra repos -- he can kill the mirroring process. This change is intended for automated management of server configurations -- changes to configuration can be deployed while mirroring is running; mirroring extra repos isn't a concern here.

I'd not implement handling of disabled repos unless there's actually a demand for it.

wstephenson:
I understand your use case, but want to keep RMT consistent

Isn't this a cheap change?
```
do
 repo = Repository.where(<as already given>).first
  ...
until Repository.where(<as already given).blank?
```
and you handle both disabled and enabled repos during mirroring, consistent semantics in both directions.

ikapelyukhin:
@wstephenson OK, fair enough. Done.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/4246/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 30210e01013b70e557ed397c4b82c2abb32b7cea into 80598413be938f78d8a2b856f0c91f21071ea867)

PR #458, opened 2019-08-19
Display a product with its repositories

### LINKED ISSUE #441: Display a product with its repositories
It would be useful to be able to display a product with all its repositories and 
their attributes (mirroring enabled, mandatory, id, last mirrored etc) over the cli. 

For example: `rmt-cli products show <product>`

Recommended output: 

```
> rmt-cli products show SLES/15/x86_64

Product: SUSE Linux Enterprise Server 15 x86_64 (id: 1575)
Description: SUSE Linux Enterprise offers a comprehensive suite of products built on a single code base. The platform addresses ...
Repositories: 
  * SLE-Product-SLES15-SP1-Pool (id: 3909) (mandatory, enabled, mirrored at 2019-08-02 09:25:10 UTC)
  * SLE-Product-SLES15-SP1-Updates (id: 3908) (mandatory, enabled, mirrored at 2019-08-02 09:25:10 UTC)
  * SLE-Product-SLES15-SP1-Source-Pool (id: 3907) (not mandatory, not enabled, not mirrored)
  * SLE-Product-SLES15-SP1-Debuginfo-Pool (id: 3906) (not mandatory, not enabled, not mirrored)
```

``` 
> rmt-cli products show help

Usage:
  rmt-cli products show PRODUCT

Description:
  Shows details of a product and its repositories.

  Examples:

  $ rmt-cli products show SLES/15/x86_64

  $ rmt-cli products show 1575
```

arharovets:
Fixes #441 .
Also fixes an unexpected behavior of `after_create` callback at `/spec/factories/services.rb`.
Additionally, more readable `long_desc` at `/lib/rmt/cli/products.rb` using [squiggly heredoc](https://github.com/ruby/ruby/pull/878).


lib/rmt/cli/products.rb
```diff
@@ -36,41 +36,80 @@ def list
...
+    $ rmt-cli products enable --all-modules SLES/15
+  REPOS
   def enable(*targets)
     change_products(targets, true, options[:all_modules])
   end
 
   desc 'disable TARGETS', _('Disable mirroring of product repositories by a list of product IDs or product strings.')
-  long_desc <<-REPOS
-#{_('Disable mirroring of product repositories by a list of product IDs or product strings.')}
+  long_desc <<~REPOS
```
digitaltom: nice :+1: 
is this supported in Ruby 2.5.0+?

spec/factories/services.rb
```diff
@@ -17,15 +17,13 @@
             :repository,
             name: "#{name_prefix}-Installer-Updates",
             mirroring_enabled: evaluator.mirroring_enabled,
-            installer_updates: true,
-            enabled: false
+            installer_updates: true
```
digitaltom: Why is this change needed? Background: `installer_updates` repos are usually set to disabled.

spec/lib/rmt/cli/products_spec.rb
```diff
@@ -468,4 +468,130 @@
...
+          end
+          output
+        end
+
+        it 'shows details of a product and its repositories' do
+          expect { described_class.start(argv) }
+            .to output(expected_output).to_stdout
+            .and output('').to_stderr
+        end
+      end
```
digitaltom: It feels like there is a lot of repetition in the last 3 tests. Is there a way to do this more concise?

lib/rmt/cli/products.rb
arharovets: @digitaltom yes, this is supported since 2.3 [prooflink](https://ruby-doc.org/core-2.5.0/doc/syntax/literals_rdoc.html#label-Here+Documents)

spec/factories/services.rb
arharovets: @digitaltom Every time this factory called, it creates 4 repos, but two of these repos are always `enabled: false`.
```ruby
let(:product) { create :product , : with_mirrored_repositories }
$ product.repositories.pluck(:name, :scc_id, :enabled, :mirroring_enabled, :last_mirrrored_at)
=> [["Product 80-Pool", 271, true, true, nil],
 ["Product 80-Updates", 272, true, true, nil],
 ["Product 80-Installer-Updates", 273, false, true, nil],
 ["Product 80-Debuginfo-Updates", 274, false, true, nil]]
```
Using this factory along with `products` factory, we can create four different variations of the product with repositories (`:with_mirrored_repositories`, `:with_disabled_mirrored_repositories`, `:with_disabled_not_mirrored_repositories`, `:with_not_mirrored_repositories`). So every time we creating a product with mirrored repositories (enabled and mirrored), there's always two disabled repos. Even though usually `installer_updates` are set to disabled, I suggest this change just to cover more possible cases as it gives us the flexibility of creating more product variants.

spec/lib/rmt/cli/products_spec.rb
arharovets: @digitaltom Sure, I'm on it.

arharovets:
> Why do you suggest to raise an exception here? I think there are a few products without repos that are valid, like products that are only delivered via container images.

@digitaltom Thank you for pointing that out! I'll fix that.

spec/factories/services.rb
ikapelyukhin: > it creates 4 repos, but two of these repos are always `enabled: false`.

That replicates a normal product setup -- two mandatory repos (Pool, Updates) and two non-mandatory repos (DebugInfo, Source).

(`repository.enabled` means enabled on the client side by default, as in `zypper mr -e`)

lib/rmt/cli/products.rb
```diff
@@ -36,41 +36,80 @@ def list
...
+  def show_product_repos(product)
+    repos = product.repositories
+      .pluck(:name, :scc_id, :enabled, :mirroring_enabled, :last_mirrored_at)
+    raise RMT::CLI::Error.new(_('Repositories are not available for this product.')) if repos.blank?
+    puts _('Repositories:')
+    repos.each do |repo|
+      repo[2] = repo[2] ? 'mandatory' : 'non-mandatory'
+      repo[3] = repo[3] ? 'enabled' : 'not enabled'
+      repo[4] = repo[4].present? ? "mirrored at #{repo[4].strftime('%Y-%m-%d %H:%M:%S %Z')}" : 'not mirrored'
+      puts _("* #{repo[0]} (id: #{repo[1]}) (#{repo[2]}, #{repo[3]}, #{repo[4]})")
```
ikapelyukhin: This won't work correctly with i18n -- an interpolated string is being passed to to `_()` in runtime (.po files won't be generated correctly).

spec/factories/services.rb
arharovets: Hi @ikapelyukhin ,

Okay, but may I suggest that the goal is to test every possible variation, not to replicate the default product setup? Also, removing this value doesn't seem to break any tests.

lib/rmt/cli/products.rb
arharovets: @ikapelyukhin Thanks! Would you recommend to use a template?

ikapelyukhin: Yes, like it's done here for example: https://github.com/SUSE/rmt/blob/f48968e6cca4bba22b76806d83b3a6d7e7ca3f5c/lib/rmt/cli/products.rb#L93-L94 

Use `_()` on regular strings, e.g.: 
```
repo[2] ? _('mandatory') : _('non-mandatory')
```
And then use internationalized strings in the template.

arharovets: @ikapelyukhin Great, thank you!

lib/rmt/cli/products.rb
ikapelyukhin: :+1: for squiggly heredoc. I was recently wondering if Ruby had something like Python's `dedent()` :slightly_smiling_face:

spec/factories/services.rb
ikapelyukhin: Real product setup is a reasonable default, it doesn't interfere with testing all possible combinations in any way.
In principle, I don't object to changing this; however I can't guarantee that there are no side effects with this change either, so I personally wouldn't change it.
But the decision of whether to change it or not is really up to @digitaltom and other RMT maintainers :grimacing:

lib/rmt/cli/products.rb
```diff
@@ -36,41 +36,85 @@ def list
...
+    repos = product.repositories
+      .pluck(:name, :scc_id, :enabled, :mirroring_enabled, :last_mirrored_at)
+    puts repos.present? ? _('Repositories:') : _('Repositories are not available for this product.')
+    repos.each do |repo|
+      repo[2] = repo[2] ? _('mandatory') : _('non-mandatory')
+      repo[3] = repo[3] ? _('enabled') : _('not enabled')
+      repo[4] = repo[4].present? ? _('mirrored at %{time}') % { time: repo[4].strftime('%Y-%m-%d %H:%M:%S %Z') } : _('not mirrored')
+      template = '* %{name} (id: %{id}) (%{mandatory}, %{enabled}, %{mirrored})'
+      data = { name: repo[0], id: repo[1], mandatory: repo[2], enabled: repo[3], mirrored: repo[4] }
+      puts _(template) % data
```
ikapelyukhin: Almost there:
```
template = _('...') # only string literals are allowed in _()
puts template % data
```
:slightly_smiling_face:

spec/factories/services.rb
digitaltom: I would keep it as it was before (with `enabled: false` on DebugInfo and Installer repos) to keep the default factory  data close to real-world. 
The four testcases (`:with_mirrored_repositories, :with_disabled_mirrored_repositories, :with_disabled_not_mirrored_repositories, :with_not_mirrored_repositories`) might even get merged into just 2, because there is always a mix of enabled + disabled repos in the output. I don't see the benefit in extra testing of a product with *only* enabled/disabled repos.

arharovets: @digitaltom @ikapelyukhin Alright, I'll leave it as it was before.

lib/rmt/cli/products.rb
arharovets: @ikapelyukhin Oh, thanks :slightly_smiling_face:

MANUAL.md
```diff
@@ -58,6 +58,15 @@ You can install and run this wizard like this:
     `rmt-cli products disable 1575`
 
     `rmt-cli products disable SLES/15/x86_64 1743`
+  
+  * `rmt-cli products show <id | string>`:
+  	Displays product with all its repositories and their attributes.
+  	
+  	Examples:
+  	
+  	`rmt-cli products show SLES/15`
```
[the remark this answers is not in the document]
digitaltom: @arharovets ^^

spec/lib/rmt/cli/products_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'rails_helper'
-
+require 'pry'
```
digitaltom: looks like a leftover from debugging?

arharovets: my bad :slightly_smiling_face:

PR #463, opened 2019-08-30
Create rmt-client-setup-res

zpetrova:
rmt-client-setup script asssumes SUSEConnect and zypper are already installed on the system. This is not true for RHEL8, so this script installs there packages and then downloads and runs rmt-client-setup script.

PR #466, opened 2019-09-05
Create rmt-client-setup-res

zpetrova:
rmt-client-setup script asssumes SUSEConnect and zypper are already installed on the system. This is not true for RHEL8, so this script installs there packages and then downloads and runs rmt-client-setup script.


josegomezr:
Closing in favor of #463 as this being unmergeable.

Issue #493, opened 2019-12-02
Make it possible to set low_speed_limit and low_speed_time


doccaz:
In some situations, there are network content checking appliances/proxies that will hold the connection open until the entire file is fetched, and *then* the client (in this case, RMT) receives the entirety of the data stream. Since the connection is mostly receiving just a few bytes/sec until the proxy/appliance is done, RMT thinks the connection is stuck, drops it and marks the file as not downloaded. This also makes downloading any large file impossible.

Analyzing the code, I noticed that RMT is using fixed values in http_request.rb. I'm proposing a change to read these two options from rmt.conf. If they're not present, just assume the current hard-coded values.

felixsch:
This is addressed here: https://github.com/SUSE/rmt/pull/494

The pull request has been merge, thats why I'm closing this.

PR #494, opened 2019-12-02
making low_speed_limit and low_speed_time configurable

doccaz:
Implementing the changes described on issue #493 .


ikapelyukhin:
Making those configurable makes sense to me. I also think it would make sense to also add those options to the default config (`/package/obs/rmt.conf`), just so that it is obvious that those options exist.

(We'd also need to document those in RMT handbook, but that's an action item for when the PR is merged)

lib/rmt/http_request.rb
```diff
@@ -17,8 +17,8 @@ def set_defaults
     end
 
     # Abort download if speed is below 512 bytes/sec for 120 sec, to prevent downloads from getting stuck
-    options[:low_speed_limit] = 512
-    options[:low_speed_time] = 120
+    options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) ? Settings.try(:http_client).try(:low_speed_limit) : 512
```
thutterer:
```suggestion
    options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) || 512
```

lib/rmt/http_request.rb
```diff
@@ -17,8 +17,8 @@ def set_defaults
...
+    options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) ? Settings.try(:http_client).try(:low_speed_time) : 120
```
thutterer:
```suggestion
    options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) || 120
```

lib/rmt/http_request.rb
```diff
@@ -17,8 +17,8 @@ def set_defaults
     end
 
     # Abort download if speed is below 512 bytes/sec for 120 sec, to prevent downloads from getting stuck
```
thutterer: Maybe update this comment since the values aren't hard-coded anymore.

doccaz:
Sorry for the delay.  Here are all the changes you requested.

felixsch:
Hi doccaz,

Thank you for you contribution! The changes look good to me. The only thing would it be possible for you to update the tests in `spec/lib/rmt/http_request_spec.rb`. They assume a hardcoded value which is now configurable.

In case there are questions how to run the tests or anything else let me know.

felixsch:
Hi doccaz,

to not longer delay the pull request. I decided to merge your and write the test myself. Thank you for your contribution again!

doccaz:
I'm so sorry! I delved into a critical problem at a customer for the whole
week and completely forgot about this. Thank you so much!

Issue #233, opened 2018-10-15
Use `repomd_parser` Gem


ikapelyukhin:
I have split off the code that parses repository metadata into a separate gem -- `repomd_parser` [[1](https://rubygems.org/gems/repomd_parser)] [[2](https://github.com/ikapelyukhin/repomd-parser)], mostly to make it possible to use in [repo-tools](https://github.com/ikapelyukhin/repo-tools).

I've done some refactoring in the gem and added access to some more metadata attributes. It would be cool to move RMT over to `repomd_parser` as well, so that it doesn't have its own implementation of those things.

Vote/comment on this issue if you think it's a good idea to do this :grimacing:

wstephenson:
Is there no existing ruby gem to do this, eg from yast?

ikapelyukhin:
@wstephenson well, I haven't found anything -- you are welcome to give another shot at searching, though.

tmuntaner:
@ikapelyukhin did you want to make the PR for this or should someone in SCC do that?

ikapelyukhin:
@tmuntaner Whoever gets to it first!

tmuntaner:
@ikapelyukhin ok, I guess it will be whoever is more annoyed about the still open issue ^_^

digitaltom:
Probably we could also make use of this in our 'duct-tape' code that parses repos for Patchfinder and Packagefinder.

digitaltom:
@ikapelyukhin are the initializers from `repomd_parser` missing the `mirror_src` parameter?

ikapelyukhin:
> @ikapelyukhin are the initializers from `repomd_parser` missing the `mirror_src` parameter?

I've dropped `mirror_src` in the gem (for the sake of versatility of the gem, `mirror_src` makes sense in the context of RMT but not necessarily in a context of the gem that parses XML).

Source packages have to be filtered out outside of calls to the gem.

digitaltom:
Ok. That makes the task a bit more complicated, I was searching for a beginner task of 2 hours.

ikapelyukhin:
> Ok. That makes the task a bit more complicated, I was searching for a beginner task of 2 hours.

Well, there's the risk, of course, but since it's mostly the same code inherited from the same codebase, I would expect this code:

```ruby
parser = RMT::Rpm::PrimaryXmlParser.new(
  File.join(@temp_metadata_dir, filename),
  @mirror_src
)
parser.parse
[...]

PR #520, opened 2020-01-30
Replace project XML parsers with `repomd_parser` gem

lcaparroz:
## Why

As mentioned in issue #233, the code for parsing XML repository files in the `rmt` project was split to a new gem, `repomd_parser`, for reusing in other projects. However, the original code for XML parsing is still in use.

### Replacing the `RMT::Rpm::RepomdXmlParser`

The first step was to replace the `RMT::Rpm::RepomdXmlParser` because its equivalent parser on the gem (`RepomdParser::RepomdXmlParser`) matches all of its functionality.

### Replacing `RMT::Rpm::PrimaryXmlParser` and  `RMT::Rpm::DeltainfoXmlParser`

When I first tried to replace the `RMT::Rpm::PrimaryXmlParser`, no tests broke. I noticed the spec for `RMT::Mirror` didn't had test scenarios for when the package architecture is `'src'` and the `mirror_src` option is `false`. This verification step was implemented directly on the `RMT::Rpm` parsers but not in the `repomd_parser` gem, since this is a business rule that only makes sense in the context of the RMT project (as explained by the gem's creator).

As both the _primary_ and _deltainfo_ `RMT::Rpm` parsers had this business rule, before replacing them, I created some fixture files for mocking RPM packages with `arch == 'src'` and I updated the `.xml.gz` repo files with the required information for these new packages. With the fixtures, I did some manual testing to assure that removing the business rule would break the tests. Only then I proceeded to replacing the other XML parsers with the gem.

I implemented the `arch == 'src' && mirror_src == false` business rule directly into the `RMT::Mirror` class, since this rule defines behavior when mirroring repository files.

### Removing `RMT::Rpm` files

As I replaced the project XML parsers by the gem ones, I also removed their files to clean up the code base. In the end, I was able to completly remove all the `RMT::Rpm` classes/files, assuring every test still passed.

### Adding a scenario to test the `arch == 'src' && mirror_src == false` rule

After assuring the project behavior was not modified when replacing the XML parsers, I recorded one of the VCR cassette fixtures to allow requests to download packages with `arch == 'src'` and I implemented a new test scenario for the `RMT::Mirror#mirror` method with `mirror_src == true`, to assure the user is able to mirror `src` packages.

### Doing some clean code

I rewrote some of the methods which depend on parsing XML files to improve readability and to remove duplicated business logic, breaking methods into smaller ones.

### VCR cassettes

I tried to rewrite the `.xml.gz` files to include RPM packages with `arch == 'src'` but I noticed that it would take no effects on the tests. Only after some time I found out these files were being served by the VCR cassettes, which included their binary content.

My strategy to create fixtures for the `src` packages was to serve these files with a localhost server (using `python3 -m http.server 80`) and then record again the VCR fixtures.

However, I am not sure how far my setup for the localhost server is from a real situation. **Does recording the VCR cassettes by making requests to my localhost server affect negatively the tests? Does it make the tests more fragile?**

### RPM `src` packages

As I don't have a solid grasp on how to specify RPM packages, I created some dummy files with the `.rpm` extension and I got their `sha256` checksums to update the `.xml.gz` files. However, I was not sure how to fake all the available information in the XML files, I just replaced `sha256` checksums and file sizes.

**Is there any implication on using these fake files for the tests? Does it make the tests more fragile?**

Furthermore, I noticed the `.xml.gz` files checksums are used as part of their filenames (I guess this is the standard way for naming these files). As I changed these files, their checksum also changed. I just updated the expected checksums on `spec/fixtures/files/dummy_repo/repodata/repomd.xml`, but I kept the file names. **Would just updating checksums but not the filenames make any negative impact on the tests?**


lib/rmt/mirror.rb
```diff
@@ -100,18 +100,9 @@ def mirror_metadata
...
-
-    metadata_files = []
-    repomd_parser.referenced_files.each do |reference|
-      metadata_files << reference
-      primary_files << reference.location if (reference.type == :primary)
-      deltainfo_files << reference.location if (reference.type == :deltainfo)
-    end
+    metadata_files = RepomdParser::RepomdXmlParser.new(local_filename).parse
+    primary_files = metadata_files.select { |reference| reference.type == :primary }
+    deltainfo_files = metadata_files.select { |reference| reference.type == :deltainfo }
```
jmammen: Just a small addition: Would it make sense to abstract the selection of `primary_files` and `deltainfo_files` into a method? If we need to select the files in multiple locations and the criteria to select them change we only need to change the selection there.

lib/rmt/mirror.rb
```diff
@@ -143,32 +134,35 @@ def mirror_data(primary_files, deltainfo_files)
...
-        File.join(@temp_metadata_dir, filename),
-        @mirror_src
-      )
-      parser.parse
-      to_download = parsed_files_after_dedup(@repository_dir, parser.referenced_files)
-      failed_downloads.concat(@downloader.download_multi(to_download, ignore_errors: true)) unless to_download.empty?
-    end
+    package_files =
+      parse_mirror_data_files(deltainfo_files, RepomdParser::DeltainfoXmlParser) +
+      parse_mirror_data_files(primary_files, RepomdParser::PrimaryXmlParser)
```
jmammen: Would it also make sense to select the needed `package_files` within in method?

lib/rmt/mirror.rb
lcaparroz: HI, @jmammen. I thought of two ways for doing what you suggested:

1. Abstracting these lines into two different methods (e.g.: `select_primary_files` and `select_deltainfo`files` or;
1. Creating just one method which receive the file type as a parameter (e.g.: `select_metadata_files_by_type(:primary)`).

The second option is more flexible, but it would be prone to mistyping: should someone mistype the filetype symbol (e.g.: `:delta_info`), the method would fail silently (it'd return no files and it'd also not raise any exceptions). Also, if we needed to select the files in multiple locations, it would incur typing the file type symbol in as many locations as needed.

The first option would incur duplicated code, though it would leave room for improving/modifying the selections separately in the future, if needed.

I'm going with the first option, but I'd appreciate your thoughts.

spec/fixtures/files/dummy_repo/repodata/repomd.xml
```diff
@@ -2,35 +2,35 @@
...
   <timestamp>1500454273</timestamp>
-  <size>402</size>
+  <size>581</size>
   <open-size>929</open-size>
 </data>
 <data type="deltainfo">
-  <checksum type="sha256">a546b430098b8a3fb7d65493a9ce608fafcb32f451d0ce8bf85410191f347cc3</checksum>
+  <checksum type="sha256">8c9d9fbbe1b87009a145986a78ec1180dedc2455407f715d7a65174470d49793</checksum>
   <open-checksum type="sha256">e71433e2373d61a2e19f335c7f634b9bcfb1d458bc912626dc9ddc5f3e124981</open-checksum>
   <location href="repodata/a546b430098b8a3fb7d65493a9ce608fafcb32f451d0ce8bf85410191f347cc3-deltainfo.xml.gz"/>
```
ikapelyukhin: Hmm, seems like you've changed the `checksum` and `size`, but not the other fields: `location` still points to the file with the old name -- which shouldn't exist anymore, `open-checksum` and `open-size` probably should have changed as well.

So if this fixture still works -- I'm not sure why.

lcaparroz: Hi @ikapelyukhin, the fact that I updated metadata and checksums manually worried me from the very beginning. I just touched the files because there were no test scenarios for when the option `mirror_src == true` and when I removed the check for the flag on the parser classes, no tests broke (more details in d69b9c7), I needed some way to test assure it would behavior as expected.

I have little to no experience in building packages; I once wrote a RPM spec, a few years ago. I am trying to follow the [official OBS beginner's guide](https://openbuildservice.org/help/manuals/obs-beginners-guide/), but I am not acquainted with all the concepts, so I am struggling a bit to make a dummy repo myself.

Also, I have searched for a few other solutions; the command `createrepo` closely matches what I tried to do, but it does not generate the `deltainfo.xml` file.

Do you know a simpler way to build the metadata files or is it strictly necessary theses files matches all the requirements from a real repository?

lib/rmt/mirror.rb
lcaparroz: I followed your suggestions on 687711c.

lib/rmt/mirror.rb
lcaparroz: I followed your suggestions on 687711c.

spec/fixtures/files/dummy_repo/repodata/repomd.xml
ikapelyukhin: We can help with creating the repo. IIRC, I've built the RPMs with OBS and then created the repo with `createrepo`.

With regards to deltainfo:
```
createrepo -h | grep delta
  --deltas              create delta rpms and metadata
                        paths to look for older pkgs to delta against
```

There has to be more than one version of the package for deltas to be created, though.

lcaparroz: I tried running the `createrepo` command against the files in `spec/fixtures/files/dummy_repo`, after removing the `.drpm` files and the `repodata` directory. I changed directory to `dummy_repo` and executed `$ createrepo --deltas `.

The command generated all the `.xml.gz` files, but instead of a `deltainfo.xml.gz`, it generate a `prestodelta.xml.gz`, that was basically empty.

I took a look at the openSUSE documentation for [Standard RPM Metadata](https://en.opensuse.org/openSUSE:Standards_Rpm_Metadata#Current_Extensions) to see how to generate `deltainfo.xml` files. It seems it recomends an extension to the `createrepo` command called [`enhancerepo`](https://en.opensuse.org/openSUSE:Enhancerepo). I tried to install it from source and a [Ruby gem](https://rubygems.org/gems/enhancerepo) with the samename. However, I could not manage it to work correctly: the command tries to open a library file on `/usr/lib/rpm`, but on both openSUSE Leap 15.1 and Fedora 29 the path is for a directory.

ikapelyukhin: `prestodelta.xml.gz` -- that's Fedora flavor, the tags in the file are slightly different, too. Not sure why DRPMs aren't created, though. I don't have any recollection of how I created those DRPMs :thinking: 

In the meanwhile, I've created [home:ikapelyukhin:dummy_repo](https://build.opensuse.org/project/show/home:ikapelyukhin:dummy_repo). I'm not done fiddling with it yet, though, it takes a while for the repos to be published after the RPMs are built.

lcaparroz: Hi @ikapelyukhin, thank you very much for the time and support! Is it possible to also generate `.src.rpm` packages within OBS? The whole point of what I did (editing manually the fixtures files) was to have some way to test the condition `arch == `src` && mirror_src == true` when mirroring packages.

Currently, the project does not have fixture files for `.src.rpm` packages, so anything that

ikapelyukhin: @lcaparroz yes, you can see the source package was [built here](https://download.opensuse.org/repositories/home:/ikapelyukhin:/dummy_repo/openSUSE_Leap_15.1/src/).

It seems like publishing of build artifacts is rather slow on public OBS instance, I'm fiddling with project settings on our internal OBS instance to build a repo that has both source and deltarpm packages.

lcaparroz: @ikapelyukhin I managed to recreate the delta RPM packages with the `makedeltarpm` package (available in Fedora through the `deltarpm` package). It seems the command generate exactly the same file:

```sh
$ cd spec/fixtures/files/dummy_repo
$ makedeltarpm apples-0.1-0.x86_64.rpm apples-0.2-0.x86_64.rpm apples-0.2-0.x86_64.drpm

$ diff apples-0.1-0.x86_64.drpm oranges-0.1-0.x86_64.drpm; echo $?
Binary files apples-0.1-0.x86_64.drpm and oranges-0.1-0.x86_64.drpm differ
1

$ diff apples-0.1-0.x86_64.drpm apples-0.2-0.x86_64.drpm; echo $?
0
```

I did not find out a way to generate the `deltainfo.xml` file, though. Would OBS generate this automatically once there are two versions of a package? I read the [project configuration](https://build.opensuse.org/projects/home:ikapelyukhin:dummy_repo/prjconf) from your repository and it has this line `Repotype: rpm-md:deltainfo`, so I thought maybe OBS has the capability.

ikapelyukhin: @lcaparroz OBS definitely has the capacity, there's openSUSE/SLES drpms come from. :slightly_smiling_face: 
I've added a new version for `apples` package, hopefully drpm will get published in the project repo soon.

lcaparroz: Nice! About the VCR cassettes fixtures in the `rmt` project, for emulating requests to a repository, the tests run requests against `localhost`. I used `python3 -m http.server 80` to serve the files, but I don't know if there is a better way.

For what I understood, changing only the package/XML files makes no effects on the tests, since the VCR fixtures record the file contents as the request response body. I think it would be nice to have some documentation for how to record again the VCR cassettes and for generating package/XML files, since people new to the project may not know how to do this.

I can create the documentation for that, splitting it in two parts:

1. Generating RPM packages and XML metadata files;
1. Serving these files in `localhost` for recording the VCR cassettes.

What do you think about it? I could create a standalone markdown file and reference it in the README "Development Setup" section.

lcaparroz: Hi @ikapelyukhin, I replaced the files (RPM packages/XML metadata) I edited manually with the ones generated in your OBS repository, on commit 27ba069.

Just a few doubts:

* Does it make sense to have "src" delta RPMs packages (`*.src.drpm`)? I noticed that OBS just generated delta RPMs for the `x86_64` packages but not for `src` ones, and that the previous deltainfo XML parser (`RMT::Rpm::DeltainfoXmlParser`) had the rule for checking if a delta RPM package has `arch == 'src'` when `mirror_src == false`.
* The original fixture files included two versions for the same package (namely `0.0.1` and `0.0.2`), but after you committed a new version of the `apples` package to OBS, it is just serving the newer version and a delta RPM. Should we have both old and new versions for the fixtures?

ikapelyukhin: > About the VCR cassettes in the `rmt` project, for emulating requests to a repository, the tests run requests against `localhost`. I used `python3 -m http.server 80` to serve the files, but I don't know if there is a better way.

Any web server will do. I believe I've had a local nginx installation when I've recorded those, but `http.server` will do just as well.

> For what I understood, changing only the package/XML files makes no effects on the tests, since the VCR fixtures record the file contents as the request's response body. I think it would be nice to have some documentation for how to record again the VCR cassettes and for generating package/XML files, since people new to the project may not know how to do this.

The way the tests are currently done now -- yes, this is correct. Although it is possible to mirror a repository with a `file://`scheme URL pointed to the fixtures directory.

> I can create the documentation for that, splitting it in two parts:
> 
>     1. Generating RPM packages and XML metadata files;
> 
>     2. Serving these files in `localhost` for recording the VCR cassettes.
> 
> 
> What do you think about it? I could create a standalone markdown file and reference it in the README "Development Setup" section.

Sure, that sounds useful.

ikapelyukhin: > Hi @ikapelyukhin, I replaced the files (RPM packages/XML metadata) I edited manually with the ones generated in your OBS repository, on commit [27ba069](https://github.com/SUSE/rmt/commit/27ba0690485d81a8ca1480cb3c0cffb857e5fea2).
> 
> Just a few doubts:
> 
>     * Does it make sense to have "src" delta RPMs packages (`*.src.drpm`)? I noticed that OBS just generated delta RPMs for the `x86_64` packages but not for `src` ones, and that the previous deltainfo XML parser (`RMT::Rpm::DeltainfoXmlParser`) had the rule for checking if a delta RPM package has `arch == 'src'` when `mirror_src == false`.

In practice there are no source drpm packages.

> 
>     * The original fixture files included two versions for the same package (namely `0.0.1` and `0.0.2`), but after you committed a new version of the `apples` package to OBS, it is just serving the newer version and a delta RPM.

Yes, this is due to OBS project settings -- by default it keeps only the latest build artifact. I'm asking around to see if it's possible to set the project up to keep more that one version.

> Should we have both old and new versions for the fixtures?

The easiest solution to this perhaps would be to keep both repos and VCR cassettes -- the new one that you've recorded and the old one.

lib/rmt/mirror.rb
ikapelyukhin: > If we need to select the files in multiple locations and the criteria to select them change we only need to change the selection there.

We will never need to, though. And when and if we do -- then we can make those changes.

> Just a small addition: Would it make sense to abstract the selection of `primary_files` and `deltainfo_files` into a method?

It is really questionable if having one-line methods is better than having one-line idiomatic Ruby code.

spec/fixtures/files/dummy_repo/repodata/repomd.xml
lcaparroz: @ikapelyukhin, I reverted the changes to the original fixtures and created another 'dummy repo' with the new files, recording a new VCR cassette (commit bd4fd5e).

lib/rmt/mirror.rb
lcaparroz: HI @ikapelyukhin, you can see the extraction of the selection logic [here](https://github.com/SUSE/rmt/commit/687711c45d3cf2374480e924623b1d5b4e6843aa#diff-f9f254fce3569c11e6eac76b1a4d28aaR114-R121).

> It is really questionable if having one-line methods is better than having one-line idiomatic Ruby code.

Yes, I agree. In this case, I just wrapped the one-line idiomatic Ruby code into a method, so you can read the code block without needing to worry about how the selection is done ("low level" details). I know clean code can be too opinionated sometimes and usually it is a matter of personal taste.

Do you think the changes impaired the readability?

spec/fixtures/files/dummy_repo/repodata/repomd.xml
lcaparroz: @ikapelyukhin, about the packages fixtures, do you think we have any other considerations?

The main point of this conversation was I fiddled with the metadata "by hand". Resuming, we followed the path below:

1. You generated the `.rpm`, `.drpm` and `.src` packages I needed for the `mirror src` test scenario;
1. I reverted the changes I did in the existing metadata, packages and VCR cassete;
1. Then I created a new VCR cassete using the newly generated packages/metadata.

Now, the original fixtures are not touched anymore and there's a test case covering the mentioned scenario.

Do you think having the documentation on how to generate these packages here to be strictly necessary at this point or could we postpone it until we have proper and richer information on this?

package/obs/rmt-server.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue May 19 12:31:53 UTC 2020 - Luís Caparroz <lcaparroz@suse.com>
+
+- Version 2.5.8
+- Use repomd_parser gem to remove repository metadata parsing code
+  (GitHub issue gh#233).
```
ikapelyukhin, line 6: This is not a valid way to reference an issue in OBS. Either leave `gh#233` out completely and open a maintenance request when there another change file entry that has a valid identifier (`bsc#...`), or file a bug in Bugzilla and reference its number.

spec/lib/rmt/downloader_spec.rb
```diff
@@ -212,9 +212,10 @@
     context 'when download exceptions occur when ignore_errors is true' do
       let(:files) { %w[package1 package2 package3] }
       let(:checksum_type) { 'SHA256' }
+      let(:file_entry_classe) { Struct.new(:location, :checksum_type, :checksum, :type) }
       let(:queue) do
         files.map do |file|
-          RMT::Rpm::FileEntry.new(
+          file_entry_classe.new(
```
ikapelyukhin, line 218: `s/classe/class/`

spec/fixtures/files/dummy_repo/repodata/repomd.xml
ikapelyukhin: @lcaparroz no, the documentation is not strictly necessary. `repomd.xml` in the new fixture now has:

```
  <tags>
    <repo>obsrepository://build.opensuse.org/home:ikapelyukhin:dummy_repo/openSUSE_Leap_15.1</repo>
    <repo>obsbuildid:1580990394</repo>
  </tags>
```

That should give a pretty good hint how this repo came to life. :slightly_smiling_face:

package/obs/rmt-server.changes
digitaltom: Thinking about it, this change is not visible to customers, so not worth submitting a maintennance release for it. So we could rely on the next change to refer a proper bug?

lib/rmt/mirror.rb
ikapelyukhin: > Do you think the changes impaired the readability?

I do, and I'm annoyed by the faulty logic that has led to that situation (`It hypothetically might be useful in the future if certain special conditions are met` -- that's not a valid way to make engineering decisions). But I'm not your boss, so you do you :slightly_smiling_face:

package/obs/rmt-server.changes
ikapelyukhin: @digitaltom yep, that should be fine.

spec/lib/rmt/downloader_spec.rb
lcaparroz: Fixed on 54cb06e.

package/obs/rmt-server.changes
lcaparroz: I removed the GitHub issue reference from the changelog (0474241).

lib/rmt/mirror.rb
lcaparroz: Sure, it makes sense, it's not an optimization but it's premature. I reverted the changes in d09b486.

PR #556, opened 2020-04-24
feature: completely remove a system by login

### LINKED ISSUE #434: tool to remove orphaned registered client?
Is there already a tool to remove orphaned/old clients from the DB? or do we have to write our own cleanup jobs?

arharovets:
Fixes #434


lib/rmt/cli/systems.rb
```diff
@@ -28,4 +31,39 @@ def scc_sync
...
+  def find_system(target)
+    System.find_by!(login: target)
+  rescue ActiveRecord::RecordNotFound
+    raise SystemNotFoundException.new(_('System with login %{login} not found.') % { login: target })
+  end
+
+  def destroy_system!(system)
+    system.destroy!
+  rescue ActiveRecord::RecordNotDestroyed
+    raise SystemNotDestroyedException.new(_('System with login %{login} cannot be removed') % { login: target })
```
arharovets, line 67: I'm not entirely sure how to test this one. @ikapelyukhin could you help me, please?

ikapelyukhin: > I'm not entirely sure how to test this one. @ikapelyukhin could you help me, please?

Sure, I'll take a look at this soon.

arharovets: @ikapelyukhin thank you so much for your time! I agree with your position about `system.destroy!`, but I also think that you might want to implement soft deletion later, and, in that case, having one method that can call `purge` or mark for deletion depending on a case would be nice. However, if that's not the case, I will make the necessary changes as your approach is much cleaner.

ikapelyukhin: > you might want to implement soft deletion later, and, in that case, having one method that can call `purge` or mark for deletion depending on a case would be nice.

In this case whoever implements this will make necessary refactoring and write the code that makes sense for that specific scenario there and then. You are sacrificing clean and maintainable code for solving a hypothetical problem, which maybe doesn't even need to be solved.

arharovets: you're right, thank you :+1: I will make the necessary changes today.

arharovets:
> https://github.com/SUSE/rmt/blob/179b9c56774435132aaacf7200e67387e95a82c7/lib/rmt/cli/systems.rb#L44-L46
> 
> This was the proper way to do it, those exceptions do need to be handled in tests though.

@ikapelyukhin I see it now, thanks for pointing this out! :+1:

digitaltom:
Will this lso drop the system from SCC when systems are synced there (`rmt-cli scc sync-systems`)?

ikapelyukhin:
> Will this lso drop the system from SCC when systems are synced there (`rmt-cli scc sync-systems`)?

Because of how amazingly well-engineered SCC sync is (:grimacing:) -- yes. If you do `system.destroy` from Rails console -- that'd get synced as well.

digitaltom:
A changelog line would be good then, we already had someone asking for exactly this feature yesterday on the mailing list :-)

ikapelyukhin:
> A changelog line would be good then, we already had someone asking for exactly this feature yesterday on the mailing list :-)

It would be good to merge first (this and other PRs) and then have somebody do the release.
We don't expect 3rd-party contributors to do RPM packaging stuff, do we?
We also have some MRs pending release, depending on their scheduled release date we might want to wait a little bit.

tmuntaner (approved):
@arharovets thank you for your contribution.

PR #571, opened 2020-06-09
Fix RPM dependencies packages

lcaparroz:
Some of the rmt-server RPM package's dependencies (mariadb and nginx) are marked as required, but they aren't really required to run, rmt-server, especially for containerized usage.

This PR marks these packages as recommended and makes some other package adjustments.

### DIFF
--- a/package/obs/rmt-server.changes
+++ b/package/obs/rmt-server.changes
@@ -1,4 +1,15 @@
 -------------------------------------------------------------------
+Tue Jun  9 07:16:15 UTC 2020 - Thorsten Kukuk <kukuk@suse.com>
+
+- Move nginx/vhosts.d directory to correct sub-package. They
+  are needed together with nginx, not rmt-server.
+- Fix dependencies especially for containerized usage:
+  - mariadb and nginx are not hard requires, could run on another host
+- Fix generic dependencies:
+  - systemd ordering was missing
+  - shadow is required for pre-install
+
+-------------------------------------------------------------------
 Tue May 19 12:31:53 UTC 2020 - Luís Caparroz <lcaparroz@suse.com>
 
 - Version 2.5.8
--- a/package/obs/rmt-server.spec
+++ b/package/obs/rmt-server.spec
@@ -47,12 +47,13 @@ BuildRequires:  libxml2-devel
 BuildRequires:  libxslt-devel
 BuildRequires:  pkgconfig(systemd)
 Requires:       gpg2
-Requires:       mariadb
-Requires:       nginx
+Recommends:     mariadb
+Recommends:     nginx
+# The config is not really required by rmt-server, but by nginx ...
 Requires:       rmt-server-configuration
 Requires(post): %{ruby_version}
 Requires(post): %{rubygem bundler}
-Requires(post): shadow
+Requires(pre):  shadow
 Requires(post): timezone
 Requires(post): util-linux
 Conflicts:      yast2-rmt < 1.0.3
@@ -60,6 +61,7 @@ Recommends:     yast2-rmt >= 1.3.0
 Recommends:     rmt-server-config
 # Does not build for i586 and s390 and is not supported on those architectures
 ExcludeArch:    %{ix86} s390
+%{systemd_ordering}
 
 %description
 This package provides a mirroring tool for RPM repositories and a registration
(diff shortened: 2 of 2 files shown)


suse-tests-fail:
Prophet reports failure.
https://ci.suse.de/job/scc-RMT-integration-tests/108496/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment.
Failed fast in project rmt on step docker-compose-build-images. (Merged f484204775c4bfdf844682d2180067bfc8c11789 into bbff48aa07415f432017ea1c0b697448543b1e2c)

ikapelyukhin:
@thkukuk hmm, we can probably make both `rmt-server-config` and `rmt-server-pubcloud` to require both MariaDB and nginx. You could then create an empty dummy package `rmt-server-docker` that doesn't require or recommend them and have it provide `rmt-server-configuration` capability (and maybe use it for whatever other stuff that might be required for docker deployments).

MariaDB and nginx aren't really optional requirements in the regular bare metal/VM use case. My concern is that users will install it `zypper --no-recommends` and then file support tickets.

Thoughts?

suse-tests-fail:
Prophet reports failure.
https://ci.suse.de/job/scc-RMT-integration-tests/108503/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment.
Failed fast in project rmt on step docker-compose-build-images. (Merged f484204775c4bfdf844682d2180067bfc8c11789 into bbff48aa07415f432017ea1c0b697448543b1e2c)

thkukuk:
> But then again, we have `Recommends: rmt-server-config`. OK, let's not overly complicate things and move forward with this slightly_smiling_face

The "Recommends: rmt-server-config" seems to be only a hint for zypper which rmt-server-configuration package to install. "rmt-server-configuration" is a Requires. Not sure if the "Recommends" here should be even a "Suggests", would be something for the solver experts.

ikapelyukhin:
> > But then again, we have `Recommends: rmt-server-config`. OK, let's not overly complicate things and move forward with this slightly_smiling_face
> 
> The "Recommends: rmt-server-config" seems to be only a hint for zypper which rmt-server-configuration package to install. "rmt-server-configuration" is a Requires. Not sure if the "Recommends" here should be even a "Suggests", would be something for the solver experts.

@thkukuk OK, what's your preference between `rmt-server-docker` and what's currently in this PR then?

thkukuk:
> @thkukuk hmm, we can probably make both `rmt-server-config` and `rmt-server-pubcloud` to require both MariaDB and nginx. You could then create an empty dummy package `rmt-server-docker` that doesn't require or recommend them and have it provide `rmt-server-configuration` capability (and maybe use it for whatever other stuff that might be required for docker deployments).

Sounds like an option, this way, you could even provide a configuration for Apache and don't force the user to use nginx. But the nginx configuration does not need mariadb, so not really fully correct.

> MariaDB and nginx aren't really optional requirements in the regular bare metal/VM use case. 

They are. There no no reason to enforce the user to run the mariadb database on the same host as RMT, you could also use an existing mariadb instance. RMT requires a mariadb database in the network, not on the same machine. Same for nginx.

> My concern is that users will install it `zypper --no-recommends` and then file support tickets.

Users will do that anyways, but if they know about --no-recommends, they also now what to do in such a case. Our experience is: the number of people using --no-recommends and complaining is really very low.

The recommends are easier for people who want to build a container out of SLES without the need to create own, additional packages.

But I don't care much as long as our SEs stop to use my patched rmt-server package but use the official one for RMT in a container setup. 
Between, podman is our tool of choice in our products (SLES, SES, CaaSP, MicroOS, Kubic, ...), not docker. And there are also solutions from our SEs for a containerized RMT using podman beside my kubernetes solution. So don't use "docker" as name.

thkukuk:
> @thkukuk OK, what's your preference between `rmt-server-docker` and what's currently in this PR then?

As written in the last comment, I prefer something which can be used out of the SLES box, without the need for our SEs or customers to create at first another RPM before they can build a container with RMT.

And please, call it 'rmt-server-container', not 'rmt-server-docker'. We don't use docker anymore in our products by default.
If rmt-server builds a 'rmt-server-container' sub-package, which would be part of SLES15, I don't care much about which of the two options we will use.

suse-tests-fail:
Prophet reports failure.
https://ci.suse.de/job/scc-RMT-integration-tests/108510/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment.
Failed fast in project rmt on step docker-compose-build-images. (Merged f484204775c4bfdf844682d2180067bfc8c11789 into bbff48aa07415f432017ea1c0b697448543b1e2c)

suse-tests-fail:
Prophet reports failure.
https://ci.suse.de/job/scc-RMT-integration-tests/108515/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment.
Failed fast in project rmt on step docker-compose-run-tests. (Merged f484204775c4bfdf844682d2180067bfc8c11789 into bbff48aa07415f432017ea1c0b697448543b1e2c)

PR #575, opened 2020-06-29
Move reloading of nginx to correct sub-packages

thkukuk:
The posttrans in the spec file reloads nginx for the main package, but the nginx configuration is in sub-packages. So if a configuration sub-package without nginx config is installed, this leads to a systemctl/RPM error.


wstephenson:
Closing because replaced by #591

PR #592, opened 2020-07-31
Enable 'Installer-Updates' repositories by default

### LINKED ISSUE #554: Enable mirroring of "Installer-Updates" repositories by default
When enabling a product for mirroring in RMT, only the repositories that will get enabled on the system initially will get mirrored by default. Example: 

```
rmt-cli products enable SLES/12.5/x86_64
Found product by target SLES/12.5/x86_64: SUSE Linux Enterprise Server 12 SP5 x86_64.
Enabling SUSE Linux Enterprise Server 12 SP5 x86_64:
  SUSE Linux Enterprise Server 12 SP5 x86_64:
    Enabled repository SLES12-SP5-Pool.
    Enabled repository SLES12-SP5-Updates.
```

This leaves out the "Installer-Updates" repositories, that can get used by the installer to patch the installation system itself.

jesusbv:
This Fixes #554


digitaltom:
I think this pr wouldn't enable the mirroring of installer repos, but enable the repo by default on the client systems?

jesusbv:
> I think this pr wouldn't enable the mirroring of installer repos, but enable the repo by default on the client systems?

Maybe I misunderstood the issue? 
When running `rmt-cli products enable <product>` the Installer-Updates repositories are not enabled and we want to them to be enabled by default so they get mirrored. 

Correct me if wrong, in order to set them enabled (mirroring_enabled) once they have been fetched from SCC, they need to be accordingly updated while syncing (`rmt-cli sync`).

In the example, the output, that the change produces after syncing, is
```
rmt-cli products enable SLES/15.1/x86_64
Found product by target SLES/15.1/x86_64: SUSE Linux Enterprise Server 15 SP1 x86_64.
Enabling SUSE Linux Enterprise Server 15 SP1 x86_64:
  SUSE Linux Enterprise Server 15 SP1 x86_64:
    Enabled repository SLE-Product-SLES15-SP1-Pool.
    Enabled repository SLE-Product-SLES15-SP1-Updates.
    Enabled Repository SLES15-SP1-Installer-Updates.
```
If we want _all_ Installer-Updates repositories (of any product) I can update the PR to reflect that.

digitaltom:
Yes, but isn't the `mirroring_enabled` flag the one to change here, and not `enabled`?

jesusbv:
Oh yes ! Fixed.

lib/rmt/scc.rb
```diff
@@ -171,6 +171,7 @@ def create_product(item, root_product_id = nil, base_product = nil, recommended
   def create_service(item, product)
     product.create_service!
     item[:repositories].each do |repo_item|
+      repo_item[:mirroring_enabled] = true if repo_item[:installer_updates] and not repo_item[:mirroring_enabled]
```
digitaltom, line 174: would there be any different behavior if we skip `and not repo_item[:mirroring_enabled]` here?

jesusbv: Not right now but I don't know if that's going to change in the future and the installer-updates repositories will come already enabled.

digitaltom: But even if they come already enabled that part of the code would make any difference?

jesusbv: Well, technically no, I wanted to be sure that even if they come wrong, we set them right.
I will remove it.

spec/lib/rmt/cli/products_spec.rb
```diff
@@ -488,6 +488,10 @@
     end
 
     describe 'success' do
+      it 'has mirroring enabled Installer Updates repositories by default' do
+        expect(expected_output).to include('Installer-Updates')
```
lcaparroz, line 492: This expectation is a bit fragile, it's asserting over a local variable and not from the result of some operation. In fact, no operation from the class under test is executed. Taking a look the other examples in the same group, we could have something like:

```suggestion
        expect { described_class.start(argv) }
          .to output(/Installer-Updates/).to_stdout
          .and output('').to_stderr
```

However, testing only the output could still be prone to false positives: the command generating an output containing a string like `'Installer-Updates'` doesn't ensure the repository has the attribute `mirroring_enabled == true` in the `repository` database table.

Maybe we can think of a more robust test scenario and I would suggest implementing it for the `RMT::SCC` class, since the logic for enabling 'Installer-Updates` repos by default is there. It seems `spec/lib/rmt/cli/products_spec.rb` only has expectation for text output. What do you think?

jesusbv: Ok, I used the same output as suggested in the issue but it didn't feel quite right.
I will implement the test for `RMT::SCC` class then.

jesusbv: Done.

PR #593, opened 2020-08-03
Exclude installer updates from registered machines

### LINKED ISSUE #559: Installer updates repos are added to registered systems
I've noticed that clients registered to RMT and to SCC have a different number of repos after registration -- RMT-registered clients have an extra repo, `SLE15-SP1-Installer-Updates`.

Seems like zypper service XML endpoint on SCC side filters out installer update repos and RMT doesn't.

It doesn't make sense to add installer repos to the registered machines, those are supposed to be used during installation only.

jesusbv:
This Fixes #559


package/obs/rmt-server.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Aug 25 10:27:33 UTC 2020 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
+
+- Version 2.5.15
+- Exclude installer updates repos to registered machines.
```
tmuntaner, line 5:
```suggestion
- RMT no longer provides the installer updates repository to systems via its zypper service. This repository is used during the installation process, as it provides an up-to-date installation experience, but it has no use on an already installed system.
```

tmuntaner (changes requested):
@jesusbv this looks good so far, but besides my small suggestion, we've already used version 2.5.15, so a new version should be used.

tmuntaner (approved):

PR #585, opened 2020-07-20
Downloaded data integrity

lcaparroz:
## Why

In order to assure downloaded data integrity between database, server disk and repository metadata, we need to start tracking all files in the database, even if they were just deduplicated.

## What

- [x] the uniqueness constraint on the index `checksum + checksum_type` on the `downloaded_files` table, in order to allow tracking deduplicated files.
- [x] Start to tracking files on deduplication (copy/hardlink file and then add it to the database).
- [x] Add files to the database if they've been deduplicated previously but are untracked.
- [x] ~Enforce uniqueness to the `local_path` field on the `downloaded_files` table; there shouldn't be two entries with the same local path.~ Add *on the fly* duplicates check to assure there is only one entry for a given local path (P.S.: enforcing uniqueness on the database could cause disruption to the customer, since we can't assert there'll be no duplicates prior to this change). 

## Considerations

We decided to maintain the data consistency during the mirroring process instead of providing separated commands or flags.

### Assumptions

* The repository metadata is the authoritative source of information (e.g.: if file on disk and on database matches, but the checksum is different from the repository metadata, file/database is corrupted);
* RMT should always track files, unless `airgap_mode` is enabled.

### Decision logic for file tracking

Here is a summary of how RMT decides whether it should download/deduplicate/track files or not.

#### Conditions

1. File with given path exists on disk
2. File on disk matches metadata checksum
3. Entry on DB w/ metadata path exists on database
4. Entry on DB w/ metadata checksum exists on database

| 1 | 2 | 3 | 4 | Status | Action |
|:----------------:|:----------------:|:----------------:|:----------------:|:----------------------------|:---------------------------------------|
|:x:               |:x:               |:x:               |:x:               | New file                    | Download -> Track                      |
|:x:               |:x:               |:x:               |:heavy_check_mark:| Same file on another path   | Deduplicate -> Track                   |
|:x:               |:x:               |:heavy_check_mark:|:x:               | Missing on disk             | Untrack -> Download -> Track           |
|:x:               |:x:               |:heavy_check_mark:|:heavy_check_mark:| Missing on disk             | Untrack -> Download -> Track           |
|:x:               |:heavy_check_mark:|:x:               |:x:               | (Impossible)                | ---                                    |
|:x:               |:heavy_check_mark:|:x:               |:heavy_check_mark:| (Impossible)                | ---                                    |
|:x:               |:heavy_check_mark:|:heavy_check_mark:|:x:               | (Impossible)                | ---                                    |
[...]

PR #594, opened 2020-08-03
Downloaded data integrity improvement

lcaparroz:
## Why

@ikapelyukhin pointed out some issues on PR #585, which should be addressed in order to assure data integrity on the customer database without affecting the mirroring process performance.

## What

* `DownloadedFile#local_path` uniqueness: I created a new migration which will remove possible duplicated `local_path` entries, if any, prior to adding an index with an uniqueness constraint to the field.
* I removed all local file validation logic from `DownloadedFile` and `RMT::Deduplicator`: now, all validation and data integrity logic is on `RMT::FileValidator`, which also fix invalid DB entries when searching for a  candidate source file for deduplication.
* Right now, file validation before deduplicating/downloading works without changes: `RMT::FileValidator` has an option for "deep verifying," which will be implemented as command flag in another PR;
* I tried to separate concerns for each of the classes I touched.

## Open to Discussion

* The `downloaded_file` table does not enforce any validation on blank fields: right now we can have entries with blank checksums, size or local paths. Should we think about it?
* I removed the exception `RMT::Deduplicator::MismatchException`, since it was only used to log to the `STDOUT` when a `downloaded_file` entry had no file matched on disk or the file had wrong data (invalid size or checksum) but RMT would take no action on top of this error, just ignore. There is the possibility of enhancing the class `RMT::FileValiator` to log its actions, like when tracking/untracking a file on DB, validating a file (invalid checksum or file does not exist) or removing invalid files. I'm not sure how much of this information would have some value to the customer.
* I'm planning to add the flag `--deep-verify` to the command `rmt-cli mirror`, in another PR. Should we consider it for other commands too?

### DIFF
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -19,26 +19,20 @@ class DownloadedFile < ApplicationRecord
       return false unless File.exist?(path)
 
       valid_checksum = RMT::ChecksumVerifier.match_checksum?(checksum_type, checksum_value, path)
-      file_db_entries = where(local_path: path)
-      matched_tracked_file = file_db_entries.find do |entry|
-        match_tracked_file?(entry, checksum_type, checksum_value)
-      end
-
-      duplicated_entries = file_db_entries.select { |entry| entry != matched_tracked_file }
-      duplicated_entries.each(&:destroy)
+      tracked_file = find_by(local_path: path)
+      matched_tracked_file = matches_tracked_file?(tracked_file, checksum_type, checksum_value)
 
       return true if valid_checksum && matched_tracked_file
-      return !add_file(checksum_type, checksum_value, path).nil? if valid_checksum && matched_tracked_file.nil?
+      return !add_file(checksum_type, checksum_value, path).nil? if valid_checksum && tracked_file.nil?
 
       FileUtils.remove_file(path, force: true)
-      matched_tracked_file.destroy unless matched_tracked_file.nil?
-
+      tracked_file.destroy unless tracked_file.nil?
       false
     end
 
     private
 
-    def match_tracked_file?(tracked_file, checksum_type, checksum_value)
+    def matches_tracked_file?(tracked_file, checksum_type, checksum_value)
       return false if tracked_file.nil? || tracked_file.checksum_type != checksum_type || tracked_file.checksum != checksum_value
       true
     end
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -20,4 +20,8 @@ class DownloadedFile < ApplicationRecord
     end
 
   end
+
+  def size
+    file_size
+  end
 end
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -1,10 +1,5 @@
 class DownloadedFile < ApplicationRecord
   class << self
-
-    def get_local_path_by_checksum(checksum_type, checksum)
-      DownloadedFile.find_by({ checksum_type: checksum_type, checksum: checksum })
-    end
-
     def track_file(checksum_type:, checksum:, local_path:, size:)
       find_or_initialize_by(local_path: local_path).tap do |record|
         record.checksum_type = checksum_type
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -7,13 +7,8 @@ class DownloadedFile < ApplicationRecord
         record.file_size = size
 
         record.save if record.changed?
-      end.persisted?
+      end
     end
-
-    def untrack_file(local_path)
-      where(local_path: local_path).destroy_all
-    end
-
   end
 
   def size
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -5,27 +5,6 @@ class DownloadedFile < ApplicationRecord
       DownloadedFile.find_by({ checksum_type: checksum_type, checksum: checksum })
     end
 
-    def valid_local_file?(checksum_type, checksum_value, path)
-      return false unless File.exist?(path)
-
-      valid_checksum = RMT::ChecksumVerifier.match_checksum?(checksum_type, checksum_value, path)
-      tracked_file = find_by(local_path: path)
-      matched_tracked_file = matches_tracked_file?(tracked_file, checksum_type, checksum_value)
-
-      return true if valid_checksum && matched_tracked_file
-
-      if valid_checksum && tracked_file.nil?
-        return !track_file(checksum_type: checksum_type,
-                           checksum: checksum_value,
-                           local_path: path,
-                           size: File.size(path)).nil?
-      end
-
-      FileUtils.remove_file(path, force: true)
-      tracked_file.destroy unless tracked_file.nil?
-      false
-    end
-
     def track_file(checksum_type:, checksum:, local_path:, size:)
       find_or_initialize_by(local_path: local_path).tap do |record|
         record.checksum_type = checksum_type
@@ -38,13 +17,6 @@ class DownloadedFile < ApplicationRecord
 
     def untrack_file(local_path)
       where(local_path: local_path).destroy_all
-    end
-
-    private
-
-    def matches_tracked_file?(tracked_file, checksum_type, checksum_value)
-      return false if tracked_file.nil? || tracked_file.checksum_type != checksum_type || tracked_file.checksum != checksum_value
-      true
     end
 
   end
--- a/app/models/downloaded_file.rb
+++ b/app/models/downloaded_file.rb
@@ -1,16 +1,6 @@
 class DownloadedFile < ApplicationRecord
   class << self
 
-    def add_file(checksum_type, checksum, local_path)
-      return unless local_path.match?(/\.(rpm|drpm)$/)
-
-      file_size = File.size(local_path)
-      DownloadedFile.find_or_create_by({ checksum_type: checksum_type,
-                                         checksum: checksum,
-                                         local_path: local_path,
-                                         file_size: file_size })
-    end
-
     def get_local_path_by_checksum(checksum_type, checksum)
       DownloadedFile.find_by({ checksum_type: checksum_type, checksum: checksum })
     end
@@ -23,11 +13,31 @@ class DownloadedFile < ApplicationRecord
       matched_tracked_file = matches_tracked_file?(tracked_file, checksum_type, checksum_value)
 
       return true if valid_checksum && matched_tracked_file
-      return !add_file(checksum_type, checksum_value, path).nil? if valid_checksum && tracked_file.nil?
+
+      if valid_checksum && tracked_file.nil?
+        return !track_file(checksum_type: checksum_type,
+                           checksum: checksum_value,
+                           local_path: path,
+                           size: File.size(path)).nil?
+      end
 
       FileUtils.remove_file(path, force: true)
       tracked_file.destroy unless tracked_file.nil?
       false
+    end
+
+    def track_file(checksum_type:, checksum:, local_path:, size:)
+      find_or_initialize_by(local_path: local_path).tap do |record|
+        record.checksum_type = checksum_type
+        record.checksum = checksum
+        record.file_size = size
+
+        record.save if record.changed?
+      end.persisted?
+    end
+
+    def untrack_file(local_path)
+      where(local_path: local_path).destroy_all
     end
 
     private
--- a/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
+++ b/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
@@ -3,14 +3,11 @@ class AddUniquenessToDownloadedFilesLocalPath < ActiveRecord::Migration[5.2]
     # Remove duplicates before adding uniqueness to `local_path`
     ActiveRecord::Base.connection.execute(
       <<~SQL
-        DELETE FROM `downloaded_files`
-        WHERE `downloaded_files`.`id` NOT IN (
-            SELECT * FROM (
-                SELECT max(`files`.`id`) AS unique_id FROM `downloaded_files` AS `files`
-                INNER JOIN `downloaded_files` ON `downloaded_files`.`id`=`files`.`id`
-                GROUP BY `files`.`local_path`
-            ) as tmp
-        )
+        DELETE df1 FROM
+            downloaded_files AS df1,
+            downloaded_files AS df2
+            WHERE df1.id < df2.id AND
+                df1.local_path = df2.local_path
       SQL
     )
 
--- a/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
+++ b/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
@@ -1,0 +1,9 @@
+class AddUniquenessToDownloadedFilesLocalPath < ActiveRecord::Migration[5.2]
+  def change
+    # Remove duplicates before adding uniqueness to `local_path`
+    unique_ids = DownloadedFile.group(:local_path).select('max(id)')
+    DownloadedFile.where.not(id: unique_ids).delete_all
+
+    add_index :downloaded_files, :local_path, unique: true
+  end
+end
--- a/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
+++ b/db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
@@ -1,8 +1,18 @@
 class AddUniquenessToDownloadedFilesLocalPath < ActiveRecord::Migration[5.2]
   def change
     # Remove duplicates before adding uniqueness to `local_path`
-    unique_ids = DownloadedFile.group(:local_path).select('max(id)')
-    DownloadedFile.where.not(id: unique_ids).delete_all
+    ActiveRecord::Base.connection.execute(
+      <<~SQL
+        DELETE FROM `downloaded_files`
+        WHERE `downloaded_files`.`id` NOT IN (
+            SELECT * FROM (
+                SELECT max(`files`.`id`) AS unique_id FROM `downloaded_files` AS `files`
+                INNER JOIN `downloaded_files` ON `downloaded_files`.`id`=`files`.`id`
+                GROUP BY `files`.`local_path`
+            ) as tmp
+        )
+      SQL
+    )
 
     add_index :downloaded_files, :local_path, unique: true
   end
--- a/lib/rmt/deduplicator.rb
+++ b/lib/rmt/deduplicator.rb
@@ -23,10 +23,10 @@ class RMT::Deduplicator
       end
 
       if track
-        ::DownloadedFile.track_file(checksum: target_file.checksum,
-                                    checksum_type: target_file.checksum_type,
-                                    local_path: target_file.local_path,
-                                    size: File.size(target_file.local_path))
+        DownloadedFile.track_file(checksum: target_file.checksum,
+                                  checksum_type: target_file.checksum_type,
+                                  local_path: target_file.local_path,
+                                  size: File.size(target_file.local_path))
       end
 
       true
@@ -35,13 +35,13 @@ class RMT::Deduplicator
     private
 
     def hardlink(src, dest)
-      ::FileUtils.ln(src, dest)
+      FileUtils.ln(src, dest)
     rescue StandardError
-      raise ::RMT::Deduplicator::HardlinkException.new("#{src} → #{dest}")
+      raise RMT::Deduplicator::HardlinkException.new("#{src} → #{dest}")
     end
 
     def copy(src, dest)
-      ::FileUtils.cp(src, dest)
+      FileUtils.cp(src, dest)
     end
 
     def make_file_dir(file_path)
--- a/lib/rmt/deduplicator.rb
+++ b/lib/rmt/deduplicator.rb
@@ -10,6 +10,34 @@ class RMT::Deduplicator
 
   class << self
 
+    def deduplicate(target_file, force_copy: false, track: true)
+      src = DownloadedFile.get_local_path_by_checksum(target_file.checksum_type,
+                                                      target_file.checksum)
+
+      if src.nil?
+        return false
+      elsif !File.exist?(src.local_path) || (src.file_size != File.size(src.local_path))
+        raise MismatchException.new(src.local_path)
+      end
+
+      make_file_dir(target_file.local_path)
+
+      if RMT::Config.deduplication_by_hardlink? && !force_copy
+        hardlink(src.local_path, target_file.local_path)
+      else
+        copy(src.local_path, target_file.local_path)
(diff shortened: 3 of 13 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/123614/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 99ff9c130dde536b1d3f4d275d0e10331fc5bf4a into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

app/models/downloaded_file.rb
```diff
@@ -1,47 +1,22 @@
...
-    def get_local_path_by_checksum(checksum_type, checksum)
-      DownloadedFile.find_by({ checksum_type: checksum_type, checksum: checksum })
+    def track_file(checksum_type:, checksum:, local_path:, size:)
+      find_or_initialize_by(local_path: local_path).tap do |record|
+        record.checksum_type = checksum_type
+        record.checksum = checksum
+        record.file_size = size
+
+        record.save if record.changed?
+      end.persisted?
```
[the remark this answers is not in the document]
lcaparroz: In fact, the return value is not used. I guess when I first thought about this function, since `model#save` doesn't throw an error, it could be useful to verify if the "track file" transaction was successful, but I don't verify it anywhere.

I don't know if RMT should really care about this, because unless the connection with the database fails, `model#save` should not fail; I used `#find_or_initialize_by` to avoid uniqueness validation errors.

I think it's safe to remove this `persisted?` by now. Any thoughts?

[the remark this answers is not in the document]
gionaufal: I think it makes sense to remove it, because right now it's just overhead since it's not being used.

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+      verify_file_on_disk(file_reference, deep_verify).tap do |valid_on_disk|
+        update_file_status_on_database(file_reference, valid_on_disk)
+      end
+    end
+
+    def find_valid_file_by_checksum(checksum, checksum_type, deep_verify:)
+      ::DownloadedFile
+        .where(checksum: checksum, checksum_type: checksum_type)
+        .map { |file| [file, validate_tracked_file_on_disk(file, deep_verify)] }
+        .find { |_, valid_on_disk| valid_on_disk }
```
ikapelyukhin, line 16: Instead of mapping an array of file records into a 2-dimensional array -- I'd consider expanding `.map` and `.find` into a regular `.each` loop.

This mega-chain of 5 methods isn't making things any easier to understand.

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+        &.first&.local_path
+    end
+
+    private
+
+    def validate_tracked_file_on_disk(file_record, deep_verify)
```
ikapelyukhin, line 22: I'd consider renaming this method into `is_orphaned_db_record?`.

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+      verify_file_on_disk(file_record, deep_verify).tap do |valid_on_disk|
+        file_record.destroy unless valid_on_disk
+      end
+    end
+
+    def verify_file_on_disk(file, deep_verify)
+      file_exist = File.exist?(file.local_path)
```
ikapelyukhin, line 29: So if the file doesn't exist, `match_metadata = false` and you aren't not removing the file and just returning false.

You can just return false immediately: `return false unless File.exist?(file.local_path)`?

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+      match_metadata = file_exist ? file_match_metadata?(file, deep_verify) : false
```
ikapelyukhin, line 30: Or you could do:

```ruby
match_metadata = (File.size(file.local_path) == file.size)
if deep_verify
  match_metadata &= RMT::ChecksumVerifier.match_checksum?(
    file.checksum_type, file.checksum, file.local_path
  )
end
```

And then you don't need those one-use one-line methods and you don't have to pass `deep_verify` two levels down into them.

Also `match_metadata` should be called either `matches_metadata` or better yet `is_valid_metadata`.

lib/rmt/mirror.rb
```diff
@@ -4,6 +4,29 @@
...
+            file.checksum = metadata.checksum
+            file.checksum_type = metadata.checksum_type
+            file.size = metadata.size
+          end
+      end
+    end
+
+    attr_reader :local_path, :relative_remote_path
+    attr_accessor :arch, :checksum, :checksum_type, :size
+    alias location relative_remote_path
```
ikapelyukhin, line 22: You aren't using `relative_remote_path` anywhere outside the class, drop the alias use `location` everywhere instead.

lib/rmt/mirror.rb
```diff
@@ -135,9 +158,15 @@ def mirror_data(primary_files, deltainfo_files)
     @downloader.destination_dir = @repository_dir
     @downloader.cache_dir = nil
 
-    package_files =
+    package_references =
       parse_mirror_data_files(deltainfo_files, RepomdParser::DeltainfoXmlParser) +
       parse_mirror_data_files(primary_files, RepomdParser::PrimaryXmlParser)
+
+    package_files = package_references.map do |reference|
```
ikapelyukhin, line 165: `package_files` is still made up of `FileReference` instances. :man_shrugging:

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      return true if match_metadata
+
+      FileUtils.remove_file(file.local_path, force: true) if file_exist
+
+      false
+    end
+
+    def update_file_status_on_database(file, valid_on_disk)
```
ikapelyukhin, line 39: `s/update_file_status_on_database/update_file_status_in_database/`

app/models/downloaded_file.rb
```diff
@@ -1,47 +1,22 @@
...
-
-      return true if valid_checksum && matched_tracked_file
-      return !add_file(checksum_type, checksum_value, path).nil? if valid_checksum && matched_tracked_file.nil?
-
-      FileUtils.remove_file(path, force: true)
-      matched_tracked_file.destroy unless matched_tracked_file.nil?
-
-      false
+    def untrack_file(local_path)
+      where(local_path: local_path).destroy_all
```
ikapelyukhin, line 14: One line method used in one place.

lib/rmt/deduplicator.rb
```diff
@@ -2,14 +2,36 @@
...
+      make_file_dir(target_file.local_path)
+
+      if RMT::Config.deduplication_by_hardlink? && !force_copy
+        hardlink(source_file_path, target_file.local_path)
+      else
+        copy(source_file_path, target_file.local_path)
+      end
+
+      if track
+        ::DownloadedFile.track_file(checksum: target_file.checksum,
```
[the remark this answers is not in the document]
lcaparroz: In fact, I've been consistent with the use of the `::` operator, mainly because of the multiple module levels, though `DownloadedFiles` would never conflict with other names. Greping at the `lib/` dir, it seems the codebase is already not consistent about this style. Do you think it's better to "fix" my changes or to make another PR for style normalization?

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
+require 'fileutils'
+require 'rmt/checksum_verifier'
+
+class RMT::FileValidator
+  class << self
+    def validate_local_file(file_reference, deep_verify:)
+      verify_file_on_disk(file_reference, deep_verify).tap do |valid_on_disk|
```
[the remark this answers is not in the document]
lcaparroz: The method returns the value of `valid_on_disk`, not of `update_file_status_on_database`; the database is a necessary step but it doesn't affect the method outcome. Using `.tap` was just a way to group both steps: updating the databse only happens if a file is verified on disk, but sure it's not the only one, I can do it like:

```ruby
valid_on_disk = verify_file_on_disk(file_reference, deep_verify)
update_file_status_on_database(file_reference, valid_on_disk)
valid_on_disk
```

spec/lib/rmt/deduplicator_spec.rb
```diff
@@ -14,83 +20,149 @@
...
+
+        it 'removes invalid source files' do
+          deduplicate
+
+          expect(File.exist?(source_path)).to be false
+        end
+
+        it 'untracks invalid source files' do
+          expect { deduplicate }
+            .to change { DownloadedFile.where(local_path: source_path).count }
```
ikapelyukhin, line 104: @lcaparroz consistent, you say? :slightly_smiling_face:

lib/rmt/deduplicator.rb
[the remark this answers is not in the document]
ikapelyukhin: It's not consistent because not everybody understands how ActiveSupport works :slightly_smiling_face: 
Do whatever is easier, if you don't want to go searching for the lines you've changed in this PR -- I'm fine with fixing everything later.

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
...
+        .map { |file| [file, validate_tracked_file_on_disk(file, deep_verify)] }
+        .find { |_, valid_on_disk| valid_on_disk }
+        &.first&.local_path
+    end
+
+    private
+
+    def validate_tracked_file_on_disk(file_record, deep_verify)
+      verify_file_on_disk(file_record, deep_verify).tap do |valid_on_disk|
+        file_record.destroy unless valid_on_disk
```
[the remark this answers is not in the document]
lcaparroz: It's the same with the method `#validate_local_file`, I need the method to return whether a file is valid on disk or not:

```ruby
valid_on_disk = verify_file_on_disk(file_record, deep_verify)
file_record.destroy unless valid_on_disk
valid_on_disk
```

The return value is used to filter which files I can use as a candidate for deduplication in `#find_valid_file_by_checksum`.

spec/lib/rmt/deduplicator_spec.rb
lcaparroz: Specs don't have problems with module levels, as you would expect in a regular project file, you usually don't need the `::` to avoid namespace conflicts. I could say I've been consistent at two different levels: tests, which don't need `::` at all, and application files, [where it might be a good idea to be explicit about namespaces](https://github.com/rubocop-hq/ruby-style-guide#namespace-definition).

lib/rmt/file_validator.rb
lcaparroz: In fact, I don't think `.each` makes things less complicated: `.each` would have the same function as `.map`, to apply `validate_tracked_file_on_disk` on **each** file. In the end, I'd still need to get rid of all invalid files on disk and choose one valid file to returns its path. I agree the two-dimensional array can be a bit confusing, though Ruby hashes are iterated the same way.

I can write something like this to remove one level of chaining:

```ruby
      ::DownloadedFile
        .where(checksum: checksum, checksum_type: checksum_type)
        .select { |file| validate_tracked_file_on_disk(file, deep_verify) }
        &.first&.local_path
```

However, I'm not figuring out how to use `.each` to make it more readable. Do you have any suggestions?

lib/rmt/file_validator.rb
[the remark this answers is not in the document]
tmuntaner: @lcaparroz, I would agree with @ikapelyukhin here. `tap` may be overkill for this use-case and, while grammatically correct, falls outside its normal style and can be confusing. This may be the time to be overly verbose and less clever :(.

lib/rmt/file_validator.rb
tmuntaner: Yes, the methods have become confusing with similar names like `validate_tracked_file_on_disk` and `verify_file_on_disk`. You can only really tell their meanings by looking at their code.

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,59 @@
+require 'fileutils'
+require 'rmt/checksum_verifier'
+
+class RMT::FileValidator
+  class << self
+    def validate_local_file(file_reference, deep_verify:)
```
tmuntaner, line 6: We're using this `deep_verify` variable, but I can only find one usage, and it's non-variable.

```
    return false if ::RMT::FileValidator.validate_local_file(file, deep_verify: false)
```

Does it need to be a variable at all?

lib/rmt/file_validator.rb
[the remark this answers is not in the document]
tmuntaner: Also, @lcaparroz, I can only find one usage of `deep_verify` and it's `false`. Maybe it's unneeded.

lib/rmt/file_validator.rb
lcaparroz: In fact, I let this `deep_verify` variable to implement the `--deep-verify` flag for the `rmt-cli mirror` command, in the other PR we did checksum verifying for all cases, and @ikapelyukhin pointed it could make the command slower. I didn't add the flag into this PR to not make it bigger yet. I can remove and readd it once we add the flag.

spec/lib/rmt/deduplicator_spec.rb
ikapelyukhin: Yeah, so you are not being consistent.
Being consistent would mean referencing every single class starting from the root namespace -- this is just not a common development practice.
Besides, why would an application need it, but not the tests? Neither one needs it.

lib/rmt/file_validator.rb
ikapelyukhin: I can open a PR with proposed changes, but that's going to take me a while.

lib/rmt/file_validator.rb
ikapelyukhin: @lcaparroz I'd suggest actually finishing the feature in one PR.
This PR is one PR too many already, because I can't easily see the full diff (without merging both PRs into a separate branch and diffing locally).

app/models/downloaded_file.rb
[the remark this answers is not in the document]
lcaparroz: Removed it :heavy_check_mark:

lib/rmt/file_validator.rb
tmuntaner: @lcaparroz The dual responsibility of deleting invalid files while also finding a valid instance makes the method confusing. 

If we need to keep it that way, here a potential `.each` block example.

```ruby
def find_valid_file_by_checksum(checksum, checksum_type, deep_verify:)
  file_path = nil
  ::DownloadedFile.where(checksum: checksum, checksum_type: checksum_type).each do |file|
    file_path ||= file.local_path if validate_tracked_file_on_disk(file, deep_verify)
  end
  local
end
```

@ikapelyukhin Is something like this what you're looking for?

lib/rmt/file_validator.rb
lcaparroz: `is_orphaned_db_record?` doesn't sounds good either, because the method tells if a file found on DB is valid on disk. Anyway, I rewrote the public method `#find_valid_file_by_checksum` and removed this method, as well as the complicate method chaining.

lib/rmt/file_validator.rb
lcaparroz: Changed :heavy_check_mark:

lib/rmt/deduplicator.rb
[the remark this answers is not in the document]
lcaparroz: I've changed the files I touched in the last commit.

lib/rmt/file_validator.rb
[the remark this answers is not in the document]
lcaparroz: I removed the `.tap` method calls where it is not clear the benefits of using it. About `deep_verify`, I'll evaluate if it's worth implementing the `--deep-verify` flag in this PR.

lib/rmt/file_validator.rb
lcaparroz: Check the last commit, I rewrote the method and maybe it got better to understand it.

lib/rmt/file_validator.rb
lcaparroz: Done :heavy_check_mark:

lib/rmt/file_validator.rb
lcaparroz: I liked your suggestion, I was annoyed to pass `deep_verify` multiple levels down. Thanks :)

I implemented it in the last commit :heavy_check_mark:

app/models/downloaded_file.rb
lcaparroz: Removed it :heavy_check_mark:

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/125550/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 2c93aea1e16b5c17a13104035110199181317e49 into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/125551/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged a2e7677df5e04c2784a110e7919bddedb93e08be into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

spec/lib/rmt/deduplicator_spec.rb
lcaparroz: I removed the `::` operators from files I changed in this PR.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/125552/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 067a40e615738537452bfe81c223297f2c96fdcc into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/125587/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged da108a949475d66643a15064bc856ef024665c64 into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+        .group_by { |f| valid_on_disk?(f, deep_verify) }
+
+      # Remove invalid files/DB entries as soon as they are found
+      valid_files_on_disk[false]&.each do |file|
+        FileUtils.remove_file(file.local_path, force: true)
+        file.destroy
+      end
+
+      valid_files_on_disk[true] || []
+    end
```
ikapelyukhin, line 32:
```suggestion
    def find_valid_files_by_checksum(checksum, checksum_type, deep_verify:)
      files = DownloadedFile.where(checksum: checksum, checksum_type: checksum_type)

      files.delete_if do |file|
        next false if valid_on_disk?(f, deep_verify)

        FileUtils.remove_file(file.local_path, force: true)
        file.destroy
        true
      end
    end
```

lib/rmt/file_validator.rb
```diff
@@ -0,0 +1,48 @@
+require 'fileutils'
+
+class RMT::FileValidator
```
ikapelyukhin, line 3: Make it a mix-in and mix it into `RMT::Mirror`, and then make `deep_verify` an attribute of `RMT::Mirror` class? Just so that you don't have to pass `deep_verify` as a parameter everywhere.

Same probably could apply for `RMT::Deduplicator` and force copy/don't track options.

lib/rmt/file_validator.rb
lcaparroz: Nice shorter version! I just needed to add `.to_a` after the `.where(...)`, I guess `ActiveRecord::Relation` implements the `Enumerable` interface but not the `Array` one; `delete_if` is specific for arrays. I'm going to commit this change together with the remaining others.

lib/rmt/mirror.rb
lcaparroz: Done :heavy_check_mark:

lib/rmt/file_validator.rb
lcaparroz: Ok, I'll come up with a separated commit, maybe we can even delegate logging to `RMT::Deduplicator`, and eliminate the wrapper method `RMT::Mirror#deduplicate`, which only logs when a file is deduplicated.

lib/rmt/file_validator.rb
lcaparroz: I'll do a follow up PR after we merge this one and #585. Adding the flag together with these changes will make it even harder for review.

lib/rmt/mirror.rb
lcaparroz: I tried to be more explicit in bbbe807. What do you think?

lib/rmt/file_validator.rb
lcaparroz: I implemented the mixins in c417d92. I guess it fits your idea.

lib/rmt/file_validator.rb
lcaparroz: Done :heavy_check_mark:

lib/rmt/file_validator.rb
lcaparroz: I guess we have some consensus now, the final implementation for this method is [here](https://github.com/SUSE/rmt/pull/594/commits/c417d92637b04ebba6f88faddb985e90fdb7df60#diff-626359a9598a29f883f7227ee01a261bR21-R33).

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/126101/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged c417d92637b04ebba6f88faddb985e90fdb7df60 into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

db/migrate/20200723124836_add_uniqueness_to_downloaded_files_local_path.rb
```diff
@@ -0,0 +1,19 @@
+class AddUniquenessToDownloadedFilesLocalPath < ActiveRecord::Migration[5.2]
+  def change
+    # Remove duplicates before adding uniqueness to `local_path`
+    ActiveRecord::Base.connection.execute(
+      <<~SQL
+        DELETE FROM `downloaded_files`
```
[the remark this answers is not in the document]
lcaparroz: I tested it locally with a few scenarios and on another branch to be sure it would not break the pipeline, it works fine. Added your suggestion in 72bfe76.

lcaparroz:
Now, I guess I addressed everyone's concerns :) If I forgot something, please let me know.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/126110/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 72bfe763e54111ab9c900d2cc24afc134bd8d9bc into ea2b2116d04bd33b3855ab1d5dbac7410d708810)

PR #595, opened 2020-08-12
Add the ability to remove un-mirrored repositories

tmuntaner:
We did not give users the ability to remove unused repositories. To fix this, we created a command to delete repository directories which aren't marked to be mirrored.

### DIFF
--- a/lib/rmt/cli/repos.rb
+++ b/lib/rmt/cli/repos.rb
@@ -16,6 +16,49 @@ class RMT::CLI::Repos < RMT::CLI::Base
   end
   map 'ls' => :list
 
+  desc 'clean', _('Remove un-mirrored repositories from the local disk.')
+  def clean
+    base_directory = RMT::DEFAULT_MIRROR_DIR
+
+    repos_to_delete = Repository.where(mirroring_enabled: false).map do |repo|
+      repository_dir = File.join(base_directory, repo.local_path)
+      Dir.exist?(repository_dir) ? repo : nil
+    end.compact
+
+    if repos_to_delete.empty?
+      puts _('No un-mirrored repositories found on local disk.')
+      return
+    end
+
+    puts _('RMT found the following un-mirrored repositories:')
+    print "\n\e[31m"
+    repos_to_delete.each do |repo|
+      puts repo.description
+    end
+    print "\n\e[0m\e[1m"
+    print _('Would you like to continue and remove these directories from your local disk?')
+    print "\n\e[22m\s\s"
+    print _("Only '%{input}' will be accepted.") % { input: 'yes' }
+    print "\n\n\s\s\e[1m"
+    print _('Enter a value:')
+    print "\e[22m\s\s"
+
+    input = $stdin.gets.to_s.strip
+    if input != 'yes'
+      puts "\n" + _('Clean cancelled.')
+      return
+    end
+
+    repos_to_delete.each do |repo|
+      path = File.join(base_directory, repo.local_path)
+      FileUtils.rm_r(path, secure: true)
+    end
+
+    print "\n\e[32m"
+    print _('Clean finished.')
+    print "\e[0m\n"
+  end
+
   desc 'enable IDS', _('Enable mirroring of repositories by a list of repository IDs')
   long_desc <<-REPOS
 #{_('Enable mirroring of repositories by a list of repository IDs')}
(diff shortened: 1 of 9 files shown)


suse-tests-pass:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lib/rmt/cli/repos.rb
```diff
@@ -16,6 +16,49 @@ def list
...
+
+    input = $stdin.gets.to_s.strip
+    if input != 'yes'
+      puts "\n" + _('Clean cancelled.')
+      return
+    end
+
+    repos_to_delete.each do |repo|
+      path = File.join(base_directory, repo.local_path)
+      FileUtils.rm_r(path, secure: true)
```
lcaparroz, line 54: At this point, I think it would be nice to also untrack all the repositories' files in the `DownloadedFile` database.

I thought of something like:

```suggestion
      FileUtils.rm_r(path, secure: true)

      # Delete files which no longer exist in disk (keep data integrity)
      files = DownloadedFile.arel_table
      DownloadedFile.where(files[:local_path].matches("#{path}%")).destroy_all
```

We could also do something like

```ruby
DownloadedFile.where('local_path LIKE ?', "#{path}%").destroy_all
```

But this version would be less portable between different databases (though the project was built to work with MySQL/MariaDB).

lib/rmt/cli/repos.rb
```diff
@@ -16,6 +16,49 @@ def list
...
+    print "\n\e[0m\e[1m"
+    print _('Would you like to continue and remove these directories from your local disk?')
+    print "\n\e[22m\s\s"
+    print _("Only '%{input}' will be accepted.") % { input: 'yes' }
+    print "\n\n\s\s\e[1m"
+    print _('Enter a value:')
+    print "\e[22m\s\s"
+
+    input = $stdin.gets.to_s.strip
+    if input != 'yes'
```
lcaparroz, line 47: I'm not sure if we have any guideline for user input on CLI tools, but what about having an option for a shorter, case-insensitive answer like only `y`:

```suggestion
    unless input.match?(/y(es)?/i)
```

If it's worth, we'd need to change line 41 above to include the shorter option.

On the other hand, accepting only `yes` is safer to the customer, two more keys to press before hitting enter could avoid accidentally accepting the clean command; `y` can lead to "automatic" behavior.

lib/rmt/cli/repos.rb
lcaparroz: What about adding some logging about which repositories were successfully cleaned? Maybe inside the `.each` loop, after completely cleaning up.

lib/rmt/cli/repos.rb
tmuntaner: I would probably just opt for the `yes` only option, as it is safer and not much more effort on the user's end.

lib/rmt/cli/repos.rb
tmuntaner: @lcaparroz I can't believe that I forgot to add the removing from downloaded files :(

It's also a very good idea to have log messages during the deletion process.

tmuntaner: @lcaparroz I added a commit with your feedback.

lib/rmt/cli/repos.rb
```diff
@@ -16,6 +16,52 @@ def list
...
+      puts "\n" + _('Clean cancelled.')
+      return
+    end
+
+    print "\n"
+    repos_to_delete.each do |repo|
+      path = File.join(base_directory, repo.local_path)
+      FileUtils.rm_r(path, secure: true)
+      DownloadedFile.where('local_path LIKE ?', "#{path}%").destroy_all
+      puts _("Deleted repository '%{repo}'.") % { repo: repo.description }
```
lcaparroz, line 57: I'm not sure about the "*deleted*" word here: we delete the repository files and directories, but, if I correctly understood, the repository would still be listed with `rmt-cli repos list`, as "Not mirrored".

In my previous review I wondered about the need to also remove repositories from the database. If this is the case, then probably the command `rmt-cli repos clean` wouldn't make so much sense. Just to make sure we agree on the meaning of **clean**:

1. It will remove all repository files from disk but still keeps the repository as disabled in the database, or;
1. In addition to files on disk, it also removes the repository entry from the database.

If we just want to clean files, maybe we could change to something like this:

```suggestion
      puts _("Deleted files from repository '%{repo}'.") % { repo: repo.description }
```

Or something that better states RMT is just removing files.

Currently, I guess `rmt-cli` don't have an option to remove a repository from the list (database).

tmuntaner: @lcaparroz You are totally right. We should never delete repositories from the database, only delete their local files.

PR #739, opened 2021-06-23
Cache janitor

### LINKED ISSUE #738: Cache files grow indefinitely, despite 20 min expiration in code.
In Production, Rails is defaulting to file-based caching; with file caching the cache files never 'expire'... they must be removed (this is documented behavior [1]).

Eventually, `/var/lib/rmt/tmp/cache` will outgrow any disk.

bear454:
## Description

"Are you gonna finish that?" - School janitors ;-)

Fixes #738 

## Change Type

## Checklist

- [x] RMT's test coverage remains at 100%.

## Other Notes

Clean up after Rails' messy file cache scheme, by enforcing the expiration time set in code. 

I didn't add a Changlog entry, as I didn't think this would get a version bump by itself, but I'm happy to add one upon request.


felixsch:
Hi thanks for pointing out the issue and providing a solution to it. I will ask my colleges to also check this pull request to get more opinion on what is the possible best way to solve this. From my point of view this looks like a easy and good solution since the files actually never deleted by any `rails` process (as pointed out by you already).

For reference: https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-filestore

jesusbv (approved):
LGTM

bear454:
> Would it make sense to switch to memory based caching (e.g. `ActiveSupport::Cache::MemoryStore`)?

MemoryStore isn't an adequate solution, as each instance of RMT ends up with a separate cache - in the default _puma_ configuration, a single RMT server runs in 10 threads (2 workers, 5 threads each) so there are 10 different MemoryStores, and no way to pin a session to a specific thread... so the cache would become useless.

(In a larger HA production deployment, using a shared Redis or Memcached backend would probably be the most performant solution, but that is a pretty big change from what's done now; in the end it made more sense to provide a fix for the current problem than to drive forward with a different solution.)

digitaltom:
As far as I can see, the Rails cache is only used by the pubcloud engines, the cache directory on our reference RMT is completely empty. What do you think of only including the cache janitor service in the pubcloud package? 

Regarding the cache method (file vs memory), I think it would make sense to add a comment to `config/environments/production.rb` that the publoud engines rely on a shared cache per host, else we might break this assumption unintentionally in the future.

felixsch:
I agree with this. I checked our instances and they do not have any cache used. Speaking with the pub cloud team showed exactly what Tom is saying. 

> As far as I can see, the Rails cache is only used by the pubcloud engines, the cache directory on our reference RMT is completely empty. What do you think of only including the cache janitor service in the pubcloud package?

This sounds like a good idea to me.

> Regarding the cache method (file vs memory), I think it would make sense to add a comment to `config/environments/production.rb` that the publoud engines rely on a shared cache per host, else we might break this assumption unintentionally in the future.

Also this sounds like a good idea to make sure nobody is breaking this feature in the future.

thutterer:
Could this maybe even be solved without any new custom systemd service, if we just changed the FileStore location to a real _tmp_ folder like `/var/tmp/rmt` where the OS could take care of not growing it too large? We would need to check how SLES/openSUSE handles this, especially when you never reboot, of course.

felixsch:
> Could this maybe even be solved without any new custom systemd service, if we just changed the FileStore location to a real _tmp_ folder like `/var/tmp/rmt` where the OS could take care of not growing it too large? We would need to check how SLES/openSUSE handles this, especially when you never reboot, of course.

This is also an option. Systemd service files come with `CacheDirectory` directive to make sure certain directory is created if not existing (https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=)

But this relies on the system being rebooted? Would this also be an option?

bear454:
Are the reference RMT servers using a different configuration that the default for this repo?

According to https://github.com/SUSE/rmt/blob/master/config/environments/production.rb#L50-L51 no cache store is specified for production, and FileStore is the default if none is defined.

digitaltom:
> Are the reference RMT servers using a different configuration that the default for this repo?
> 
> According to https://github.com/SUSE/rmt/blob/master/config/environments/production.rb#L50-L51 no cache store is specified for production, and FileStore is the default if none is defined.

We're running default config on rmt.scc.suse.de. I think the reason that the cache directory is empty is that there is no usage of `Rails.cache` in the standard RMT code. It's only used in the publoud engines which are shipped as a subpackage and only are installed in the public cloud use case.

bear454:
:facepalm:

bear454:
I'll try to move this to the pubcloud package.

bear454:
`Rails.cache` is called in both the `instance_verification` and `zypper_auth` engines... So should this be included in both of those subpackages?

digitaltom:
> `Rails.cache` is called in both the `instance_verification` and `zypper_auth` engines... So should this be included in both of those subpackages?

All of the engines are in the `pubcloud` subpackage: https://github.com/SUSE/rmt/blob/master/package/obs/rmt-server.spec#L272

bear454:
Rebased onto master

bear454:
Any idea why I'm seeing test failures here? Adding some additional systemd stuff shouldn't fail Rails integration tests...

digitaltom:
@bear454 we just migrated to github actions and are still fixing some setups, see https://github.com/SUSE/rmt/pull/746, https://github.com/SUSE/rmt/pull/749, https://github.com/SUSE/rmt/pull/750

ngetahun:
@bear454 Please rebase your branch.

bear454:
Can this be merged @digitaltom ?

digitaltom:
@bear454 good to merge from our side.

bear454:
Will this PR, or the resolved issue be updated when this is included in a release?

digitaltom:
> Will this PR, or the resolved issue be updated when this is included in a release?

No, we don't have a mechanism to notify in gh prs on release. It would be useful to have a changelog for the pubcloud package, so you can see which version has the update.

PR #1001, opened 2023-05-18
Handle missing X-Original-URI header

jesusbv:
## Description

In SLE-Micro 5.4 instance in GCE fails to update running the command `transactional-update dup`

One of the issues is header not being present, which IMHO is not the roo/core issue but it needs fixing

When the header is not present, it produces a 500 error due to `undefined include? method for nil class`

This is part of the fix for bsc#1211398

## Change Type

## Checklist

## Other Notes

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -172,7 +172,7 @@ module SccProxy
       active_products_ids = scc_systems_activations.map { |act| act['service']['product']['id'] if act['status'].casecmp('active').zero? }.flatten
       products = Product.where(id: active_products_ids)
       product_paths = products.map { |prod| prod.repositories.pluck(:local_path) }.flatten
-      active_subscription = product_paths.any? { |path| headers['X-Original-URI'].include?(path) }
+      active_subscription = product_paths.any? { |path| headers['X-Original-URI'].to_s.include?(path) }
       if active_subscription
         { is_active: true }
       else
@@ -186,7 +186,7 @@ module SccProxy
         end
         products = Product.where(id: expired_products_ids)
         product_paths = products.map { |prod| prod.repositories.pluck(:local_path) }.flatten
-        expired_subscription = product_paths.any? { |path| headers['X-Original-URI'].include?(path) }
+        expired_subscription = product_paths.any? { |path| headers['X-Original-URI'].to_s.include?(path) }
         if expired_subscription
           {
             is_active: false,


suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/313062/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 9111621deea6b42267e721e4bb1f260a30a49bd0 into fc9f05171157bb0fa06282812ee3870baa9acacd)

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -172,7 +172,7 @@ def scc_check_subscription_expiration(headers, login, system_token, logger)
       active_products_ids = scc_systems_activations.map { |act| act['service']['product']['id'] if act['status'].casecmp('active').zero? }.flatten
       products = Product.where(id: active_products_ids)
       product_paths = products.map { |prod| prod.repositories.pluck(:local_path) }.flatten
-      active_subscription = product_paths.any? { |path| headers['X-Original-URI'].include?(path) }
+      active_subscription = product_paths.any? { |path| headers['X-Original-URI'].to_s.include?(path) }
```
digitaltom, line 175: The more beautiful way to solve this would be: 
```suggestion
      active_subscription = product_paths.any? { |path| headers.fetch('X-Original-URI', '').include?(path) }
```

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -186,7 +186,7 @@ def scc_check_subscription_expiration(headers, login, system_token, logger)
         end
         products = Product.where(id: expired_products_ids)
         product_paths = products.map { |prod| prod.repositories.pluck(:local_path) }.flatten
-        expired_subscription = product_paths.any? { |path| headers['X-Original-URI'].include?(path) }
+        expired_subscription = product_paths.any? { |path| headers['X-Original-URI'].to_s.include?(path) }
```
digitaltom, line 189:
```suggestion
        expired_subscription = product_paths.any? { |path| headers.fetch('X-Original-URI', '').include?(path) }
```

digitaltom (approved):
See suggestion, LGTM

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: Will update the code with the suggestion

Does it mean we should adapt that kind of checks with the suggestion ?
like [here](https://github.com/SUSE/rmt/blob/fc9f05171157bb0fa06282812ee3870baa9acacd/engines/zypper_auth/lib/zypper_auth/engine.rb#L10) ?
I followed the same style/pattern, happy to update it in the rest of the code 
I understand the code I shared is different and works as it is, I ask as a matter of same style
```ruby
# change
instance_data = Base64.decode64(request.headers['X-Instance-Data'].to_s)
# to
instance_data = Base64.decode64(request.headers.fetch('X-Instance-Data', ''))
```

digitaltom: It's better style to define the default value explicitly and not rely on `nil.to_s`. But imho no need to change the whole codebase, just saw that in your new code.

suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/313247/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 49b13c1f0a874a942a9bdb8a1ce5ac4ac18823ac into fc9f05171157bb0fa06282812ee3870baa9acacd)

PR #1062, opened 2024-01-03
Fix secrets file access mode

ngetahun:
## Description
Set group readable permissions to rails credentials file.

## Change Type

*Please select the correct option.*

- [x] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [ ] I have verified that my code follows RMT's coding standards with `rubocop`.
- [ ] I have reviewed my own code and believe that it's ready for an external review.
- [ ] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [ ] RMT's test coverage remains at 100%.
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Other Notes

Please use this space to provide notes or thoughts to the team, such as tips on how to review/demo your changes.

PR #1064, opened 2024-01-04
Fix secrets file access mode - for real!

felixsch:
## Description

The problem is that the file permission where originally wrong and breaks RMT completely because it can not read the secrets. This fixes the permissions for both files without relying on handling this in the spec file where it should **not** be handled anyway.

image

original:
```
-rwxr-xr-x 1 root root  1277 Dec 12 05:42 /usr/share/rmt/config/secrets.yml
-rw-r--r-- 1 root nginx  280 Jan  2 13:52 /usr/share/rmt/config/secrets.yml.enc
-rw------- 1 root nginx   32 Jan  2 13:52 /usr/share/rmt/config/secrets.yml.key
```
but we need due to hardening:

```
-rwxr-xr-x 1 root root  1277 Dec 12 05:42 /usr/share/rmt/config/secrets.yml
-rw-r----- 1 root nginx  280 Jan  2 13:52 /usr/share/rmt/config/secrets.yml.enc
-rw-r----- 1 root nginx   32 Jan  2 13:52 /usr/share/rmt/config/secrets.yml.key
```
see: https://bugzilla.suse.com/show_bug.cgi?id=1215176

Fixes file permission for `config/secrets.yml.key` and `config/secrets.yml.enc` to `0640` during fresh install.

**part of**: https://trello.com/c/hL2IyVT1/3061-fix-rmt-server-install-creates-secretsymlkey-with-root-owner
**fixed version of**: https://github.com/SUSE/rmt/pull/1062

**How to test this pull request:**

```bash
$ docker run --rm -it registry.suse.com/suse/sle15:15.4
> zypper ar https://download.opensuse.org/repositories/systemsmanagement:/SCC:/RMT/SLE_15_SP4/systemsmanagement:SCC:RMT.repo
> zypper refresh
> zypper in rmt-server
> ls -la /usr/share/rmt/config
# expect: File permissions for .enc and .key are 0640
```

**Thank you for reviewing this pull request** :rocket:

### DIFF
--- a/lib/tasks/encrypted_key.rake
+++ b/lib/tasks/encrypted_key.rake
@@ -6,6 +6,8 @@ namespace :rmt do
 
       Rails::Generators::EncryptionKeyFileGenerator
         .new.add_key_file('config/secrets.yml.key')
+
+      FileUtils.chmod(0o640, 'config/secrets.yml.key')
     end
 
     desc 'Create the `secret_key_base` for Rails'
@@ -13,6 +15,8 @@ namespace :rmt do
       Rails::Secrets.write(
         { 'production' => { 'secret_key_base' => SecureRandom.hex(64) } }.to_yaml
       )
+
+      FileUtils.chmod(0o640, 'config/secrets.yml.enc')
     end
   end
 end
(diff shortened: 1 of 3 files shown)


suse-tests-pass:
https://ci.suse.de/job/scc-RMT-integration-tests/351568/console
If the given link has expired,you can force a Prophet rerun by just deleting this comment. (Merged 6b0381b6e8f38aedef5e65ed7a3fcba73e9ae9ba into 40dd490a960f56cb2b641696bf5495eace3a8a7d)

digitaltom (approved):
Afaiks this will only fix fresh installations. Systems that already have the secret files with wrong permissions will not get fixed by this patch.

PR #1087, opened 2024-02-08
Add missing build require for sqlite; the real pre-release

felixsch:
## Description

sqlite is missing as build requirement.

**How to review this pull request:**

- It builds in obs!

### DIFF
--- a/package/obs/rmt-server.spec
+++ b/package/obs/rmt-server.spec
@@ -49,6 +49,7 @@ BuildRequires:  gcc
 BuildRequires:  libcurl-devel
 BuildRequires:  libffi-devel
 BuildRequires:  libmysqlclient-devel
+BuildRequires:  sqlite-devel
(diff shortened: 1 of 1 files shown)


josegomezr (approved):
LGTM

lib/rmt.rb
```diff
@@ -1,5 +1,5 @@
 module RMT
-  VERSION ||= '2.15'.freeze
+  VERSION ||= '2.15-pre'.freeze
```
ngetahun, line 2: Why are you using the `-pre` version specifier?

felixsch: I was testing something and pushed it accidentally. I removed it!

josegomezr: Good catch, I thought that with our shiny fast CI we were doing pre-releases ;)

ngetahun (approved):
LGTM!

PR #1089, opened 2024-02-12
Add user provides to .spec file (contrib by dimstar@opensuse.org)

digitaltom:
## Description

Fix build for Factory by adding user provides.


felixsch (approved):
lgtm :+1:

Thanks for picking it up!

PR #1094, opened 2024-02-19
Add registry support

jesusbv:
## Description

Add registry support to RMT

## Change Type

## Checklist

- [ ] RMT's test coverage remains at 100%.

## Other Notes


engines/registry/README.md
```diff
@@ -0,0 +1,3 @@
+# Registry
+Instances in the cloud have the option to use registry.suse.com for their containers,
+this engine provides the endpoint to auth and handling of that access to the registry
```
rjschwei, line 3: I don't get what this is trying to express and why registry.suse.com comes into play here. I suggest:
"""
Provide an authentication end point for a container registry. This supports access control based on system credentials to registry paths that are set to have access restrictions.
"""

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -0,0 +1,92 @@
+module Registry
+  class RegistryController < ::ApplicationController
+    REGISTRY_SERVICE = 'SUSE Linux Docker Registry'.freeze
```
rjschwei, line 3: s/Docker/OCI/

rjschwei (changes requested):

config/application.rb
```diff
@@ -58,6 +58,15 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('app', 'validators')
 
+    # registry config needed
+    config.autoloader = :classic
+    registry_cert = "-----BEGIN PRIVATE KEY-----
```
bear454, line 63: Can we use the RMT server's private key here? The we can have the registry use the rmt-ca.crt for verification.

config/application.rb
```diff
@@ -58,6 +58,15 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('app', 'validators')
 
+    # registry config needed
+    config.autoloader = :classic
+    config.registry_private_key = OpenSSL::PKey::RSA.new(
+      File.read('/etc/rmt/ssl/rmt-server.key')
+      )
+    config.registry_public_key = config.registry_private_key.public_key
+    config.access_policies = '/etc/rmt/access_policies.yml'
```
jesusbv, line 67: The other option is to add the path to `/etc/rmt.conf`

config/application.rb
```diff
@@ -58,6 +58,15 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('app', 'validators')
 
+    # registry config needed
+    config.autoloader = :classic
+    config.registry_private_key = OpenSSL::PKey::RSA.new(
+      File.read('/etc/rmt/ssl/rmt-server.key')
+      )
+    config.registry_public_key = config.registry_private_key.public_key
```
rjschwei, line 66: What are we doing with the key?

bear454: The OAuth token must be signed with a key, for which a certificate is available to the registry server, with the RMT server's FQDN as a SAN.

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    def basic_auth
+      # skip authentication if this is not a login request
+      return unless request.authorization
+
+      authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
+        begin
+          @client = Registry::AuthenticatedClient.new(login, password)
+        rescue StandardError
+          logger.info _('Could not find system with login \"%{login}\" and password \"%{password}\"') %
+            { login: login, password: password }
```
rjschwei, line 65: his message does not help the user, it only tells the user that they are SOL. We need to tell the user what to do. In the special case we are currently implementing, i.e. for the public cloud update infrastructure only this should read.
"Registration expired please run cloudregistryauth command".  For air-gaped environments auth probably doesn't matter and our specialized message may be sufficient, but that requires more thought later. For now let's make sure we tell the user what they need to do.

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+        true
+      end
+    end
+
+    def catalog_token_auth
+      authenticate_or_request_with_http_token(authorize_api_registry_url, 'authentication required') do |token|
+        begin
+          @client = Registry::CatalogClient.new(token)
+        rescue JWT::DecodeError
+          respond_with_error(message: 'Invalid token', status: :unauthorized) and return
```
rjschwei, line 80: We cannot simply tell the user that the token is invalid.

engines/registry/app/models/access_token.rb
```diff
@@ -0,0 +1,46 @@
+require 'base32'
+
+# Following Docker distribution token auth specs,
+# see docs here: https://github.com/distribution/distribution/blob/main/docs/spec/auth/token.md
+class Registry::AccessToken
+  def initialize(account, service, granted_scopes)
+    @account = account
+    @service = service # "SUSE Linux Docker Registry"
```
rjschwei, line 8: s/Docker/OCI/

engines/registry/app/controllers/registry/registry_controller.rb
digitaltom: I'm don't think podman will show custom messages to the user at all. If I recall correctly it's showing it's own messages depending on the reponse code.

jesusbv: It does, it would show 
```bash
> podman search registry-ec2.susecloud.net/
ERRO[0001] error getting search results from v2 endpoint "registry-ec2.susecloud.net": requested access to the resource is denied 
Error: 1 error occurred:
	* couldn't search registry "registry-ec2.susecloud.net": requested access to the resource is denied
```
that log information is for internal use, not for the user, happy to change the log message though

engines/registry/app/models/access_token.rb
```diff
@@ -0,0 +1,46 @@
+require 'base32'
+
+# Following Docker distribution token auth specs,
+# see docs here: https://github.com/distribution/distribution/blob/main/docs/spec/auth/token.md
+class Registry::AccessToken
+  def initialize(account, service, granted_scopes)
+    @account = account
+    @service = service # "SUSE Linux OCI Registry"
```
jesusbv, line 8: Implies a change on the registry, as it currently has `Docker`

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+      # skip authentication if this is not a login request
+      return unless request.authorization
+
+      authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
+        begin
+          @client = Registry::AuthenticatedClient.new(login, password)
+        rescue StandardError
+          logger.info _('Could not find system with login \"%{login}\" and password \"%{password}\"') %
+            { login: login, password: password }
+          error = ActionController::TranslatedError.new(N_('Invalid registry credentials'))
```
rjschwei, line 64: I repeat we need to tell the user what to do. A generic message like "Invalid registry credentials" will do nothing but trigger the creation of support requests. We need to make at least an attempt to get the "you need to re-authenticate" message (different wording required) to the client. Making podman display it may be a follow up project in upstream podman if it doesn't display custom messages.

rjschwei (changes requested):

engines/registry/app/controllers/registry/registry_controller.rb
jesusbv: @rjschwei what would `cloudregistryauth` command be ? 
We get the credentials when registering the instance on the client, 
I understand the command would be `registercloudguest` but, do we want the user
to re register the system to use (re-auth on) the registry ?

engines/registry/app/controllers/registry/registry_controller.rb
cwh42: We are replicating a OCI compliant registry API here and this is the expected response in case the token validation fails.
The client (docker/podman) will just get a new token, then. Usually no user action required.

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+          error.status = :unauthorized
+          raise error
+        end
+
+        @client.authorized_for_catalog?
+      end
+    end
+
+    # is called by authenticate_or_request_with_http_token when client provides no token
+    def request_http_token_authentication(realm = authorize_api_registry_url, message = 'authentication required')
```
cwh42, line 89: Please have a look at the most recent version of [glue/app/controllers/api/registry_controller.rb](https://github.com/SUSE/happy-customer/blob/3f818723ea1b8cceab28e01505ea13d0570e28e1/glue/app/controllers/api/registry_controller.rb#L122
)
We had to add a fix here that broke some registry clients.

engines/registry/app/models/access_scope.rb
```diff
@@ -0,0 +1,92 @@
...
+  def allowed_paths(system = nil)
+    repo_list = RegistryCatalogService.new.repos(reload: false, system: system)
+    access_policies_yml = YAML.safe_load(
+      File.read(Rails.application.config.access_policies)
+    )
+    active_products = system.activations.includes(:product).pluck(:product_class)
+
+    allowed_products = (active_products & access_policies_yml.keys) + ['free']
+    allowed_glob_paths = access_policies_yml.values_at(*allowed_products).flatten
+    @allowed_paths = repo_list.select { |repo| repo if allowed_glob_paths.any? { |glob_path| File.fnmatch(glob_path, repo) } }
```
cwh42, line 90: `File.fnmatch()` does not handle * and ** the same way as we defined it in SCC. So this does not give the right results!
For example `File.fnmatch('*')` matches *all* repos in the registry and not only those in the "root" directory as SCC does.

engines/registry/app/models/access_token.rb
cwh42: Not sure if any of the clients does care about this.

rjschwei: Docker is a TM and a couple of years ago we had an effort to remove the word "Docker" from everything that was not referring directly to the command itself.

engines/registry/app/controllers/registry/registry_controller.rb
rjschwei: There will be no new token without re-authentication, i.e. we want/need to verify that the instance from which the request is coming from is still the instance that triggered the token generation when the system was registered. When the token expires, i.e. when we remove the system from the auth cache, then the token is expired until a re-authentication process happens. Since neither podman nor docker can send us the necessary data, see https://github.com/containers/podman/discussions/21697, we have to work around the limitation of those tools with our own code. Per agreement today that is going to be `cloudguestregistryauth`

rjschwei: That said if the standard only allows for this response then that is the response we have to give and we have to figure out how to reduce the number of tickets generated via some other means

engines/registry/app/models/access_scope.rb
jesusbv: mmmm, I had run some tests and the results seem OK ?

```
> podman search registry-ec2.susecloud.net/bci
NAME                                                 DESCRIPTION
registry-ec2.susecloud.net/opensuse/bci/bci-busybox  
registry-ec2.susecloud.net/opensuse/bci/ruby         
> podman search registry-ec2.susecloud.net/tool
NAME                                                                                                         DESCRIPTION
registry-ec2.susecloud.net/opensuse/leap/micro/5.2/images/totest/containers/opensuse/leap-micro/5.2/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.3/images/totest/containers/opensuse/leap-micro/5.3/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.4/images/totest/containers/opensuse/leap-micro/5.4/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.5/images/images/opensuse/leap-micro/5.5/toolbox             
registry-ec2.susecloud.net/opensuse/leap/micro/5.5/images/totest/containers/opensuse/leap-micro/5.5/toolbox  
```
and no term/pattern returns _all_ free repos 
```
> podman search registry-ec2.susecloud.net/
NAME                                                                                                         DESCRIPTION
registry-ec2.susecloud.net/home/marcus.schaefer/delta_containers/containers_tw/fleet                         
registry-ec2.susecloud.net/home/marcus.schaefer/delta_containers/containers_tw/joe                           
registry-ec2.susecloud.net/opensuse/bci/bci-busybox                                                          
registry-ec2.susecloud.net/opensuse/bci/ruby                                                                 
registry-ec2.susecloud.net/opensuse/leap/micro/5.2/images/totest/containers/opensuse/leap-micro/5.2/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.3/images/totest/containers/opensuse/leap-micro/5.3/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.4/images/totest/containers/opensuse/leap-micro/5.4/toolbox  
registry-ec2.susecloud.net/opensuse/leap/micro/5.5/images/images/opensuse/leap-micro/5.5/toolbox             
registry-ec2.susecloud.net/opensuse/leap/micro/5.5/images/totest/containers/opensuse/leap-micro/5.5/toolbox  
> 
```

are those results not OK ?

digitaltom: What @cwh42 meant is the meaning of * and ** in access policies: 

'*' applies to all images in the current path. For example, the policy `suse/*` applies to all images in the `suse/` directory, like `suse/389-ds`. But not to `suse/ltss/sle15.3/bci-base` for example. 

image

'**' applies to all images and images in subdirectories of the path. Example for `suse/sles/**`: 

image

jesusbv: Updated

engines/registry/app/services/registry_catalog_service.rb
```diff
@@ -0,0 +1,68 @@
...
+  def fetch_registry_repos(system)
+    Rails.logger.info('Fetch registry repos')
+    response = catalog_token(system)
+    catalog_auth_token = JSON.parse(response.body).fetch('token', '')
+    response = all_repos(catalog_auth_token)
+    JSON.parse(response.body).fetch('repositories', [])
+  end
+
+  def catalog_token(system)
+    system = System.where(proxy_byos: false).last if system.nil?
```
rjschwei, line 39: Why do we not allow systems that use the update server as a proxy? We want to allow any system that can register to the update server to access the registry if that is "allowed" according to what is included in a product.

package/files/nginx-pubcloud/nginx-registry-https.conf
```diff
@@ -0,0 +1,138 @@
...
+## since nginx is auth-ing before proxying.
+map $upstream_http_docker_distribution_api_version $docker_distribution_api_version {
+  '' 'registry/2.0';
+}
+
+
+server {
+    listen 443 default_server  ssl;
+    listen [::]:443 default_server ssl;
+    server_name rmt;
```
rjschwei, line 22: But rmt, i.e. access to the repositories is covered in the existing config file `rmt-server-pubcloud-https.conf` why do we have to repeat the settings here?

package/files/nginx-pubcloud/nginx-registry-https.conf
```diff
@@ -0,0 +1,138 @@
...
+    # smt.crt is used by cloud-regionsrv-client < 8.1.1
+    location /smt.crt {
+        alias /etc/rmt/ssl/rmt-ca.crt;
+    }
+}
+
+server {
+    listen 443   ssl;
+    listen [::]:443 ssl;
+    server_name ~^registry;
```
rjschwei, line 97: Do we not need an auth setup for the registry path?

rjschwei (changes requested):

engines/registry/app/services/registry_catalog_service.rb
jesusbv: Good catch

That check does not ban access to BYOS system, we do allow them access but, for the catalog/search request, during testing, there was a request made without system, system being nil, and it crashed

That scenario is fixed (login, pull and search operations succeed) and the check should be removed

bear454:
@jesusbv we looked at two different approaches for the nginx config, but the current state reflects half of each.

* One https config for both rmt & registry, with the search endpoint of registry assigned as a location to rmt;
* Two separate configs, with the search endpoint in the registry config redirecting to rmt.

Which way did you intend to go?

engines/registry/app/services/registry_catalog_service.rb
```diff
@@ -36,8 +36,8 @@ def fetch_registry_repos(system)
   end
 
   def catalog_token(system)
-    system = System.where(proxy_byos: false).last if system.nil?
-    uri = URI.parse(AUTH_URL)
+    # rmt.conf must have the root_url
```
bear454, line 39: Can we use the Request object to introspect the FQDN?

jesusbv:
> @jesusbv we looked at two different approaches for the nginx config, but the current state reflects half of each.
> 
>     * One https config for both rmt & registry, with the search endpoint of registry assigned as a location to rmt;
> 
>     * Two separate configs, with the search endpoint in the registry config redirecting to rmt.
> 
> 
> Which way did you intend to go?

The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that

Do you think one file is better ?

bear454:
> The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that
> 
> Do you think one file is better ?

I think it's fine as long as it's been proven to work with both `podman` and `docker`.

rjschwei:
> > The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that
> > Do you think one file is better ?
> 
> I think it's fine as long as it's been proven to work with both `podman` and `docker`.

Separate config is preferable, that allows a 3rd party that doesn't want to run the registry to not install the config and then the paths are not available.

bear454:
> > The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that
> > Do you think one file is better ?
> 
> I think it's fine as long as it's been proven to work with both `podman` and `docker`.

I'm withdrawing my objection here - as `docker` is incapable of searching via an authenticated endpoint, and only uses the unauthenticated /v1/search method. So, search of our registry will only be capable via `podman`.

bear454:
> > > The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that
> > > Do you think one file is better ?
> > 
> > 
> > I think it's fine as long as it's been proven to work with both `podman` and `docker`.
> 
> Separate config is preferable, that allows a 3rd party that doesn't want to run the registry to not install the config and then the paths are not available.

Well, the problem is around search. Because we want to emulate the behavior of SCC, where the `/v2/_catalog` API endpoint is overriden, and served by _rmt_ instead of _distribution_, there is an intimate linking. The two-config solution requires that the registry config introduce a `302 REDIRECT` on that API endpoint, which also means the _rmt_ FQDN needs to be specified in the config.

With one config file, the `v2/_catalog`  API endpoint is overridden not to a `302 REDIRECT` but directly served by the defined _rmt_ application, as part of the reverse proxy configuration. The client doesn't see a redirect (that's not defined in the registry API) and get served the direct results on the original request. But yes, both _rmt_ and _registry_ vhosts have to be defined in the same file for that to work.

package/files/nginx-pubcloud/nginx-registry-https.conf
bear454: No, the auth is defined in the registry config.

rjschwei:
> > > > The latter, split config, I think having the two configs separate is better, I updated the PR now to reflect that
> > > > Do you think one file is better ?
> > > 
> > > 
> > > I think it's fine as long as it's been proven to work with both `podman` and `docker`.
> > 
> > 
> > Separate config is preferable, that allows a 3rd party that doesn't want to run the registry to not install the config and then the paths are not available.
> 
> Well, the problem is around search. Because we want to emulate the behavior of SCC, where the `/v2/_catalog` API endpoint is overriden, and served by _rmt_ instead of _distribution_, there is an intimate linking. The two-config solution requires that the registry config introduce a `302 REDIRECT` on that API endpoint, which also means the _rmt_ FQDN needs to be specified in the config.
> 
> With one config file, the `v2/_catalog` API endpoint is overridden not to a `302 REDIRECT` but directly served by the defined _rmt_ application, as part of the reverse proxy configuration. The client doesn't see a redirect (that's not defined in the registry API) and get served the direct results on the original request. But yes, both _rmt_ and _registry_ vhosts have to be defined in the same file for that to work.

OK, thanks for the the details, then we are stuck with one config. But thta also means we should have 1 config, i.e. the existing config file should change rather than adding a new one.

package/files/nginx-pubcloud/nginx-registry-https.conf
```diff
@@ -0,0 +1,59 @@
...
+      proxy_set_header  X-Real-IP         $remote_addr; # pass on real client's IP
+      proxy_set_header  X-Forwarded-For   $proxy_add_x_forwarded_for;
+      proxy_set_header  X-Forwarded-Proto $scheme;
+      proxy_read_timeout                  900;
+    }
+
+    # override registry search
+    location /v2/_catalog {
+      # redirect to the catalog endpoint
+      rewrite /v2/_catalog https://smt-ec2.susecloud.net/v2/_catalog redirect;
```
bear454, line 57: So, I woke up overnight and realized - this line is why we'll need to do the one-config approach. We don't want every RMT server in the wild redirecting to `smt-ec2.susecloud.net`... and there's no way to do the redirect without an FQDN.

jesusbv: yes, hard coded string had to be fixed

config/application.rb
```diff
@@ -58,6 +58,19 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('app', 'validators')
 
+    # :nocov:
+    if Rails.env.production?
```
felixsch, line 62: I believe this needs to be gated behind a check if its actually running in public cloud mode, otherwise it will interfere with regular customers setups

jesusbv: Added

rjschwei (approved):

rjschwei:
What's missing, I think, is the cache handling. We had agreed that clients that access the registry have a longer TTL than access to the repositories. Anyway, that change should come in a separate PR. This PR is big enough as it is and we have not made the desired timely progress in getting it merged.

bear454:
> What's missing, I think, is the cache handling. We had agreed that clients that access the registry have a longer TTL than access to the repositories. Anyway, that change should come in a separate PR. This PR is big enough as it is and we have not made the desired timely progress in getting it merged.

Yes, agreed, this needs to be in a separate PR.

bear454 (approved):
There are still some magic strings in the code, but we can work on moving them out in subsequent PRs. That this is working, has a reliable config, and test coverage is good enough. Let's get rolling.

jesusbv:
> > What's missing, I think, is the cache handling. We had agreed that clients that access the registry have a longer TTL than access to the repositories. Anyway, that change should come in a separate PR. This PR is big enough as it is and we have not made the desired timely progress in getting it merged.
> 
> Yes, agreed, this needs to be in a separate PR.

Agree, yes, cache is not on this PR, and it should go on a different PR

PR #1124, opened 2024-04-18
Expire registry tokens in 8 hours, or as configured

bear454:
## Description

Registry engine creates JWT tokens for authz/n to an affiliated container registry. The tokens should expire in 8 hours, unless configured otherwise. Issuance is condition on the system having been 'last_seen' within the token expiration period.

## Change Type

## Checklist

- [x] I have documented the `MANUAL.md` file with any changes to the user experience.
- [x] RMT's test coverage remains at 100%.


MANUAL.md
```diff
@@ -297,6 +297,16 @@ The `web_server` section lets you tune the performance of your RMT server.
...
+
+**Container registry settings**
+
+The `registry` setting lets you adjust behavior of the container registry integration.
+
+  * `token_expiration` setting:
+    Specifies the timeout, from issuance, of the authentication/authorization token used by container clients (e.g. podman, docker).
+    Default value is 8 hours.
+    Acceptable values: Integer greater than 1, in seconds.
+
```
digitaltom, line 309: As the registry auth is a functionality only available from the public cloud engine and not really supported/available for customers with standard RMT package, I'd suggest to move this to the engine's documentation.

config/rmt.yml
```diff
@@ -47,3 +47,6 @@ web_server:
 
 connect_api:
   disable_system_tokens: false
+
+registry:
+  token_expiration: 28800 # 8.hours.to_i
```
digitaltom, line 52: Could we hardcode the default of 8h in the engine code, and only overwrite when this config value is present? Then we would not need to add this default to the config. (Same reason as above, I think this might confuse users that are using RMT without pubcloud engines (99.9%)).

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -16,7 +16,7 @@ def initialize(login, password)
 
   def authenticate_by_system_credentials(login, password)
     @systems = System.get_by_credentials(login, password)
-    if @systems.present?
+    if @systems.any? { |system| system.last_seen_at > Settings[:registry].token_expiration.seconds.ago }
```
digitaltom, line 19: The `last_seen_at` is touched on each API interaction of the client, so by querying the API each hour, the client would always get authenticated here. Not sure if that's an issue though.

engines/registry/app/models/access_token.rb
```diff
@@ -20,7 +20,7 @@ def claim
       hash['iss']    = 'RMT' # "matching issuer in registry auth token config"
       hash['sub']    = @account
       hash['aud']    = @service
-      hash['exp']    = Time.now.getlocal.to_i + (5 * 60) # expires at
+      hash['exp']    = Time.now.getlocal.to_i + Settings[:registry].token_expiration # expires at
```
digitaltom, line 23: As far as I can remember, the docker/podman client was not re-using access tokens for subsequent commands, so this might not have an effect here. @cwh42 looked deeper into this, can you share what was the client behavior regarding token expiration?

engines/registry/app/models/registry/authenticated_client.rb
rjschwei: Yes it would be. The idea is to expire the token after 8 hours no matter how often the system comes to visit in those 8 hours.

MANUAL.md
jesusbv: @digitaltom updated

config/rmt.yml
jesusbv: Updated

engines/registry/app/models/access_token.rb
jesusbv: I have checked the behavior on testing infra server and it works as expected:
- running search/pull within the time frame before expiring => returns search or it is able to pull a container
- running search/pull outside the time frame                           => requires credentials refresh

Updated to use hardcoded or settings from file too

engines/registry/app/models/registry/authenticated_client.rb
bear454: As a user, if auth is valid for 8 hours, and I re-auth every, say 4 hours, I should never encounter a time when I cannot access the system. Forcing expiration, even with a re-auth, doesn't make sense to me.

jesusbv: I see the behavior correct as it is, if the re-auth happens after X amount of time, default 8h, is because the credentials have not been refreshed/used, meaning there is no guarantee that after that time the system is still registered or valid or whatever state the system is

If during those 8h, system calls any API that updates `last_seen_at` that guarantees that the system is registered and can have access to stuff that required registration/auth, like the registry

rjschwei: I think we need to pick this apart some more to make sure we are all on the same page.

"auth" i.e. verification of the system credentials against the IMDS data happens only in 2 ways:

1. The RMT server is accessed with zypper, because zypper always sends the IMDS data and if the system is not in the cache then the full verification path takes place
2. The user executes a forced re-auth via the new command we will provide client side.

If the update of `last_seen_at` in the DB only happens when zypper comes to visit then the implementation as is in this PR works as expected because we are guaranteed that the proper auth happened. Repository access expires after 20 minutes and then a full verification happens when the repository path is visited again. However, if `last_seen_at` is updated on the registry path as well then the implementation is not correct because we cannot do a full verification on the registry path since the client, docker or podman, have no way of sending the metadata.

My question is, when is `last_seen_at` updated? 
@digitaltom states "on each API interaction of the client" which would be every time the heartbeat beacon runs, i.e. hourly, and then we are not meeting the requirement of "a full verification every 8 hours for the registry path." It could also be when the registry path is accessed, same problem.

We would only meet the requirement if `last_seen_at` is updated **only** on the repository access path. According to the comment from @digitaltom this is not true. Therefore the implementation in this PR does not meet the requirement.

It appears to me, based on [1] that `last_seen_at` is updated only on successful authentication of a system (instance) which in the Public Cloud means access via the repository path. Somone with more knowledge about the details will need to confirm this please.

[1] https://github.com/SUSE/rmt/blob/master/app/controllers/application_controller.rb#L25

config/rmt.yml
rjschwei: Well the idea of the hard coding was to remove it from the config. At least in my diff in the UI I still see the value in the config.

engines/registry/app/models/registry/authenticated_client.rb
jesusbv: > My question is, when is last_seen_at updated?

It gets updated with `zypper` calls (e.g.`zypper ref`, `zypper up`, `zypper in <whatever>`) if done at least 3min apart [2] or `SUSEConnect -p <whatever>` 
Meaning if the system's `last_seen_at` is at, let's say 10:04:00, and expiration is set to be at 10:05:00 and user ran any of those commands,`last_seen_at` would change to, let's say 10:07:00 then user would not need to run reauth command as `zypper` updated that time and access would be granted which I understood to be OK behavior

> The idea is to expire the token after 8 hours no matter how often the system comes to visit in those 8 hours

I understood the behavior of the PR correct until that comment, if we want the credentials to expire after 8 hours, independently of how many times `zypper` or `SUSEConnect -p` are called (which is what that comment states), or how many times `last_seen_at` has been updated (if any) then we'd need another way. If we do not mind that the system access for the registry has been granted by running `zypper` and that it may not expire every 8 hours because of the same reason, then this PR is OK 

IIRC, `last_seen_at` in the public cloud gets updated on [2], as in the public cloud the system token is disabled


[2] https://github.com/SUSE/rmt/blob/master/app/controllers/application_controller.rb#L30

rjschwei: My comment 

> The idea is to expire the token after 8 hours no matter how often the system comes to visit in those 8 hours

is confusing, you are correct. I should have been more explicit it should be:

**The token expires after 8 hours no matter how often the client comes to visit on the registry path in those 8 hours.**

With this, since a full validation has to happen before `last_seen_at` gets updated this implementation will work. Thanks for the clarification.

rjschwei (approved):

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -16,7 +16,8 @@ def initialize(login, password)
 
   def authenticate_by_system_credentials(login, password)
     @systems = System.get_by_credentials(login, password)
-    if @systems.present?
+    expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
+    if @systems.any? { |system| system.last_seen_at > expiration_value.seconds.ago }
```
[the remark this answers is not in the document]
rjschwei: Thanks @digitaltom . I withdraw my approval, based on this, and will unresolve the previous conversation on this topic to make it easier to find.

And I will restate the requirement:

The token expires 8 hours after the last full verification process, i.e. when we verified the system credentials (username & password) with the metadata.

So if there are paths that do not trigger the full verification but trigger the update of `last_seen_at` then we would not meet the requirement with the implementation in this PR.

That said on the client side the plan was to use the `activations` API to force a token refresh, i.e. put the system back in the cache. Meaning at this point I am still lacking clarity as to how the puzzle fits together.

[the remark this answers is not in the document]
jesusbv: we can use `activations` endpoint to validate the instance metadata and put it in the cache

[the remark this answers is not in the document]
digitaltom: It seems you want to achieve something like: 
`if @systems.any? { |system| system.last_instance_verification_time > expiration_value.seconds.ago }` and assumed that `last_seen_at` == `last_instance_verification_time` which is not the case afaics. (`last_instance_verification_time` is not an existing attribute)

PR #1126, opened 2024-04-18
Revert "Add registry support"

ngetahun:
Reverts SUSE/rmt#1094 as this is meant to be released with rmt 2.17


jesusbv (approved):
Yes, leaving it out of 2.16 release

Plan is to land it on 2.17

PR #1128, opened 2024-04-19
Force reauth

jesusbv:
## Description

If the header for registry cache is set, create an empty file (remote IP-system_login as a name) in
`/var/lib/rmt/tmp/registry/cache/` to use as a cache

Use the cache file together with the expiration date to auth registry access

Once the cache time has expired, remove cache file

Fixes # (issue)

## Change Type

## Checklist

### DIFF
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -54,9 +54,11 @@ module Registry
       # skip authentication if this is not a login request
       return unless request.authorization
 
+      @remote_ip = request.remote_ip
+
       authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
         begin
-          @client = Registry::AuthenticatedClient.new(login, password)
+          @client = Registry::AuthenticatedClient.new(login, password, @remote_ip)
         rescue StandardError
           logger.info _('Could not find system with login \"%{login}\" and password \"%{password}\"') %
             { login: login, password: password }
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -54,11 +54,9 @@ module Registry
       # skip authentication if this is not a login request
       return unless request.authorization
 
-      @remote_ip = request.remote_ip
-
       authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
         begin
-          @client = Registry::AuthenticatedClient.new(login, password, @remote_ip)
+          @client = Registry::AuthenticatedClient.new(login, password, request.remote_ip)
         rescue StandardError
           logger.info _('Could not find system with login \"%{login}\" and password \"%{password}\"') %
             { login: login, password: password }
--- a/engines/registry/app/models/registry/authenticated_client.rb
+++ b/engines/registry/app/models/registry/authenticated_client.rb
@@ -15,11 +15,12 @@ class Registry::AuthenticatedClient
   private
 
   def authenticate_by_system_credentials(login, password, remote_ip)
+    @systems = System.get_by_credentials(login, password)
     expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
     cache_key = [remote_ip, login].join('-')
     registry_cache_path = Rails.root.join('tmp', 'registry', 'cache', cache_key)
     cache_created = File.exist?(registry_cache_path)
-    if cache_created
+    if !@systems.empty? && cache_created
       is_registry_cache_active = File.ctime(registry_cache_path) > expiration_value.seconds.ago
       if is_registry_cache_active
         @account = login
--- a/engines/registry/app/models/registry/authenticated_client.rb
+++ b/engines/registry/app/models/registry/authenticated_client.rb
@@ -3,8 +3,8 @@ class Registry::AuthenticatedClient
 
   attr_reader :auth_strategy
 
-  def initialize(login, password)
-    authenticate_by_system_credentials(login, password)
+  def initialize(login, password, remote_ip)
+    authenticate_by_system_credentials(login, password, remote_ip)
     if @auth_strategy
       Rails.logger.info("Authenticated '#{self}'")
     else
@@ -14,13 +14,18 @@ class Registry::AuthenticatedClient
 
   private
 
-  def authenticate_by_system_credentials(login, password)
+  def authenticate_by_system_credentials(login, password, remote_ip)
     @systems = System.get_by_credentials(login, password)
     expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
-    if @systems.any? { |system| system.last_seen_at > expiration_value.seconds.ago }
+    cache_key = [remote_ip, login].join('-')
+    registry_cache_path = Rails.root.join('tmp', 'registry', 'cache', cache_key)
+    cache_created = File.exist?(registry_cache_path)
+    time_has_not_expired = @systems.any? { |system| system.last_seen_at > expiration_value.seconds.ago }
+    if time_has_not_expired && cache_created
       @account = login
       @auth_strategy = :system_credentials
     end
+    File.delete(registry_cache_path) unless time_has_not_expired
     @auth_strategy
   end
 end
--- a/engines/registry/app/models/registry/authenticated_client.rb
+++ b/engines/registry/app/models/registry/authenticated_client.rb
@@ -15,17 +15,18 @@ class Registry::AuthenticatedClient
   private
 
   def authenticate_by_system_credentials(login, password, remote_ip)
-    @systems = System.get_by_credentials(login, password)
     expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
     cache_key = [remote_ip, login].join('-')
     registry_cache_path = Rails.root.join('tmp', 'registry', 'cache', cache_key)
     cache_created = File.exist?(registry_cache_path)
-    time_has_not_expired = @systems.any? { |system| system.last_seen_at > expiration_value.seconds.ago }
-    if time_has_not_expired && cache_created
-      @account = login
-      @auth_strategy = :system_credentials
+    if cache_created
+      is_registry_cache_active = File.ctime(registry_cache_path) > expiration_value.seconds.ago
+      if is_registry_cache_active
+        @account = login
+        @auth_strategy = :system_credentials
+      end
+      File.delete(registry_cache_path) unless is_registry_cache_active
     end
-    File.delete(registry_cache_path) unless time_has_not_expired
     @auth_strategy
   end
 end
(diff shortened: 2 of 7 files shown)


engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -54,9 +54,11 @@ def basic_auth
       # skip authentication if this is not a login request
       return unless request.authorization
 
+      @remote_ip = request.remote_ip
+
       authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
         begin
-          @client = Registry::AuthenticatedClient.new(login, password)
+          @client = Registry::AuthenticatedClient.new(login, password, @remote_ip)
```
Adnilson, line 61: If this instance variable is not being used anywhere, you can just pass `request.remote_ip`.

It is one less variable assignment.

jesusbv: The original idea was to use it in other places, good catch

rjschwei:
I do not quite understand, can you please provide the connection between this PR and #1124? 

In #1124 the token that the client (docker or podman) use to authenticate expires after 8 hours from `last_seen_at`, why do we need additional handling?

Based on #1124 when the token expires then a new login would be required, i.e. the user either accesses the repository path or runs the new auth command we will provide with the client code.

Sorry, I am confused.

jesusbv:
No need it, as we do not want to force reauth, `last_seen_at` approach is OK

engines/registry/lib/registry/engine.rb
```diff
@@ -1,6 +1,32 @@
...
     isolate_namespace Registry
     config.generators.api_only = true
+
+    config.after_initialize do
+      Api::Connect::V3::Systems::ActivationsController.class_eval do
+        before_action :handle_cache, only: %w[index]
+
+        def handle_cache
+          # get request header
+          if request.headers['X-Refresh-Registry-Creds'] && ZypperAuth.verify_instance(request, logger, @system)
```
jesusbv, line 14: The header name can be changed

The main part is to validate the instance metadata, notice that the validation of the instance metadata would put it in the RMT cache if valid

jesusbv: That header is sent to that endpoint in every use, I used a different one to make the verification explicit for this use case, meaning not instance verification if not registry call

Using `X-Instance-Header` is fine but it would mean we would trigger an instance verification and a potential update for non registry calls

digitaltom: What I understand from my research for https://github.com/SUSE/rmt/pull/1130 suseconnect never sends that header. Or am I wrong there? 
I think it's fair to validate the instance data when it is send and the client is not in the instance data cache.

jesusbv: the [client](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L421-L426)  to `connect/systems/activations` route
 
 > it's fair to validate the instance data when it is send
 
 I agree, I will change the header and will work on adjusting the tests for this

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -14,12 +14,19 @@ def initialize(login, password)
...
+    cache_key = [remote_ip, login].join('-')
+    registry_cache_path = Rails.root.join('tmp', 'registry', 'cache', cache_key)
+    cache_created = File.exist?(registry_cache_path)
+    if !@systems.empty? && cache_created
+      is_registry_cache_active = File.ctime(registry_cache_path) > expiration_value.seconds.ago
+      if is_registry_cache_active
+        @account = login
+        @auth_strategy = :system_credentials
+      end
+      File.delete(registry_cache_path) unless is_registry_cache_active
```
digitaltom, line 29: I think that line is not needed, because you'll need a cronjob anyway(?) to wipe out cache files from that directory older 8h.

jesusbv, line 29: I think we dont mind refreshing the cache while it is still active, given the previous conversations regarding that

engines/registry/lib/registry/engine.rb
```diff
@@ -1,6 +1,32 @@
...
+            FileUtils.mkdir_p(registry_cache_dir_path)
+
+            # update cache if expired
+            # if file exists and exists longer than the cache expiration time
+            # it needs to be updated
+            cache_key = [request.remote_ip, @system.login].join('-')
+            registry_cache_path = File.join(registry_cache_dir_path, cache_key)
+            expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
+            cache_is_valid = File.exist?(registry_cache_path) && File.ctime(registry_cache_path) > expiration_value.seconds.ago
+            FileUtils.touch(registry_cache_path) unless cache_is_valid
```
jesusbv, line 26: I think we dont mind refreshing the cache while it is still active, given the previous conversations regarding that, so I will remove that line unless objections

engines/registry/app/models/registry/authenticated_client.rb
jesusbv: Yes, in case a user do not come back for a period of time, those files would remain on disk
so probably it is better to have a task to remove those altogether instead of when accessing a single one

engines/registry/lib/registry/engine.rb
```diff
@@ -1,6 +1,32 @@
...
 module Registry
   class Engine < ::Rails::Engine
     isolate_namespace Registry
     config.generators.api_only = true
+
+    config.after_initialize do
+      Api::Connect::V3::Systems::ActivationsController.class_eval do
+        before_action :handle_cache, only: %w[index]
+
+        def handle_cache
```
digitaltom, line 12:
```suggestion
        def update_registry_auth_cache
```
would be much more descriptive here

rjschwei:
This approach, which is similar to what we do for repository access with in principal certainly work. That said it appears the scope of the topic and the requirement is not covered by the changes. It appears like another "throw something at the wall and hope it sticks" approach.

While we are under time pressure we can also not produce a quality implementation by approaching the problem with no plan. Based on the conversation in #1124 and the comments in this PR please present a comprehensive plan that covers the requirements. Then we'll proceed to the implementation phase.

This PR for example neglects the "needs to be configurable" requirement. It also does not cover the "8 hours after the last full verification" requirement. With this implementation what would happen is that if I run "zypper up" on the client 7 hour and 59 minutes after the last access when the cache file was created I would still have to run the client side re-auth script because the cache handling here is completely disconnected from the repository access path.

Please develop an understanding of the requirements the we have, which are really only 2

1.) configuration of the expiration period
2.) default 8 hour access after the last full verification of the client system

once you have an understanding of the problem space for these requirements present a plan  how to go about the implementation and then implement in reasonably small parts that are self contained that can be related back to the plan.

engines/registry/lib/registry/engine.rb
```diff
@@ -1,6 +1,32 @@
...
+            registry_cache_dir_path = Rails.root.join('tmp/registry/cache')
+            FileUtils.mkdir_p(registry_cache_dir_path)
+
+            # update cache if expired
+            # if file exists and exists longer than the cache expiration time
+            # it needs to be updated
+            cache_key = [request.remote_ip, @system.login].join('-')
+            registry_cache_path = File.join(registry_cache_dir_path, cache_key)
+            expiration_value = Settings[:registry].try(:token_expiration) || 8.hours.to_i
+            cache_is_valid = File.exist?(registry_cache_path) && File.ctime(registry_cache_path) > expiration_value.seconds.ago
```
digitaltom, line 25: Why do we need to check here if the cache is currently valid? 
The system passed the `ZypperAuth.verify_instance` check, so we should always `touch` the `registry_cache_path` file I think.

engines/registry/lib/registry/engine.rb
```diff
@@ -1,6 +1,32 @@
...
     isolate_namespace Registry
     config.generators.api_only = true
+
+    config.after_initialize do
+      Api::Connect::V3::Systems::ActivationsController.class_eval do
+        before_action :handle_cache, only: %w[index]
+
+        def handle_cache
+          # get request header
+          if request.headers['X-Instance-Data'] && ZypperAuth.verify_instance(request, logger, @system)
```
digitaltom, line 14: Is it ok to fail silently here when `ZypperAuth.verify_instance` fails? Probably you want to raise an exception, so that the client can show an error in this case?

engines/registry/spec/api/connect/v3/systems/activations_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+    end
+
+    context 'with invalid X-Instance-Data headers' do
+      let(:headers) { auth_header.merge(version_header).merge({ 'X-Instance-Data' => 'instance_data' }) }
+
+      it 'Instance metadata verification and no cache update' do
+        expect(ZypperAuth).to receive(:verify_instance)
+        expect(FileUtils).not_to receive(:touch)
+        get '/connect/systems/activations', headers: headers
```
digitaltom, line 23: Related to my comment above, maybe we want to check for an error code here?

engines/registry/spec/api/connect/v3/systems/activations_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    context 'with X-Instance-Data headers' do
+      let(:headers) { auth_header.merge(version_header).merge({ 'X-Instance-Data' => 'instance_data' }) }
+
+      it 'verify the instance metadata and update the cache' do
+        expect(FileUtils).to receive(:touch)
+        allow(File).to receive(:join).and_return('foo')
+
+        allow(FileUtils).to receive(:mkdir_p)
+        allow(ZypperAuth).to receive(:verify_instance).and_return(true)
+        get '/connect/systems/activations', headers: headers
```
digitaltom, line 36: Verify that the response is a success?

engines/registry/spec/app/controllers/registry_controller_spec.rb
```diff
@@ -16,9 +18,12 @@ module Registry
         let(:auth_headers) { { 'Authorization' => ActionController::HttpAuthentication::Basic.encode_credentials(system.login, system.password) } }
 
         it 'succeeds with login + password from secrets' do
+          FileUtils.touch(Rails.root.join('foo'))
```
digitaltom, line 21: more readable: 
```suggestion
          FileUtils.touch(Rails.root.join('tmp', 'registry_auth_cache_test'))
```

engines/registry/lib/registry/engine.rb
digitaltom: I think we have to refresh the cache even when it's active. Else the cache will expire at some point and the client's podman will run into errors until the cache refresh (the scheduled registercloudguest call) will be run.

jesusbv:
Closing this PR in favor of 
#1140 
#1141 
#1142 
#1143

PR #1130, opened 2024-04-24
Extending the pubcloud engine READMEs

digitaltom:
## Description

Adding more details to the READMEs of the engines.


digitaltom:
@jesusbv  @rjschwei @bear454 I extended the pubcloud engine READMEs with my current understanding of what they do. It would be great if you could verify that this is correct, or extend if it's incomplete. 

Q: I understand that instance metadata changes over time, but are those changes relevant for the RMT authentication? I wonder why it's not good enough to use the instance metadata that was send by the system initially during announce.

rjschwei (approved):

rjschwei:
> @jesusbv @rjschwei @bear454 I extended the pubcloud engine READMEs with my current understanding of what they do. It would be great if you could verify that this is correct, or extend if it's incomplete.
> 

Thanks


> Q: I understand that instance metadata changes over time, but are those changes relevant for the RMT authentication? I wonder why it's not good enough to use the instance metadata that was send by the system initially during announce.

@digitaltom, if we stick with the instance metadata that was sent during announce then the user can copy the system credentials to thousands of systems and access the repositories while only paying for the use of one instance. The periodic re-verification of the instance metadata ensures that all systems accessing the repositories have the proper entitelment.

digitaltom:
> @digitaltom, if we stick with the instance metadata that was sent during announce then the user can copy the system credentials to thousands of systems and access the repositories while only paying for the use of one instance. The periodic re-verification of the instance metadata ensures that all systems accessing the repositories have the proper entitelment.

Thanks, understood. In the instance_data cache, there is the client IP combined to the instance_data to make sharing it across systems more complicated I guess. A similar approach could also be done with the initially stored instance_data, to avoid re-validation. But I don't know how expensive the re-validation is, just an idea that came to my mind when studying the workflow.

PR #1140, opened 2024-05-06
Change cache

jesusbv:
## Description

Change the cache for the instances repositories

## Change Type

*Please select the correct option.*

- [ ] **Bug Fix** (a non-breaking change which fixes an issue)
- [X] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [X] I have verified that my code follows RMT's coding standards with `rubocop`.
- [X] I have reviewed my own code and believe that it's ready for an external review.
- [X] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [X] RMT's test coverage remains at 100%.
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Other Notes

Please use this space to provide notes or thoughts to the team, such as tips on how to review/demo your changes.

PR #1139, opened 2024-05-03
Update cache scrubber

rjschwei:
## Description

We will handle the system cache different for access on the repository and registry path. File cache will be in locations determined by a file written from the rails app. The app will also set the expiry time of the cache. Update the scrubber to use the file written by RMT.

Fixes # (issue)

## Change Type

*Please select the correct option.*

- [ ] **Bug Fix** (a non-breaking change which fixes an issue)
- [ x] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [ ] I have verified that my code follows RMT's coding standards with `rubocop`.
- [ x] I have reviewed my own code and believe that it's ready for an external review.
- [ ] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [ x] RMT's test coverage remains at 100%.
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Other Notes

Please use this space to provide notes or thoughts to the team, such as tips on how to review/demo your changes.

PR #1141, opened 2024-05-06
Add registry cache

jesusbv:
## Description

Add the cache for the registry

## Change Type

## Checklist

## Other Notes


jesusbv:
First, merge #1140

rjschwei:
If you build branches on top of each other and then send PRs to the origin branch then the reviewer sees the full diff to the origin branch. Meaning things need to get reviewed twice.

You either need to hold the updates that build on top of each other, or submit PRs into the branch that you are using as origin for the follow on work.

jesusbv:
> If you build branches on top of each other and then send PRs to the origin branch then the reviewer sees the full diff to the origin branch. Meaning things need to get reviewed twice.
> 
> You either need to hold the updates that build on top of each other, or submit PRs into the branch that you are using as origin for the follow on work.

OK, updated the origin branch of the subsequent PRs

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -3,7 +3,9 @@ class Registry::AuthenticatedClient
 
   attr_reader :auth_strategy
 
-  def initialize(login, password)
+  def initialize(login, password, remote_ip)
+    raise Registry::Exceptions::InvalidCredentials.new(message: 'invalid cache', login: login) unless cache_file_exist?(remote_ip, login)
```
rjschwei, line 7: s/invalid cache/expired credentials/

rjschwei (changes requested):

config/application.rb
```diff
@@ -69,8 +69,11 @@ class Application < Rails::Application
       config.access_policies = '/etc/rmt/access_policies.yml'
 
       config.cache_config_file = '/var/lib/rmt/rmt-cache-trim.sh'
+      registry_cache_expiration_time = Settings[:registry].try(:token_expiration) || 480 # 8 hours in minutes
```
digitaltom, line 72: Can we move these engine specific configs to the engine directory?

jesusbv: Good idea

config/application.rb
```diff
@@ -67,6 +67,7 @@ class Application < Rails::Application
         )
       config.registry_public_key = config.registry_private_key.public_key
       config.access_policies = '/etc/rmt/access_policies.yml'
+      config.regisry_cache_dir = '/run/rmt/cache/registry'
```
bear454, line 70: s/regisry/registry/

bear454: This seems to be the only change in `config/application.rb`. Should it be moved along with the other configs (in `production.rb` I think?)

https://github.com/SUSE/rmt/pull/1141/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eL23

bear454 (changes requested):

config/application.rb
jesusbv: Yes, for this PR is OK
@digitaltom suggested to move `Registry` engine config info to the config inside the engine, which makes sense to me but I do not think it should be part of the PRs of registry _cache_

thoughts ?

bear454: The suggested move is good, and I agree that can come in a separate PR.

But, this one line is the only change now on `config/application.rb` - so let's be consistent with the rest of the PR and move it in with the rest of the changes... then `config/application.rb` doesn't get touched.

bear454 (approved):

config/environments/production.rb
```diff
@@ -24,7 +24,15 @@
...
     'REPOSITORY_CACHE_EXPIRY_MINUTES=20'
-  ].join("\n")
+  ]
+  if defined?(Registry::Engine)
+    config.registry_cache_dir = '/run/rmt/cache/registry'
+    cache_config_content.append[
+      "REGISTRY_CLIENT_CACHE_DIRECTORY=#{config.registry_cache_dir}",
+      "REGISTRY_CACHE_EXPIRY_MINUTES=#{Settings[:registry].try(:token_expiration) || 480}" # 480: 8 hours in minutes
+    ]
+  end
```
rjschwei, line 34: This is a nice idea but it will break https://github.com/SUSE/rmt/pull/1139. IN the case where a 3rd party does not set up the registry they still need the scrubber and the scrubber depends on the variables set or the systemd service will exit with an error. While the scrubbing command for he repository path will still work errors will be emitted for `find -mmin -type f -delete` because the keys `REGISTRY_CLIENT_CACHE_DIRECTORY` and `REGISTRY_CACHE_EXPIRY_MINUTES` will not be define the shell will set these to ' ' empty. Defining the variable even the case where there is no registry engine does no harm and as such they should be set always when the Public Cloud path is installed.

rjschwei (changes requested):

config/environments/production.rb
bear454: good catch @rjschwei

rjschwei (approved):

PR #1142, opened 2024-05-06
Add registry credentials refresh

jesusbv:
## Description

Refresh the registry cache explicitly using the activations endpoint

Fixes # (issue)

## Change Type

## Checklist

## Other Notes


engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -1,16 +1,18 @@
 require 'fileutils'
 
 module InstanceVerification
-  def self.update_cache(remote_ip, system_login, product_id, _is_byos)
+  def self.update_cache(remote_ip, system_login, product_id, is_byos: false, registry: false) # rubocop:disable Lint/UnusedMethodArgument
```
rjschwei, line 4: Can't we just drop the unused `is_byos` argument. I am doubtful the solution to the reduce the number of requests to scc with the same set of credentials will include a simple flag.

engines/registry/lib/registry/engine.rb
```diff
@@ -2,5 +2,17 @@ module Registry
   class Engine < ::Rails::Engine
     isolate_namespace Registry
     config.generators.api_only = true
+
+    config.after_initialize do
+      Api::Connect::V3::Systems::ActivationsController.class_eval do
+        before_action :handle_cache, only: %w[index]
+
+        def handle_cache
```
rjschwei, line 10: s/handle_cache/handle_auth_cache/

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: Yes, I will take care of removing that in a different PR

>  I am doubtful the solution to the reduce the number of requests to scc with the same set of credentials will include a simple flag

Well, not the place to talk about it but the solution could follow the same approach as repository|registry and 
has its own directory
`REPO_PAYG` and REPO_BYOS` with its own timer for the scrubber, in any case, that flag will be removed in a different PR

engines/registry/lib/registry/engine.rb
```diff
@@ -2,5 +2,17 @@ module Registry
...
+          unless request.headers['X-Instance-Data'] && ZypperAuth.verify_instance(request, logger, @system, registry: true)
```
rjschwei, line 11: I think it makes sense to push the check for the existence of the header `X-Instance-Data` to the `verify_instance` code. The `verify_instance` code assumes the header is there but should have a shortcut `false` exit condition if it doesn't exist, IMHO.

rjschwei (changes requested):

engines/registry/lib/registry/engine.rb
jesusbv: I like the change you suggest, that implies...that _all_ the calls/requests that pass through `ZypperAuth.verify_instance` *must* have the header set, which is fine by me but implies 
a few changes more than updating that line on registry engine

engines/registry/lib/registry/engine.rb
jesusbv: Done

engines/registry/lib/registry/engine.rb
jesusbv: Updated

engines/registry/lib/registry/engine.rb
```diff
@@ -2,5 +2,17 @@ module Registry
   class Engine < ::Rails::Engine
     isolate_namespace Registry
     config.generators.api_only = true
+
+    config.after_initialize do
+      Api::Connect::V3::Systems::ActivationsController.class_eval do
+        before_action :handle_auth_cache, only: %w[index]
+
+        def handle_auth_cache
+          unless ZypperAuth.verify_instance(request, logger, @system, registry: true)
```
bear454, line 11: It _feels_ like we've overloaded `ZypperAuth` here, having it verify authN for a registry.

IMO it's out of scope for this PR, but it seems that `#verify_instance` should be in `InstanceVerification`, alongside `#update_cache`. As a bonus; the key pattern could be written _once_ in one function, instead of being inline in `InstanceVerification#update_cache` and `ZypperAuth#verify_instance`.

bear454 (approved):

engines/instance_verification/lib/instance_verification/engine.rb
rjschwei: I would be surprised if the solution looks like that.

engines/registry/lib/registry/engine.rb
rjschwei: Why would we care in the authentication path that the full validation happens because of registry access? I do not understand. That just appears to be completely incorrect. A full system validation means we validate the metadata we receive, that is instance data must match the product data. If everything matches we put the system in the cache. So why is the path how we verify the instance important?

bear454:
@jesusbv tests are failing after merging #1140, #1141. Please review & resolve.

engines/registry/lib/registry/engine.rb
rjschwei: GitHub still showed me the `registry` argument, if it's gone then we are good to go.

engines/instance_verification/lib/instance_verification/engine.rb
rjschwei: Still would like to see the rubocop exception dropped, as the need for `is_byos` as an argument is speculation. ANyway, given the time constraints, this is more os a nitpick, thus approving this for merge.

rjschwei (approved):

PR #1161, opened 2024-05-22
Do not carry over the request host, the URL is the same

jesusbv:
## Description

Use the framework value to determine the full URL
and set it in the method instead of carrying over
the different methods and classes

Tests updated

## Change Type

## Checklist

## Other Notes

### DIFF
--- a/engines/registry/app/services/registry_catalog_service.rb
+++ b/engines/registry/app/services/registry_catalog_service.rb
@@ -19,24 +19,25 @@ class RegistryCatalogService
   end
 
   # be aware that this takes about 20-25 seconds to be finished if not in cache
-  def repos(reload: false, system: nil, origin_url: nil)
+  def repos(reload: false, system: nil)
     Rails.cache.fetch(@catalog_api_url, expires_in: 1.hour, force: reload) do
-      fetch_registry_repos(system, origin_url)
+      fetch_registry_repos(system)
     end
   end
 
   private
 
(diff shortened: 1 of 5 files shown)


rjschwei (approved):

engines/registry/app/services/registry_catalog_service.rb
```diff
@@ -19,24 +19,25 @@ def initialize(url = CATALOG_API_URL)
...
+    response = catalog_token(system)
     catalog_auth_token = JSON.parse(response.body).fetch('token', '')
     response = all_repos(catalog_auth_token)
     JSON.parse(response.body).fetch('repositories', [])
   end
 
-  def catalog_token(system, origin_url)
-    uri = URI.parse(URI.join(origin_url, AUTH_URL).to_s)
+  def catalog_token(system)
+    framework = InstanceVerification.provider.name.rpartition('::')[2].downcase
```
bear454, line 39: A note should be added to the README of each InstanceVerifiction provider plugin package, stating that the name is used in this way to inform the URL of the registry server.

bear454 (approved):

engines/registry/app/services/registry_catalog_service.rb
jesusbv: good point, was not sure where to add it to keep track of the dependency, adding the comment there makes sense to me

PR #1165, opened 2024-06-03
Move /tmp to repo directory

### LINKED ISSUE #1137: rmt-cli mirror should use better temporary directory
It seems that all files are downloaded to `/tmp` before they are moved to `/usr/share/rmt/public/repo/`.

Since the directory `/usr/share/rmt/public/repo/` points to the actual storage (the RMT admin owns that symlink), it can easily happen that `rmt-cli mirror` will run into `ENOSPC` because `/tmp` may be too small.

I suggest to use `/usr/share/rmt/public/repo/.tmp` as temporary storage, unless TMPDIR is already set.

Adnilson:
## Description

**What**: RMT does store temporary downloaded metadata in repository root rather then in `/tmp`

**Why**: `/tmp` can not be guaranteed to be big enough to host all downloaded metadata and might run out of memory on download.

Fixes #1137

## Change Type

## Checklist

- [ ] RMT's test coverage remains at 100%.

## Other Notes


features/mirror_spec.rb
```diff
@@ -1,7 +1,8 @@
 require File.expand_path('support/command_rspec_helper', __dir__)
 
 describe 'mirror' do
-  let(:path) { '/var/lib/rmt/public/repo/SUSE/Products/SLE-Product-SLES/15-SP5/x86_64' }
+  let(:files) { %w[repomd.xml repomd.xml.asc repomd.xml.key *-primary.xml.gz *-filelists.xml.gz *-other.xml.gz] }
+  let(:path) { '/var/lib/rmt/public/repo/SUSE/Updates/SLE-Product-SLES/15' }
```
felixsch, line 5:
```suggestion
  let(:path) { '/var/lib/rmt/public/repo/SUSE/Products/SLE-Product-SLES/15-SP5/x86_64' }  
```
the new path seems to be incorrect here. The old path works fine and fixes the broken tests.

features/mirror_spec.rb
```diff
@@ -35,3 +38,25 @@
     expect(`find #{File.join(path, 'product', 'x86_64')} -maxdepth 1 -name sles-release-*.rpm`).to include('sles-release')
   end
 end
+
+describe 'mirror multiple times' do
+  let(:path) { '/var/lib/rmt/public/repo/SUSE/Updates/SLE-Product-SLES/15' }
```
felixsch, line 43: this can be removed no?

features/mirror_spec.rb
```diff
@@ -18,6 +19,8 @@
   let(:metadata_files) { %w[repomd.xml repomd.xml.asc repomd.xml.key *-primary.xml.gz *-filelists.xml.gz *-other.xml.gz] }
 
   it 'has valid metadata mirrored' do
+    expect(File.join(path, 'repodata/repomd.xml')).to exist
```
felixsch, line 22:
```suggestion
    expect(File.exist?(File.join(path, 'repodata/repomd.xml'))).to eq(true)
```
Sadly the initial attempt does not work correctly. Falling back to exist(path) == true works

lib/rmt/mirror/base.rb
```diff
@@ -132,18 +133,22 @@ def need_to_download?(ref)
   end
 
   def cleanup_stale_metadata
-    # A bug introduced in 2.16 writes metadata into its own directory if it exists
-    # resulting in a directory structure like repodata/repodata.
+    logger.debug('Removing stale metadata files...')
```
felixsch, line 136: do we really want the logging here? We had it for development. Do you think we should keep it?

Adnilson: Maybe. Should the user know that the stale data was removed

felixsch (approved):
lgtm :+1: 

**How I reviewed this pull request:**

- Checked the code changes for potential improvments I want to suggest
- Tested mirroring working with following types of repositories:
  - existing repository
  - new repository
  - debian repository
  - repository with licenses

cheers,

PR #1188, opened 2024-07-11
Add index to Systems table

jesusbv:
## Description

Make system_token an index in systems table

* [bsc#1227131](https://bugzilla.suse.com/show_bug.cgi?id=1227131)

## Change Type

## Checklist

## Review

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_01_29_140413) do
+ActiveRecord::Schema.define(version: 2024_07_11_154147) do
 
   create_table "activations", charset: "utf8", force: :cascade do |t|
     t.bigint "service_id", null: false
@@ -174,6 +174,7 @@ ActiveRecord::Schema.define(version: 2024_01_29_140413) do
     t.text "instance_data"
     t.index ["login", "password", "system_token"], name: "index_systems_on_login_and_password_and_system_token", unique: true
     t.index ["login", "password"], name: "index_systems_on_login_and_password"
+    t.index ["system_token"], name: "index_systems_on_system_token"
     t.check_constraint "json_valid(`system_information`)", name: "system_information"
   end
 
--- a/db/migrate/20240711154147_add_index_to_systems_system_token.rb
+++ b/db/migrate/20240711154147_add_index_to_systems_system_token.rb
@@ -1,0 +1,5 @@
+class AddIndexToSystemsSystemToken < ActiveRecord::Migration[6.1]
+  def change
+    add_index :systems, %i[system_token]
+  end
+end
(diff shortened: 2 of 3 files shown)


db/schema.rb
```diff
@@ -174,6 +174,7 @@
     t.text "instance_data"
     t.index ["login", "password", "system_token"], name: "index_systems_on_login_and_password_and_system_token", unique: true
     t.index ["login", "password"], name: "index_systems_on_login_and_password"
+    t.index ["system_token"], name: "index_systems_on_system_token"
```
digitaltom, line 177: In SCC, we have this index (https://github.com/SUSE/happy-customer/blob/master/glue/db/schema.rb#L640-L641): 
`t.index ["login", "password", "system_token"], name: "index_systems_on_login_and_password_and_system_token", unique: true` 
Would it make sense to stay aligned?

rjschwei: I have taken a semi brief look at the code down to https://github.com/SUSE/rmt/blob/master/lib/suse/connect/api.rb#L109 and figure that after this there is a DB query which gets stuck. But I wasn't able to pin point that spot.

A commit message that describes the connection and is more than one sentence would sure be nice such that as a reviewer one doesn't have to reconstruct the whole chain of events.

jesusbv: @digitaltom when running the request for `unscoped_products` we do use that [index](https://github.com/SUSE/rmt/blob/master/engines/scc_suma_api/app/controllers/scc_suma_api/scc_suma_api_controller.rb#L51)


@rjschwei I added the bsc reference, hoped that would explain the details, in any case see statement/link above

digitaltom: Ok, I see how it is used in https://github.com/SUSE/rmt/blob/master/engines/scc_suma_api/app/controllers/scc_suma_api/scc_suma_api_controller.rb#L51 now. Is system login and password not available at that point to query for the system in the standard way?

jesusbv: nope, that info is not available at that point

digitaltom: The password should only show as filtered in the log, but those are not system credentials that SUMA is sending but organization credentials. 
Ok then, it's uncommon to query for a system by token only, but if that's the only way in the SUMA engine then it's ok to add that index. I'd suggest to add a comment there why the system is queried by system token. And test if the migration to add the index creates an issue on the large pubcloud RMT instances.

digitaltom (approved):
Ok then, it's uncommon to query for a system by token only, but if that's the only way in the SUMA engine then it's ok to add that index. I'd suggest to add a comment there why the system is queried by system token. And test if the migration to add the index creates an issue on the large pubcloud RMT instances.

db/schema.rb
rjschwei: I read the bug and I found 

"""
The issue is that there is a query for an unindexed field,
system_token, once that field is made an index:
"""

The underlying assumption in this comment also appears to be that everyone knows the context of the query. Writing good commit messages really helps, https://cbea.ms/git-commit/ just in case you need a reference. A commit message for this case should have probably started with  "When SUMa queries the unscoped data...." then setting the context on why `system_token` is needed in the DB, especially since we have `disable_system_tokens: true` on our systems, which raises the question how SUMa can query by token when RMT doesn't hand on eout in the first place.

rjschwei (approved):

PR #1189, opened 2024-07-12
Print proper error message instead of stacktrace when rmt config cannot get read

digitaltom:
## Description

How to test: 

* Remove your rmt config file
* Run bin/rmt-cli sync
* See proper error message


Adnilson (approved):
How I reviewed:

1. Removed `config/rmt.yml `

2. Ran `bin/rmt-cli sync`

PR #1194, opened 2024-07-30
Remove proxy_byos column and adapt code for proxy_byos_mode column

jesusbv:
## Description

Remove proxy_byos column and its variable use,
adapt the code base to use the new proxy_byos_mode column
to prepare the code for LTSS support

## How to test 

setting a system with proxy_byos_mode should not produce any errors
setting a system with proxy_byos should not be allowed

## Change Type

## Checklist

## Review

### DIFF
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -108,7 +108,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
     # in BYOS mode, we rely on the activation being performed in
     # `engines/scc_proxy/lib/scc_proxy/engine.rb` and don't need further checks here
-    return activation if @system.proxy_byos
+    return activation if @system.byos?
 
     if @subscription.present?
       activation.subscription = @subscription
@@ -124,7 +124,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   def load_subscription
     # Find subscription by regcode if provided, otherwise use the first subscription (bsc#1220109)
-    if params[:token].present? && !@system.proxy_byos
+    if params[:token].present? && @system.payg?
       @subscription = Subscription.find_by(regcode: params[:token])
       unless @subscription
         raise ActionController::TranslatedError.new(N_('No subscription with this Registration Code found'))
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -124,7 +124,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
 
   def load_subscription
     # Find subscription by regcode if provided, otherwise use the first subscription (bsc#1220109)
-    if params[:token].present? && @system.payg?
+    if params[:token].present? && !@system.byos?
       @subscription = Subscription.find_by(regcode: params[:token])
       unless @subscription
         raise ActionController::TranslatedError.new(N_('No subscription with this Registration Code found'))
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -5,8 +5,6 @@ class System < ApplicationRecord
   # PAYG (pay as you go),
   # BYOS (bring your own subscription) or
   # a mix of both (hybrid).
-  # enum proxy_byos_mode: {:no_cloud, :payg, :byos, :hybrid]
-  # System.create(hostname: 'hostname')
   enum proxy_byos_mode: { not_applicable: 0, payg: 1, byos: 2, hybrid: 3 }
 
   after_initialize :init
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -1,6 +1,10 @@
 class System < ApplicationRecord
-
-  enum proxy_byos_mode: { byos: 1, hybrid: 2 }
+  # This value is only used in public cloud scenarios
+  # This value indicates that the system is using
+  # PAYG (pay as you go),
+  # BYOS (bring your own subscription) or
+  # a mix of both (hybrid).
+  enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }
 
   after_initialize :init
 
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -1,10 +1,13 @@
 class System < ApplicationRecord
-  # This value is only used in public cloud scenarios
+  # This value has meaning/relevance only used in public cloud scenarios
   # This value indicates that the system is using
+  # NOT_APPLICABLE (systems outside the public cloud),
   # PAYG (pay as you go),
   # BYOS (bring your own subscription) or
   # a mix of both (hybrid).
-  enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }
+  # enum proxy_byos_mode: {:no_cloud, :payg, :byos, :hybrid]
+  # System.create(hostname: 'hostname')
+  enum proxy_byos_mode: { not_applicable: 0, payg: 1, byos: 2, hybrid: 3 }
 
   after_initialize :init
 
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -1,6 +1,6 @@
 class System < ApplicationRecord
 
-  enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }
+  enum proxy_byos_mode: { byos: 1, hybrid: 2 }
 
   after_initialize :init
 
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -1,7 +1,8 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    add_column :systems, :proxy_byos_mode, :integer, default: nil
-    System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: 1
+    add_column :systems, :proxy_byos_mode, default: 0
+    System.where('proxy_byos = 0 and instance_data IS NOT NULL').in_batches.update_all proxy_byos_mode: :payg
+    System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: :byos
   end
 
   def down
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -1,7 +1,6 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    add_column :systems, :proxy_byos_mode, :integer, default: 0
-    System.where(proxy_byos: false).in_batches.update_all proxy_byos_mode: 0
+    add_column :systems, :proxy_byos_mode, :integer, default: nil
     System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: 1
   end
 
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -1,6 +1,6 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    add_column :systems, :proxy_byos_mode, default: 0
+    add_column :systems, :proxy_byos_mode, :integer, default: 0
     System.where('proxy_byos = 0 and instance_data IS NOT NULL').in_batches.update_all proxy_byos_mode: :payg
     System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: :byos
   end
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -1,11 +1,8 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    safety_assured do
-      add_column :systems, :proxy_byos_mode, :integer, default: 0
-      execute 'Update systems SET proxy_byos_mode = 0 WHERE proxy_byos = false'
-      execute 'Update systems SET proxy_byos_mode = 1 WHERE proxy_byos = true'
-    end
-    # remove_column :systems, :proxy_byos
+    add_column :systems, :proxy_byos_mode, :integer, default: 0
+    System.where(proxy_byos: false).in_batches.update_all proxy_byos_mode: 0
+    System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: 1
   end
 
   def down
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -5,6 +5,7 @@ class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   end
 
   def down
+    System.where(proxy_byos_mode: :byos).in_batches.update_all proxy_byos: true
     remove_column :systems, :proxy_byos_mode
   end
 end
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -6,7 +6,6 @@ class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   end
 
   def down
-    System.where(proxy_byos_mode: :byos).in_batches.update_all proxy_byos: true
     remove_column :systems, :proxy_byos_mode
   end
 end
--- a/db/migrate/20240729103525_update_proxy_byos_column_type.rb
+++ b/db/migrate/20240729103525_update_proxy_byos_column_type.rb
@@ -1,7 +1,7 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
     add_column :systems, :proxy_byos_mode, :integer, default: 0
-    System.where('proxy_byos = 0 and instance_data IS NOT NULL').in_batches.update_all proxy_byos_mode: :payg
+    System.where(proxy_byos: false).where.not(instance_data: nil).in_batches.update_all proxy_byos_mode: :payg
     System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: :byos
   end
 
--- a/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
+++ b/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
@@ -1,11 +1,5 @@
 class RemoveProxyByosBoolean < ActiveRecord::Migration[6.1]
-  def up
-    safety_assured { remove_column :systems, :proxy_byos }
-  end
-
   def down
-    add_column :systems, :proxy_byos, :boolean, default: false
-    System.where(proxy_byos_mode: 0).in_batches.update_all proxy_byos: false
-    System.where(proxy_byos_mode: 1).in_batches.update_all proxy_byos: true
+    System.where(proxy_byos_mode: :byos).in_batches.update_all proxy_byos: true
   end
 end
--- a/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
+++ b/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
@@ -1,0 +1,11 @@
+class RemoveProxyByosBoolean < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured { remove_column :systems, :proxy_byos }
+  end
+
+  def down
+    add_column :systems, :proxy_byos, :boolean, default: false
+    System.where(proxy_byos_mode: 0).in_batches.update_all proxy_byos: false
+    System.where(proxy_byos_mode: 1).in_batches.update_all proxy_byos: true
+  end
+end
--- a/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
+++ b/db/migrate/20240730090211_remove_proxy_byos_boolean.rb
@@ -1,5 +1,0 @@
-class RemoveProxyByosBoolean < ActiveRecord::Migration[6.1]
-  def down
-    System.where(proxy_byos_mode: :byos).in_batches.update_all proxy_byos: true
-  end
-end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -174,6 +174,7 @@ ActiveRecord::Schema.define(version: 2024_07_30_090211) do
     t.text "instance_data"
     t.index ["login", "password", "system_token"], name: "index_systems_on_login_and_password_and_system_token", unique: true
     t.index ["login", "password"], name: "index_systems_on_login_and_password"
+    t.index ["system_token"], name: "index_systems_on_system_token"
     t.check_constraint "json_valid(`system_information`)", name: "system_information"
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_07_30_090211) do
+ActiveRecord::Schema.define(version: 2024_07_29_103525) do
 
   create_table "activations", charset: "utf8", force: :cascade do |t|
     t.bigint "service_id", null: false
@@ -168,6 +168,7 @@ ActiveRecord::Schema.define(version: 2024_07_30_090211) do
     t.datetime "updated_at", null: false
     t.datetime "scc_registered_at"
     t.bigint "scc_system_id", comment: "System ID in SCC (if the system registration was forwarded; needed for forwarding de-registrations)"
+    t.boolean "proxy_byos", default: false
     t.integer "proxy_byos_mode", default: nil
     t.string "system_token"
     t.text "system_information", size: :long
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -168,7 +168,7 @@ ActiveRecord::Schema.define(version: 2024_07_30_090211) do
     t.datetime "updated_at", null: false
     t.datetime "scc_registered_at"
     t.bigint "scc_system_id", comment: "System ID in SCC (if the system registration was forwarded; needed for forwarding de-registrations)"
-    t.integer "proxy_byos_mode", default: 0
+    t.integer "proxy_byos_mode", default: nil
     t.string "system_token"
     t.text "system_information", size: :long
     t.text "instance_data"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -169,7 +169,7 @@ ActiveRecord::Schema.define(version: 2024_07_29_103525) do
     t.datetime "scc_registered_at"
     t.bigint "scc_system_id", comment: "System ID in SCC (if the system registration was forwarded; needed for forwarding de-registrations)"
     t.boolean "proxy_byos", default: false
-    t.integer "proxy_byos_mode", default: nil
+    t.integer "proxy_byos_mode", default: 0
     t.string "system_token"
     t.text "system_information", size: :long
     t.text "instance_data"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_07_29_103525) do
+ActiveRecord::Schema.define(version: 2024_07_30_090211) do
 
   create_table "activations", charset: "utf8", force: :cascade do |t|
     t.bigint "service_id", null: false
@@ -168,7 +168,6 @@ ActiveRecord::Schema.define(version: 2024_07_29_103525) do
     t.datetime "updated_at", null: false
     t.datetime "scc_registered_at"
     t.bigint "scc_system_id", comment: "System ID in SCC (if the system registration was forwarded; needed for forwarding de-registrations)"
-    t.boolean "proxy_byos", default: false
     t.integer "proxy_byos_mode", default: 0
     t.string "system_token"
     t.text "system_information", size: :long
--- a/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
+++ b/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
@@ -14,7 +14,7 @@ module RegistrationSharing
         if !params.key?(:proxy_byos_mode) && system.attribute_names.include?('proxy_byos_mode')
           # the info comes from a sibling that does not have proxy_byos_mode
           # to a sibling does have proxy_byos_mode
-          system.proxy_byos_mode = params[:proxy_byos] ? :byos : :payg
+          system.proxy_byos_mode = system.proxy_byos ? :byos : :payg
         end
         # end todo
         system.activations = []
--- a/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
+++ b/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
@@ -10,6 +10,13 @@ module RegistrationSharing
         system = System.find_or_create_by(login: params[:login])
         system.update(system_params)
 
+        # TODO: remove this block when proxy_byos column gets dropped
+        if !params.key?(:proxy_byos_mode) && system.attribute_names.include?('proxy_byos_mode')
+          # the info comes from a sibling that does not have proxy_byos_mode
+          # to a sibling does have proxy_byos_mode
+          system.proxy_byos_mode = params[:proxy_byos] ? :byos : :payg
+        end
+        # end todo
         system.activations = []
         params[:activations].each do |activation|
           product = Product.find_by(id: activation[:product_id])
--- a/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
+++ b/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
@@ -34,7 +34,7 @@ module RegistrationSharing
     protected
 
     def system_params
-      params.permit(:login, :password, :hostname, :proxy_byos, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
+      params.permit(:login, :password, :hostname, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
     end
 
     def authenticate
--- a/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
+++ b/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
@@ -34,7 +34,7 @@ module RegistrationSharing
     protected
 
     def system_params
-      params.permit(:login, :password, :hostname, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
+      params.permit(:login, :password, :hostname, :proxy_byos, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
     end
 
     def authenticate
--- a/engines/registration_sharing/lib/registration_sharing/client.rb
+++ b/engines/registration_sharing/lib/registration_sharing/client.rb
@@ -21,7 +21,7 @@ class RegistrationSharing::Client
   def peer_register_system(system)
     params = {}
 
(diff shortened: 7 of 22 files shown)


engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -34,7 +34,7 @@ def destroy
     protected
 
     def system_params
-      params.permit(:login, :password, :hostname, :proxy_byos, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
```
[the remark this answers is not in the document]
rjschwei: Updates are staged and during version update we always have a time window of about 48 hours where we operate in a mixed environment, i.e. 2 versions of RMT share registration data.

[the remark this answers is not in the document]
jesusbv: added `proxy_byos` to the params permit list in [259110c](https://github.com/SUSE/rmt/pull/1194/commits/259110cb7df74b499476e79debe5dbb9b0d5dea0)

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveProxyByosBoolean < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured { remove_column :systems, :proxy_byos }
+  end
+
+  def down
+    add_column :systems, :proxy_byos, :boolean, default: false
+    System.where(proxy_byos_mode: 0).in_batches.update_all proxy_byos: false
```
bear454, line 8: This is not necessary, since `default: false`.

db/migrate/20240729103525_update_proxy_byos_column_type.rb
```diff
@@ -1,7 +1,6 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    add_column :systems, :proxy_byos_mode, :integer, default: 0
-    System.where(proxy_byos: false).in_batches.update_all proxy_byos_mode: 0
+    add_column :systems, :proxy_byos_mode, :integer, default: nil
```
bear454, line 3: Why use `nil` as the default? This causes a different behavior than `System#proxy_byos` had, where `default: false` means the default value is PAYG.

IMO since the enum for `System#byos_proxy_mode` defines `payg: 0`, then the default _should be_ `0`.

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -108,7 +108,7 @@ def create_product_activation
 
     # in BYOS mode, we rely on the activation being performed in
     # `engines/scc_proxy/lib/scc_proxy/engine.rb` and don't need further checks here
-    return activation if @system.proxy_byos
+    return activation if @system.byos?
```
bear454, line 111: Where is the method `System#byos?` defined?

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
bear454, line 1: Why break this into two migrations?

lib/rmt/cli/systems.rb
```diff
@@ -8,12 +8,12 @@ class RMT::CLI::Systems < RMT::CLI::Base
   desc 'list', _('List registered systems.')
   option :limit, aliases: '-l', type: :numeric, default: 20, desc: _('Number of systems to display')
   option :all, aliases: '-a', type: :boolean, desc: _('List all registered systems')
-  option :proxy_byos, type: :boolean, desc: _('Filter BYOS systems using RMT as a proxy')
+  option :proxy_byos_mode, type: :boolean, desc: _('Filter BYOS systems using RMT as a proxy')
```
bear454, line 11: If this is `type: :boolean`, how does one filter for hybrid systems?

lib/rmt/scc.rb
```diff
@@ -84,7 +84,8 @@ def sync_systems
     scc_api_client = SUSE::Connect::Api.new(Settings.scc.username, Settings.scc.password)
 
     # do not sync BYOS proxy systems to SCC
-    systems = System.where('scc_registered_at IS NULL OR last_seen_at > scc_registered_at').where(proxy_byos: false)
+    # using where in because `where not` does NOT work for enum on this Rails version
+    systems = System.where('scc_registered_at IS NULL OR last_seen_at > scc_registered_at').where(proxy_byos_mode: [nil, :hybrid])
```
bear454, line 88: Why are we selecting `proxy_byos_mode: nil` ? `nil` is not defined in the _enum_, so therefore should not be considered a valid condition.

bear454 (changes requested):
Get rid of `System#proxy_byos_mode == nil`; the enum for `System#proxy_byos_mode` defines 3 states: PAYG, BYOS, HYBID. We don't need an ambiguous 4th state.

jesusbv:
> Get rid of `System#proxy_byos_mode == nil`; the enum for `System#proxy_byos_mode` defines 3 states: PAYG, BYOS, HYBID. We don't need an ambiguous 4th state.

we cant use `PAYG`, we can use 0 (to match the previous behavior as `false`)
we really have two states 
- BYOS
- HYBRID
and the default one (0 or nil) which for RMT in the public cloud means PAYG

lib/rmt/cli/systems.rb
jesusbv: this is not fixed in this PR
this PR changes the column to a new column and makes sure things are the same way they were before the change

There will be a PR for filtering all the modes

app/controllers/api/connect/v3/systems/products_controller.rb
jesusbv: in the [language](https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html) ?

db/migrate/20240729103525_update_proxy_byos_column_type.rb
jesusbv: OK, will change from `nil` to 0, same meaning

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
jesusbv: good catch, removed the unnecessary from the up method, missed the down one

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
jesusbv: prefer to keep it in small changes regarding the db, one to add the new column, one to remove the old column

app/controllers/api/connect/v3/systems/products_controller.rb
bear454: #TIL ;-)

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
bear454: Do you expect that end users will perform each migration separately, as opposed to `rails db:migrate`, which will perform both `up` sections?

app/models/system.rb
```diff
@@ -1,6 +1,6 @@
 class System < ApplicationRecord
 
-  enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }
+  enum proxy_byos_mode: { byos: 1, hybrid: 2 }
```
bear454, line 3: The default value should be included in the enum (you were right the first time IMO).

Also, please include a validation limiting the field to enum keys, for example
```
validates :proxy_byos_mode, inclusion: { in: proxy_byos_mode.keys }
```

db/schema.rb
```diff
@@ -168,8 +168,7 @@
     t.datetime "updated_at", null: false
     t.datetime "scc_registered_at"
     t.bigint "scc_system_id", comment: "System ID in SCC (if the system registration was forwarded; needed for forwarding de-registrations)"
-    t.boolean "proxy_byos", default: false
-    t.integer "proxy_byos_mode", default: 0
+    t.integer "proxy_byos_mode", default: nil
```
bear454, line 171: Still showing `default: nil`.

lib/rmt/scc.rb
bear454: Still querying on `proxy_byos_mode == nil` .

bear454 (changes requested):
There's still some ambiguity around `System#proxy_byos_mode == nil` .

lib/rmt/cli/systems.rb
rjschwei: I don't understand, I thoght the hole ide of `proxy_byos_mode` was that we get an enum that represents
PAYG -> 0
BYOS -> 1
Hybrid -> 2

Shouldn't we create the new column and during the creation read from the existing value and then populate the new column accordingly?

At the same time when we receive a `proxy_byos` entry from a sibling we should populate both DB entries. 

In the next iteration of RMT we can then drop the sending of `proxy_byos` data and only send `proxy_byos_mode`. To finish up the switch over sequence in a 3rd release of RMT we can then safely drop the `proxy_byos` column.

I do not see a path where this transition can happen in fewer steps given that during upgrade we always run in a mixed environment, i.e. 2 different versions of RMT have to talk to each other.

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveProxyByosBoolean < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured { remove_column :systems, :proxy_byos }
```
rjschwei, line 3: So if the `proxy_byos` column is removed what happens if the sibling server sends that value?

app/models/system.rb
felixsch: Why now removing the `payg: 0` or shouldn't it be rather `payg: nil`?

lib/rmt/cli/systems.rb
jesusbv: what you are describing is an option and valid transition path

the other option is to update [this method](https://github.com/SUSE/rmt/blob/master/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb#L36)
to handle both values

```
    def system_params
      if System.column_names.include?('proxy_byos_mode') && params.include?('proxy_byos')
        # read the old value and convert it to the new value
        foo = params[:proxy_byos]
        params[:proxy_byos_mode] = <correct_value_from_foo>
        params.permit(:login, :password, :hostname, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data) 
      elsif System.column_names.include?('proxy_byos') && params.include?('proxy_byos_mode')
        # update accordingly to proxy_byos
        params.permit(:login, :password, :hostname, :proxy_byos, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
      else # proxy_byos -> proxy_byos case
        ...
      else # proxy_byos_mode -> proxy_byos_mode
      end
    end
```
Missed the if when updated this part on a previous commit

With that we can remove the column at this time, in any case, we would need another release to update this part
I do not have strong opinions on the steps to remove the column, maybe keeping the column and removing it in a different RMT version when all siblings have `proxy_byos_mode` is simpler

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
digitaltom: The code that receives the instance sharing data needs to convert `proxy_byos=true` to `proxy_byos_mode=byos`.

lib/rmt/scc.rb
jesusbv: I do not understand, I haven't committed any change yet, what do you mean with "still querying nil" ?

app/models/system.rb
jesusbv: because when I ran tests on `payg: 0` all systems were marked as `proxy_byos_mode = 'payg'` that did not have `proxy_byos` set before which is not correct for RMT outside the public cloud scenario

digitaltom, line 3:
```suggestion
  # This value is only used in public cloud scenarios and can indicate that the system is not using PAYG (pay as you go), but BYOS (bring your own subscription) or a mix of both (hybrid).
  enum proxy_byos_mode: { byos: 1, hybrid: 2 }
```

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
jesusbv: we would need to adapt a regsharing method on the controller, see  [this comment](https://github.com/SUSE/rmt/pull/1194#discussion_r1699604905)

I'm fine keeping the column and dropping once all siblings are OK

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
jesusbv: I expect users will run `rake db:migrate` nothing to do with how the users are going to run migration
one PR to add the new column
one PR to remove the old column + keep the same behavior as the old column

easier to review IMO

digitaltom: A RMT is automatically running db:migrate on start: 
https://github.com/SUSE/rmt/blob/update-based-on-enum/package/files/systemd/rmt-server-migration.service#L22

app/models/system.rb
jesusbv: > The default value should be included in the enum (you were right the first time IMO).
> 
> Also, please include a validation limiting the field to enum keys, for example
> 
> ```
> validates :proxy_byos_mode, inclusion: { in: proxy_byos_mode.keys }
> ```

Validation is created by default, no need to set anything 
```
irb(main):007:0> System.create!(hostname: 'hostname', proxy_byos_mode: :aa)
        1: from (irb):7
ArgumentError ('aa' is not a valid proxy_byos_mode)
irb(main):008:0> 
```

jesusbv: wrt to the default value and the enum,
yes, `enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }` is more explicit
with that we would still the default to be `nil` to support non pubcloud scenario

which leads to the question _why would we have two ways to express the same thing_ 
I personally prefer 
```
# This value is only used in public cloud scenarios 
# It indicates that the system is using 
# PAYG (pay as you go) if nil, 
# BYOS (bring your own subscription) or 
# a mix of both (hybrid).
enum proxy_byos_mode: { byos: 1, hybrid: 2 }
```

jesusbv: wrt to the default value and the enum,
yes, `enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }` is more explicit
with that we would still need the default to be `nil` to support non pubcloud scenario

which leads to the question _why would we have two ways to express the same thing_ 
I personally prefer 
```
# This value is only used in public cloud scenarios 
# It indicates that the system is using 
# PAYG (pay as you go) if nil, 
# BYOS (bring your own subscription) or 
# a mix of both (hybrid).
enum proxy_byos_mode: { byos: 1, hybrid: 2 }
```

jesusbv: wrt to the default value and the enum,
yes, `enum proxy_byos_mode: { payg: 0, byos: 1, hybrid: 2 }` is more explicit
with that we would still need the default to be `nil` to support non pubcloud scenario

which leads to the question _why would we have two ways to express the same thing_ 
I personally prefer 
```
# This value is only used in public cloud scenarios 
# It indicates that the system is using 
# PAYG (pay as you go) if nil, 
# BYOS (bring your own subscription) or 
# a mix of both (hybrid).
enum proxy_byos_mode: { byos: 1, hybrid: 2 }
```

jesusbv: Thoughts between `nil, :payg, :byos, :hybrid` and `nil, :byos, :hybrid` ?

lib/rmt/scc.rb
jesusbv: the nil query must remain unless we use `nil, :payg, :byos, :hybrid`
Rails has the negative scope for enum borked
```bash
# emtpy DB
irb(main):022:0> System.all
D, [2024-08-01T10:54:39.143333 #9120] DEBUG -- :   System Load (0.7ms)  SELECT `systems`.* FROM `systems` /* loading for inspect */ LIMIT 11
=> #<ActiveRecord::Relation []>
irb(main):023:0> System.create(hostname: 'hostname', proxy_byos_mode: :byos)
D, [2024-08-01T10:54:57.262121 #9120] DEBUG -- :   TRANSACTION (0.3ms)  BEGIN
D, [2024-08-01T10:54:57.263324 #9120] DEBUG -- :   System Create (0.8ms)  INSERT INTO `systems` (`hostname`, `created_at`, `updated_at`, `proxy_byos_mode`) VALUES ('hostname', '2024-08-01 09:54:57', '2024-08-01 09:54:57', 1)
D, [2024-08-01T10:54:57.338011 #9120] DEBUG -- :   TRANSACTION (73.9ms)  COMMIT
=> #<System id: 5, login: nil, password: nil, hostname: "hostname", registered_at: nil, last_seen_at: nil, created_at: "2024-08-01 09:54:57.000000000 +0000", updated_at: "2024-08-01 09:54:57.000000000 +0000", scc_registered_at: nil, scc_system_id: nil, proxy_byos: false, system_token: nil, system_information: nil, instance_data: nil, proxy_byos_mode: "byos">
irb(main):024:0> System.where.not(proxy_byos_mode: :byos)
D, [2024-08-01T10:55:10.947890 #9120] DEBUG -- :   System Load (0.8ms)  SELECT `systems`.* FROM `systems` WHERE `systems`.`proxy_byos_mode` != 1 /* loading for inspect */ LIMIT 11
=> #<ActiveRecord::Relation []>
irb(main):026:0> System.create(hostname: 'hostname')
D, [2024-08-01T10:55:32.926403 #9120] DEBUG -- :   TRANSACTION (0.4ms)  BEGIN
D, [2024-08-01T10:55:32.927511 #9120] DEBUG -- :   System Create (0.7ms)  INSERT INTO `systems` (`hostname`, `created_at`, `updated_at`) VALUES ('hostname', '2024-08-01 09:55:32', '2024-08-01 09:55:32')
D, [2024-08-01T10:55:32.973673 #9120] DEBUG -- :   TRANSACTION (45.2ms)  COMMIT
=> #<System id: 6, login: nil, password: nil, hostname: "hostname", registered_at: nil, last_seen_at: nil, created_at: "2024-08-01 09:55:32.000000000 +0000", updated_at: "2024-08-01 09:55:32.000000000 +0000", scc_registered_at: nil, scc_system_id: nil, proxy_byos: false, system_token: nil, system_information: nil, instance_data: nil, proxy_byos_mode: nil>


# this should show the system with nil on proxy_byos_mode
irb(main):027:0> System.where.not(proxy_byos_mode: :byos)
D, [2024-08-01T10:55:36.176109 #9120] DEBUG -- :   System Load (0.8ms)  SELECT `systems`.* FROM `systems` WHERE `systems`.`proxy_byos_mode` != 1 /* loading for inspect */ LIMIT 11
=> #<ActiveRecord::Relation []>
irb(main):028:0> System.where(proxy_byos_mode: [:hybrid, nil])
D, [2024-08-01T10:55:38.069228 #9120] DEBUG -- :   System Load (0.9ms)  SELECT `systems`.* FROM `systems` WHERE (`systems`.`proxy_byos_mode` = 2 OR `systems`.`proxy_byos_mode` IS NULL) /* loading for inspect */ LIMIT 11
=> #<ActiveRecord::Relation [#<System id: 6, login: nil, password: nil, hostname: "hostname", registered_at: nil, last_seen_at: nil, created_at: "2024-08-01 09:55:32.000000000 +0000", updated_at: "2024-08-01 09:55:32.000000000 +0000", scc_registered_at: nil, scc_system_id: nil, proxy_byos: false, system_token: nil, system_information: nil, instance_data: nil, proxy_byos_mode: nil>]>
irb(main):029:0> 
```

db/migrate/20240730090211_remove_proxy_byos_boolean.rb
digitaltom: It's pretty unconventional to have a migration which only defines a `down` method. I'd also suggest to merge that into the previous migration, which has the corresponding change in its `up` method.

jesusbv: done :heavy_check_mark:

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -34,7 +34,7 @@ def destroy
     protected
 
     def system_params
-      params.permit(:login, :password, :hostname, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
+      params.permit(:login, :password, :hostname, :proxy_byos, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
```
rjschwei, line 37: What happens when we send `proxy_byos_mode` from a sibling that is in RMT 2.19 to a sibling that is on RMT 2.17? On 2.17 `proxy_byos_mode` does not exist in the DB.

engines/registration_sharing/lib/registration_sharing/client.rb
```diff
@@ -21,7 +21,7 @@ def sync_system
   def peer_register_system(system)
     params = {}
 
-    %w[login password hostname proxy_byos system_token registered_at created_at last_seen_at instance_data].each do |attribute|
+    %w[login password hostname proxy_byos_mode system_token registered_at created_at last_seen_at instance_data].each do |attribute|
```
rjschwei, line 24: Doesn't the client still need to receive `proxy_byos`?

rjschwei:
In the "how to test" section you state 

"""
setting a system with proxy_byos should not be allowed
"""

Can you please explain how the transition from `proxy_byos` to `proxy_byos_mode` is supposed to happen in one step?

During upgrade in the infrastructure we will run in mixed mode, i.e. some systems will be on RMT. 2.17 and others will be on RMT 2.19. Meaning when a BYOS registration is received on an RMT 2.17 system it will send `proxy_byos` to the RMT 2.19 server as part of sharing the client system entry. I do not see code that handles this case.

Then when a byos registration comes along and hits the RMT 2.19 server first it will send `proxy_byos` and `proxy_byos_mode` to an RMT 2.17 server, but the RMT 2.17 server doesn't know what to do with `proxy_byos_mode` what happens then?

And we will have the situation where we will have 2 RMT 2.19 systems and 1 RMT 2.17 system that have to stay consistent with each other and not loose any information.

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jesusbv: we can [adapt/update](https://github.com/SUSE/rmt/pull/1194#discussion_r1699604905) the code in the regsharing engine to handle that scenario

```ruby
    def system_params
      if System.column_names.include?('proxy_byos_mode') && params.include?('proxy_byos')
        # read the old value and convert it to the new value
        foo = params[:proxy_byos]
        params[:proxy_byos_mode] = <correct_value_from_foo>
        params.permit(:login, :password, :hostname, :proxy_byos_mode, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data) 
      elsif System.column_names.include?('proxy_byos') && params.include?('proxy_byos_mode')
        # update accordingly to proxy_byos
        params.permit(:login, :password, :hostname, :proxy_byos, :system_token, :registered_at, :created_at, :last_seen_at, :instance_data)
      else # proxy_byos -> proxy_byos case
        ...
      else # proxy_byos_mode -> proxy_byos_mode
      end
    end
```
With that we can remove the column at this time, in any case, we would need another release to update this part
I do not have strong opinions on the steps to remove the column, maybe keeping the column and removing it in a different RMT version when all siblings have `proxy_byos_mode` is simpler

engines/registration_sharing/lib/registration_sharing/client.rb
jesusbv: not necessarily, depending on how we handle the regsharing info

if we keep the column `proxy_byos` to be removed in a different iteration, then yes, we would need to send `proxy_byos` 

if we are [handling the info](https://github.com/SUSE/rmt/pull/1194#discussion_r1701587459) on the regsharing receiving side, then `proxy_byos_mode` should do

jesusbv:
> In the "how to test" section you state
> 
> """ setting a system with proxy_byos should not be allowed """
> 
> Can you please explain how the transition from `proxy_byos` to `proxy_byos_mode` is supposed to happen in one step?
> 
Yes, that is valid if we decide to remove the column on this step, which I think we do not want, and that test is valid or focused on to test registration on `RMT 2.19`, I did not include registration sharing on the test step

for registration sharing test the behavior is as you described

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
digitaltom: I'd suggest to stay compatible with registration sharing with RMTs < 2.19, the registration sharing will need to send both parameters (`proxy_byos` + `proxy_byos_mode`) and be able to receive both parameters (`proxy_byos` + `proxy_byos_mode`) to cover both directions of sync between new + old RMTs: 
RMT 2.17 -> RMT 2.19, RMT 2.19 -> RMT 2.17

Once all pubcloud RMTs are upgraded, we can remove all usage of `proxy_byos` + also drop the column.

jesusbv: yes, I agree with that, and I have kept the `proxy_byos` column (no more `remove` migration code), regsharing code needs to be update to include that column, and depending on the enum we need to update the actual migration

jesusbv:
OK, added a value for non public cloud systems to the enum and, after agreement on the name, 
enum got updated to `enum proxy_byos_mode: { not_applicable: 0, payg: 1, byos: 2, hybrid: 3 }` 
default value being 0, so migration got updated to populate the `:payg` systems

we keep the `proxy_byos` column (to be removed on a different RMT version when all siblings got `proxy_byos_mode`)
and regsharing got updated to reflect that

engines/registration_sharing/lib/registration_sharing/client.rb
```diff
@@ -21,7 +21,7 @@ def sync_system
   def peer_register_system(system)
     params = {}
 
-    %w[login password hostname proxy_byos system_token registered_at created_at last_seen_at instance_data].each do |attribute|
+    %w[login password hostname proxy_byos proxy_byos_mode system_token registered_at created_at last_seen_at instance_data].each do |attribute|
```
digitaltom, line 24: I would have expected some code in the registration sharing that translates from `proxy_byos=true` to `proxy_byos_mode: :byos`? So that when an older RMT sends that value, still the new field gets updated.

digitaltom (commented):
Don't we need an additional test for the hybrid activation? Having an existing payg system, and activating a product with registration code. I cannot see any code in the pr that actually sets a system to `proxy_byos_mode = :hybrid`.

db/migrate/20240729103525_update_proxy_byos_column_type.rb
```diff
@@ -1,11 +1,11 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
-    add_column :systems, :proxy_byos_mode, :integer, default: 0
-    System.where(proxy_byos: false).in_batches.update_all proxy_byos_mode: 0
```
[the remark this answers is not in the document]
jesusbv: maybe the diff does not show correctly [this line](https://github.com/SUSE/rmt/pull/1194/files#diff-eb253f372efbd10b83ead541463aa1a0fb307e2e4cc3e67eba6b67f3e868a52eR4) but on the migration the line has

```ruby
System.where('proxy_byos = 0 and instance_data IS NOT NULL').in_batches.update_all proxy_byos_mode: :payg
```

db/migrate/20240729103525_update_proxy_byos_column_type.rb
```diff
@@ -1,11 +1,11 @@
...
+    add_column :systems, :proxy_byos_mode, :integer, default: nil
     System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: 1
   end
 
   def down
+    System.where(proxy_byos_mode: :byos).in_batches.update_all proxy_byos: true
```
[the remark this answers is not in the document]
jesusbv: My understanding is that those system would get back being `payg` instead of `hybrid` which in the `proxy_byos` context means `false` or `0`

jesusbv:
> Don't we need an additional test for the hybrid activation? Having an existing payg system, and activating a product with registration code. I cannot see any code in the pr that actually sets a system to `proxy_byos_mode = :hybrid`.

No hybrid systems yet, AFAIK or AFAUI, hybrid test will be added in the subsequent PRs, my intent was to introduce the new column and keep the system as it was with `proxy_byos` but using the new column, does that make sense ?

engines/registration_sharing/lib/registration_sharing/client.rb
jesusbv: :white_check_mark:

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
rjschwei: Sorry, I still do not follow. 

> regsharing code needs to be update to include that column,

yes the regsharing code needs to be updated, but it cannot be updated retroactively into RMT < 2.19. If I understand it correctly, at present, if RMT 2.19 gets deployed and we send along the new entry then RMT < 2.19 will blow up because the DB entry does not yet exist.

What am I missing?

jesusbv: In the scenario of the migration removing the `proxy_byos` column on RMT 2.19:
Yes, but when the params for regsharing are sent, the RMT 2.19 can send the two (`proxy_byos` and `proxy_byos_mode` even if the column does not exist in current server, we know the values it should have

in the receiving RMT < 2.19 we can check if the system instance has `proxy_byos` or `proxy_byos_mode` and the content of the params and act accordingly

So, it would work but, as mentioned in a different comment, we would need an extra step to remove that handling logic, hopefully that question/topic is resolved now


That said, this PR does not remove the `proxy_byos` column on this step as we are not saving any step doing that in this PR and seems simpler that handling the logic stated above

db/migrate/20240729103525_update_proxy_byos_column_type.rb
```diff
@@ -1,11 +1,12 @@
 class UpdateProxyByosColumnType < ActiveRecord::Migration[6.1]
   def up
     add_column :systems, :proxy_byos_mode, :integer, default: 0
-    System.where(proxy_byos: false).in_batches.update_all proxy_byos_mode: 0
-    System.where(proxy_byos: true).in_batches.update_all proxy_byos_mode: 1
+    System.where('proxy_byos = 0 and instance_data IS NOT NULL').in_batches.update_all proxy_byos_mode: :payg
```
bear454, line 4: Consider `System.where(proxy_byos: false).where.not(instance_data: nil)` for consistency.

jesusbv: good point

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
bear454: How about adding tests for a 2.19 system sharing with <2.19, and vice versa? This should clarify the behavior.

db/migrate/20240729103525_update_proxy_byos_column_type.rb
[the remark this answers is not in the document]
bear454: OK

lib/rmt/scc.rb
```diff
@@ -84,7 +84,7 @@ def sync_systems
     scc_api_client = SUSE::Connect::Api.new(Settings.scc.username, Settings.scc.password)
 
     # do not sync BYOS proxy systems to SCC
-    systems = System.where('scc_registered_at IS NULL OR last_seen_at > scc_registered_at').where(proxy_byos: false)
+    systems = System.where('scc_registered_at IS NULL OR last_seen_at > scc_registered_at').not_byos
```
bear454, line 87: Wouldn't this include `:not_applicable` ?

jesusbv: yes, same behavior as before

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -10,6 +10,13 @@ def create
         system = System.find_or_create_by(login: params[:login])
         system.update(system_params)
 
+        # TODO: remove this block when proxy_byos column gets dropped
+        if !params.key?(:proxy_byos_mode) && system.attribute_names.include?('proxy_byos_mode')
```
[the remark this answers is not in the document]
jesusbv: because this RMT version includes `proxy_byos_mode`

`proxy_byos` is included in 2.17 and 2.19

so if a system gets registered in 2.17 and shared the info to a 2.19, we want `proxy_byos_mode` to be populated as well

db/migrate/20240729103525_update_proxy_byos_column_type.rb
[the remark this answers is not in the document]
rjschwei: I don't think the case where `proxy_byos_mode: hybrid` is relevant. It would imply that we managed to update the server in the update infrastructure to RMT 2.19 or greater, see my other comments and then decide to to backwards.

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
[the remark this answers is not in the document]
bear454: My concern is that the code checks that `proxy_byos_mode` param is _not present_ but does not check that `proxy_byos` param _is present_, even though that's the param used in subsequent code.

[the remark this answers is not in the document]
bear454: I understand that right now in our codebase those things are logically equal, but it won't always be the case, so IMO it makes more sense to test for the thing you're going to use. Am I missing something?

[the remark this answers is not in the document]
jesusbv: `proxy_byos _is_ present, we have not removed that column/param neither before this version nor in this version, why check something we know it is there ?

db/migrate/20240729103525_update_proxy_byos_column_type.rb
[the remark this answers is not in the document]
jesusbv: which comment, @rjschwei ?

I guess we can remove line 8 as the `proxy_byos` column isn't getting removed

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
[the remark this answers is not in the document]
jesusbv: said with different words, that handling needs to happen if `proxy_byos_mode` is not present in the params and the RMT server has the _new_ param, otherwise, nothing to do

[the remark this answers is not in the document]
bear454: Okay, acknowledging that we're just going to live with this code smell for now.

bear454 (approved):
LGTM

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
rjschwei: > in the receiving RMT < 2.19 we can check if the system instance has proxy_byos or proxy_byos_mode and the content of the params and act accordingly

How? We cannot do anything that is not already there, the code for RMT < 2.19 is frozen....

Anyone else getting an "oh shit" moment? Or am I really missing something that is obvious to everyone else?

jesusbv: OK, so RMT < 2.19 is frozen, on RMT < 2.19 we can send _whatever_ we want, let's say we send `proxy_byos` **and** `proxy_byos_mode` (from an RMT 2.19, for example) then [we allow some params to be present](https://github.com/SUSE/rmt/blob/56ed01ffc0e88f8bc48dd453dd27819cd1209d0f/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb#L37) meaning....the info on RMT < 2.19 is OK

did I explain it better this time ?

Now if we send the the info from RMT < 2.19 to RMT 2.19 [we have a logic to handle the missing param](https://github.com/SUSE/rmt/blob/master/engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb#L14) `proxy_byos_mode`, as in to include the value into the `system` being created as a consecuence of regsharing

PR #1197, opened 2024-08-02
Check for X-Instance-Data header to refresh the registry

jesusbv:
## Description

While the method verify_instance of ZypperAuth engine  checks for the presence of this header, we need to check its presence when we are handling the regitry credentials refresh to keep the cache alive

A side effect of not checking the header before ZypperAuth engine method call is that any other call that tries to call 'activations' endpoint would fail

For examble, when the client runs SUSEConnect -s or SUSEConnect -l

This Fixes bsc#1228774

* Related Issue / Ticket / Trello card: [bugzilla entry](https://bugzilla.suse.com/show_bug.cgi?id=1228774)

## How to test 

After registering a client to RMT 2.17 

running `sudo SUSEConnect -s` or `sudo SUSEConnect -l` must work
running `sudo /usr/bin/cloudguestregistryauth` must work

## Change Type

## Checklist

## Review


kvanderveer (commented):
Tom's on holiday at least on Monday. Please add Felix as reviewer.

rjschwei (approved):

engines/registry/lib/registry/engine.rb
```diff
@@ -15,7 +15,7 @@ class Engine < ::Rails::Engine
         before_action :handle_auth_cache, only: %w[index]
```
digitaltom, line 15: I think it would be more readable like this: 

```suggestion
        # only run instance verification and cache refresh in PAYG mode
        before_action :refresh_auth_cache, only: %w[index], if: -> { request.headers['X-Instance-Data'] }

        def refresh_auth_cache
          unless ZypperAuth.verify_instance(request, logger, @system)
            render(xml: { error: 'Instance verification failed' }, status: :forbidden)
          end
        end
```

digitaltom (approved):
I'm lacking a test setup for this, and don't have full insights into the instance verification workflow, but LGTM reading the code. 
Added a small note about an improvement for making the intent of the before_action more clear.

jesusbv:
> I'm lacking a test setup for this

is [this](https://github.com/SUSE/rmt/pull/1197/files#diff-bd2f861ea06b597df3c4f710af2490628c95fa076ac628b8ce11b02863ac04f4R10-R15) not enough ?

engines/registry/lib/registry/engine.rb
bear454: That's a really elegant fix @tomschr .

felixsch:
@jesusbv can you add a Changelog entry (starting 2.19 release) which adds this PR including a bugzilla entry?

digitaltom:
It's approved and can get merged from my pov.

PR #1198, opened 2024-08-05
Update scc proxy engine to support LTSS 

jesusbv:
## Description

Update scc_proxy engine to support LTSS scenario

## How to test 

On a registered PAYG instance, adding LTSS extension with a valid registration code must succeed

## Change Type

## Checklist

## Review


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -80,24 +80,32 @@ def prepare_scc_announce_request(uri_path, auth, params)
...
         hwinfo: params['hwinfo'],
-        byos: true
-      }.to_json
+        byos_mode: params['proxy_byos_mode']
+      }
+      if params['proxy_byos_mode'].casecmp('hybrid').zero?
+        scc_req_body[:login] = params['scc_login']
+        scc_req_body[:password] = params['scc_password']
+      end
       scc_request
```
bear454, line 92: `scc_request` has no body.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -113,11 +121,11 @@ def announce_system_scc(auth, params)
       JSON.parse(response.body)
     end
 
-    def scc_activate_product(product, auth, token, email)
+    def scc_activate_product(product, auth, params, mode)
```
bear454, line 124: please explain the change in method parameters

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,11 +272,33 @@ def has_no_regcode?(auth_header)
...
+              base_prod = @system.products.find_by(product_type: :base)
+              if base_prod.present? && @product.arch == base_prod.arch && @product.version == base_prod.version
+                request.headers['proxy_byos_mode'] = mode
+                request.headers['scc_login'] = @system.login
+                request.headers['scc_password'] = @system.password
+                request.headers['instance_data'] = params[:instance_data]
+                request.headers['hwinfo'] = params[:hwinfo]
+                response = SccProxy.announce_system_scc(auth, request.headers)
+              end
+            end
```
bear454, line 299: Let's reduce the complexity instead of ignoring the check. Consider:

```
mode = nil
if @system.byos?
  mode = 'byos'
elsif !@product.free? && @product.extension?
  mode = 'hybrid'
  ...
end

**Review by bear454 — CHANGES_REQUESTED** [review]

**Comment by bear454** [comment]
Before this gets merged, someone from the SCC team needs to approve the stubbed SCC requests & responses.

File: engines/scc_proxy/lib/scc_proxy/engine.rb [hunk 4]
```diff
@@ -80,24 +80,32 @@ def prepare_scc_announce_request(uri_path, auth, params)
...
-      scc_request = Net::HTTP::Post.new(uri_path, headers(auth, nil))
+    def prepare_scc_request(uri_path, product, auth, params, mode)
+      params_header = params
+      if mode == 'byos'
+        params_header = nil
+      end
+      scc_request = Net::HTTP::Post.new(uri_path, headers(auth, params_header))
       scc_request.body = {
-        token: token,
+        token: params[:token],
```
rjschwei, line 102: Should this also have a `|| nil` , like `email` look up below for consistency, or vice versa?

engines/scc_proxy/lib/scc_proxy/engine.rb
rjschwei: Yep the entry condition is overly complex.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -276,9 +307,15 @@ def scc_activate_product
             end
             logger.info "Product #{@product.product_string} successfully activated with SCC"
             InstanceVerification.update_cache(request.remote_ip, @system.login, @product.id)
+            # if the system was PAYG and the registration code is valid for the extension,
```
rjschwei, line 310: s/was/is/

rjschwei: The comment suggest that we are checking the validity of the registration code. I do not see such a check.

jesusbv: I think the check is there ? in line 300 `response = SccProxy.scc_activate_product(@product, auth, params, mode)`

the SCC endpoint for activating a product with a regcode/token performs the check, so if we reach that point, regcode/token is validation is OK

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: if a system is PAYG, and activates an extension that is not free, would only end up on SCC if a regcode is set, or
are we passing any extension activation of a PAYG system to SCC (even without regcode) ?

Maybe there should be a check for the token, earlier than this point for PAYG systems

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: good catch

engines/instance_verification/spec/requests/api/connect/v3/systems/products_controller_spec.rb
```diff
@@ -20,200 +21,587 @@
...
+        stub_request(:post, scc_activate_url)
+          .to_return(
+            status: 201,
+            body: {}.to_json,
+            headers: {}
+          )
+        # stub the fake announcement call PAYG has to do to SCC
+        # to create the system before activate product (and skip validation)
+        stub_request(:post, 'https://scc.suse.com/connect/subscriptions/systems')
+          .to_return(status: 201, body: { ok: 'OK' }.to_json, headers: {})
```
digitaltom, line 572: The response from SCC is a hash with the system's login + password (docs: https://scc.suse.com/connect/v4/documentation#/subscriptions/post_subscriptions_systems): 

```
{
  "id": 1234567,
  "login": "SCC_3b336b126db1503a9513a14e92a6a62e",
  "password": "24f057b7941e80f9cf2d51e16e8af2d6"
}
```

In case of the HYBRID announcement call, RMT needs to include the login/password in the announcement payload. It's already implemented in SCC to allow announcement with preset login/pw.

jesusbv: what about the response body for `POST connect/systems/products` ?

.github/workflows/lint-unit.yml
```diff
@@ -77,7 +77,7 @@ jobs:
 
       - name: Prepare database
         run: |
-          bundle exec rails db:migrate
+          bundle exec rails db:reset db:migrate
```
digitaltom, line 80: `db:reset` is short for `db:drop db:create  db:schema:load  db:seed`. This loads the schema from schema.rb which is not desired here, because we want to test that the migrations are working. 

What you could do if you want to make sure here that the db is re-setup: `rake db:drop db:create  db:migrate`

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: Before, the mode was `byos` now we need to send something to indicate `hybrid` as well, 
so instead of 3 parameter (`token`, `email` and `mode`), I decided to send 2 (`params` which include `token` and `email` and `mode`)

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: @rjschwei ^

.github/workflows/lint-unit.yml
```diff
@@ -77,7 +77,7 @@ jobs:
 
       - name: Prepare database
         run: |
-          bundle exec rails db:migrate
+          bundle exec rails db:drop db:create db:migrate
```
bear454, line 80: Why not use `db:reset` ?

engines/scc_proxy/lib/scc_proxy/engine.rb
bear454: I think it would have been more straightforward, and require less change, to just add the `mode` argument to the function. I don't see the value in also changing the behavior of  `token` and `email` arguments.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +277,52 @@ def has_no_regcode?(auth_header)
 
         protected
 
+        # rubocop:disable Metrics/PerceivedComplexity
+        # rubocop:disable Metrics/CyclomaticComplexity
         def scc_activate_product
           logger.info "Activating product #{@product.product_string} to SCC"
-          auth = request.headers['HTTP_AUTHORIZATION']
+          auth = nil
+          auth = request.headers['HTTP_AUTHORIZATION'] if request.headers.include?('HTTP_AUTHORIZATION')
```
bear454, line 285: What does `request.headers['HTTP_AUTHORIZATION']` evaluate to if it's not included in the headers? It should be `nil` anyway, in which case you can just shorten L284 & 285 to 

```
auth = request.headers['HTTP_AUTHORIZATION']
```

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +277,52 @@ def has_no_regcode?(auth_header)
...
             unless response.code_type == Net::HTTPCreated
               error = JSON.parse(response.body)
               logger.info "Could not activate #{@product.product_string}, error: #{error['error']} #{response.code}"
               error['error'] = SccProxy.parse_error(error['error']) if error['error'].include? 'json'
               raise ActionController::TranslatedError.new(error['error'])
             end
+            # if the system is PAYG and the registration code is valid for the extension,
+            # then the system is hybrid
+            # update the system to HYBRID mode if HYBRID MODE and system not HYBRID already
+            @system.hybrid! if mode == 'hybrid' && @system.payg?
```
bear454, line 317: How would we end up with `mode == 'hybrid'` and **not** `@system.payg?` ?

.github/workflows/lint-unit.yml
jesusbv: [0d1f394](https://github.com/SUSE/rmt/pull/1198/commits/0d1f394f17e3ef58cdd636d288e46af87c040567)

https://github.com/SUSE/rmt/pull/1198#discussion_r1707105507

> It feels like this needs to be in a separate PR

without that the tests do not pass

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: when a `hybrid` system activates an extension with a registration code ?

bear454: So, this is in anticipation of a 2nd extension being activated?

engines/scc_proxy/lib/scc_proxy/engine.rb
rjschwei: Only add on products will be passed to SCC.

engines/scc_proxy/lib/scc_proxy/engine.rb
rjschwei: > when a `hybrid` system activates an extension with a registration code ?

A system can only be hybrid when it is PAYG and an add-on product gets activated. The activation of an add-on product requires a registration code.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +276,60 @@ def has_no_regcode?(auth_header)
...
+            # the extensions must be the same version and arch
+            # than base product
+            base_prod = @system.products.find_by(product_type: :base)
+            if base_prod.present? && @product.arch == base_prod.arch && @product.version == base_prod.version
+              request.headers['proxy_byos_mode'] = mode
+              request.headers['scc_login'] = @system.login
+              request.headers['scc_password'] = @system.password
+              request.headers['instance_data'] = params[:instance_data]
+              request.headers['hwinfo'] = params[:hwinfo]
+              response = SccProxy.announce_system_scc(auth, request.headers)
```
bear454, line 306: On further review, I don't think you need `auth` at all...

Consider dropping https://github.com/SUSE/rmt/pull/1198/files#diff-0e5c13122cb7577a624fa4e30eaa3fcfcc0b84a046770be73b96160ba993beb4R284-R285 
... and using here
```
response = SccProxy.announce_system_scc(request.headers['HTTP_AUTHORIZATION'], request.headers)
```

bear454: ... and adding a note to later refactor `SccProxy#announce_system_scc` to not be redundant ;-)

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +276,60 @@ def has_no_regcode?(auth_header)
...
+            end
+          end
+
+          unless mode.nil?
+            # if system is byos or hybrid and there is a token
+            # make a request to SCC
+            response = SccProxy.scc_activate_product(@product, auth, params, mode)
```
bear454, line 313: One more consumer of `auth`

bear454 (commented):
Any further comments I have are purely stylistic.

jesusbv:
> Any further comments I have are purely stylistic.

so, all the changes left are more cosmetic than adding/removing functionality, am I wrong ?
if I understood it right, then we can merge this PR so we can move on to the next PR
and have a subsequent PR (once all the LTSS functionality is merged) to refactor `auth` on `scc_proxy` engine
Thoughts ?

bear454 (approved):
I don't have confidence that I understand that the PR does what it's supposed to do. The code looks to mostly conform to style expectations and passes the tests, but most of the PR is additional tests that mock behavior of SCC, and I have no confidence that I understand that those interactions function as they're supposed to.

So, I would strongly prefer that either @felixsch or @digitaltom approve as well before merging.

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -124,7 +124,7 @@ def remove_previous_product_activations(product_ids)
 
   def load_subscription
     # Find subscription by regcode if provided, otherwise use the first subscription (bsc#1220109)
```
digitaltom, line 126:
```suggestion
        # Find subscription by regcode if provided and not a public cloud system, 
        # otherwise check if there's already an activation with a matching 
        # subscription (for migrations, bsc#1220109)
```

engines/instance_verification/config/environments/test.rb
```diff
@@ -6,5 +6,5 @@
 Rails.application.configure do
   config.cache_config_file = Rails.root.join('engines/registry/spec/data/rmt-cache-trim.sh')
   config.repo_cache_dir = 'repo/cache'
-  config.registry_cache_dir = 'registry/cache'
+  config.registry_cache_dir = 'test_registry/test_cache'
```
digitaltom, line 9: Why does the test env need a different directory here, and not in the other cases ('repo_cache_dir')? If this is needed, I'd suggest to only use a different subdir, like 'registry/test_cache'. In general, the better place might have been 'tmp/cache/registry' but I guess that's too late to change?

digitaltom: Then you can drop this change here, and tackle the cache directory updates in another pr.

engines/registry/spec/app/models/access_scope_spec.rb
```diff
@@ -157,6 +157,7 @@
 
         it 'returns default auth actions (no free repos included)' do
           yaml_string = access_policy_content
+          FileUtils.cp('engines/registry/spec/data/access_policy_yaml.yml', 'engines/registry/spec/data/access_policy_yaml_old.yml')
```
digitaltom, line 160: I suggest to pull out this unrelated change from this pr into another one, refactoring the registry access tests. I don't understand why the file is not at 'spec/data/access_policies.yml' and needs to get copied around.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -80,24 +80,36 @@ def prepare_scc_announce_request(uri_path, auth, params)
...
         hostname: params['hostname'],
         hwinfo: params['hwinfo'],
-        byos: true
-      }.to_json
+        byos_mode: params['proxy_byos_mode']
+      }
+      # when system is payg, we do not know whether it's hybrid or not
+      # we send the login and password information to skip the validation
+      # on the SCC side, that info is enough to validate the product later on
+      # if the system is, in fact, hybrid
```
digitaltom, line 91: This comment reads like as if PAYG systems will announce to SCC. That should not happen. Only in the case when the system is detected as byos or hybrid it should get announced to SCC.

engines/registry/spec/app/models/access_scope_spec.rb
jesusbv: > I don't understand why the file is not at 'spec/data/access_policies.yml' and needs to get copied around

the files is `spec/data/access_policies_yaml.yml` which of course can be renamed, it gets _moved_ because
the tests checks the content, it gets to its original content so no `.gitignore` is needed

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: we don't know if PAYG is hybrid until after validate the registration code

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -80,24 +80,36 @@ def prepare_scc_announce_request(uri_path, auth, params)
...
+      if params['proxy_byos_mode'].casecmp('hybrid').zero?
```
digitaltom, line 92: Let's make clear why we add the system's credentials here. Also, I don't understand why we need to compare case insensitive here?

```suggestion
      # When activating a BYOS extension on top of a PAYG system ('hybrid mode'), the system already has credentials in RMT, but is not known to SCC. We announce it to SCC including its credentials in this case.
      if params['proxy_byos_mode'] == 'hybrid'
```

In general, wouldn't it be useful to provide the @system instance to all of the methods in this file, that would safe us from passing around all those parameters.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +276,60 @@ def has_no_regcode?(auth_header)
 
         protected
 
+        # rubocop:disable Metrics/PerceivedComplexity
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/AbcSize
         def scc_activate_product
```
digitaltom, line 282: This method needs refactoring to be easier to understand. This can be refactored in another pr, but it's important that this workflow is readable. How about a structure like this: 

```
# Proxying BYOS product activations to SCC
def scc_activate_product
  if byos_activation? 
    SccProxy.announce_system_scc unless @system.scc_id 
    SccProxy.scc_activate_product
  end
end
```

engines/scc_proxy/spec/requests/api/connect/v3/systems/products_controller_debug_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'rails_helper'
```
digitaltom, line 1: Is this file a debugging leftover?

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -263,22 +276,60 @@ def has_no_regcode?(auth_header)
...
+            mode = 'byos'
+          elsif !@product.free? && @product.extension?
+            regcode_error_message = 'A registration code is required to activate this product'
+            raise ActionController::TranslatedError.new(regcode_error_message) if (@system.payg? || @system.hybrid?) && params[:token].nil?
+
+            # system is not BYOS and
+            # product is a non free extension
+            # and a registration code is provided
+            mode = 'hybrid'
+            # the extensions must be the same version and arch
```
digitaltom, line 297: Product extensions not necessarily have the same version as the base.

engines/scc_proxy/lib/scc_proxy/engine.rb
digitaltom: I think it's just the wording here. I saw below that only activations attempts with regcode would get announced to SCC, that's fine.

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: Humm, OK, so this `PAYG + registration code that is not expired + base product that matches the version for the LTSS subscription` from the dev plan, does not apply when activating products ?
My understanding was that it does apply

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: I agree on the refactoring, yes, not only that but more stuff can be refactored on that method

engines/scc_proxy/lib/scc_proxy/engine.rb
digitaltom: For LTSS this works, but in general extensions can have different versions. We can still change this when more extensions will be supported.

digitaltom (approved):

PR #1203, opened 2024-08-13
Handle non free extension with regcode

jesusbv:
## Description

When registering a PAYG system with a registration code for an add-on product, if the product needs to be activated via SCC, do not raise the exception unless there is not token for the activation

## How to test 

Register a system with the LTSS regcode (new LTSS code on the client), this exception should not raise

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -58,7 +58,7 @@ module InstanceVerification
 
           if product.base?
             verify_base_product_activation(product)
-          elsif !product.free? && (!params.key?(:token) || params[:token].empty?)
+          elsif !product.free? && params[:token].blank?
             verify_payg_extension_activation!(product)
           end
         rescue InstanceVerification::Exception => e
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -58,9 +58,8 @@ module InstanceVerification
 
           if product.base?
             verify_base_product_activation(product)
-          elsif !product.free? && @system.payg?
-            token_not_provided = !params.key?(:token) || params[:token].empty?
-            verify_payg_extension_activation!(product) if token_not_provided
+          elsif !product.free? && (!params.key?(:token) || params[:token].empty?)
+            verify_payg_extension_activation!(product)
           end
         rescue InstanceVerification::Exception => e
           unless @system.byos?
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -58,8 +58,9 @@ module InstanceVerification
 
           if product.base?
             verify_base_product_activation(product)
-          else
-            verify_extension_activation(product)
+          elsif !product.free? && @system.payg?
+            token_not_provided = !params.key?(:token) || params[:token].empty?
+            verify_payg_extension_activation!(product) if token_not_provided
           end
         rescue InstanceVerification::Exception => e
           unless @system.byos?
@@ -79,11 +80,10 @@ module InstanceVerification
           raise ActionController::TranslatedError.new('Unexpected instance verification error has occurred')
         end
 
-        def verify_extension_activation(product)
+        def verify_payg_extension_activation!(product)
           return if product.free?
 
           base_product = @system.products.find_by(product_type: :base)
-
           subscription = Subscription.joins(:product_classes).find_by(
             subscription_product_classes: {
               product_class: base_product.product_class
@@ -91,7 +91,7 @@ module InstanceVerification
           )
 
           # This error would occur only if there's a problem with subscription setup on SCC side
-          raise "Can't find a subscription for base product #{base_product.product_string}" unless subscription
+          raise InstanceVerification::Exception, "Can't find a subscription for base product #{base_product.product_string}" unless subscription
 
           allowed_product_classes = subscription.product_classes.pluck(:product_class)
 
@@ -100,6 +100,8 @@ module InstanceVerification
               'The product is not available for this instance'
             )
           end
+          logger.info "Product #{@product.product_string} available for this instance"
+          InstanceVerification.update_cache(request.remote_ip, @system.login, @product.id)
         end
 
         def verify_base_product_activation(product)
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -100,7 +100,7 @@ module InstanceVerification
             )
           end
           logger.info "Product #{@product.product_string} available for this instance"
-          InstanceVerification.update_cache(request.remote_ip, @system.login, @product.id)
+          InstanceVerification.update_cache(request.remote_ip, @system.login, product.id)
         end
 
         def verify_base_product_activation(product)
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -307,6 +307,7 @@ module SccProxy
 
         # rubocop:disable Metrics/PerceivedComplexity
         # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/AbcSize
         def scc_activate_product
           logger.info "Activating product #{@product.product_string} to SCC"
           auth = nil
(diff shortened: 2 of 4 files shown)


engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -100,6 +100,11 @@ def verify_extension_activation(product)
               'The product is not available for this instance'
             )
           end
+        rescue InstanceVerification::Exception
+          # could not find a subscription for a non-free extension/add-on product
+          # it may require a token in order to be activated via SCC
+          # if no token provided, raise the exception
+          raise unless params.key?(:token) && !params[:token].empty?
```
digitaltom, line 107: Why does this method getting called at all in the case of a byos/hybrid activation? 
Would it be possible to rename it to `verify_payg_extension_activation!`, and not calling it at all in byos/hybrid case?

rjschwei: @digitaltom I am not certain I fully understand the question, but in the hybrid case we do need to check that the underlying base product is a PAYG system.

rjschwei: It appears that the logic could be changed here.

In line 94 we raise and that Exception would get caught immediately on line 103 where we then check if there is a token. And we raise potentially on 99 to be caught immediately on line 103.

Appears to me we should check the token condition first, i.e.
```
if ! (params.key?(:token) && !params[:token].empty?)
``` 
and then have the existing code be inside this if condition. Or am I missing something?

digitaltom: This method (`verify_extension_activation`), currently only handles payg extension activations, by checking if the extension product is included in the system's PAYG base product subscription. That's why I proposed to rename it to reflect this. 

In the hybrid case, we have a payg system and a requested product activation with provided registration code. I think the presence of the registration code is the trigger there to go into a different code path that forwards the activation request to SCC. And not go into the payg verification, raise and exception, and then hide the exception in the rescue block.

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -58,8 +58,9 @@ def verify_product_activation
 
           if product.base?
             verify_base_product_activation(product)
-          else
-            verify_extension_activation(product)
+          elsif !product.free? && @system.payg?
```
jesusbv, line 61: shouldn't we just ask if product is free ?

Meaning a `hybrid` system may have extensions included in its subcription and may does not need to call SCC for that product ?

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -287,26 +287,23 @@ def scc_activate_product
...
-              request.headers['scc_password'] = @system.password
-              request.headers['instance_data'] = params[:instance_data]
-              request.headers['hwinfo'] = params[:hwinfo]
-              response = SccProxy.announce_system_scc(auth, request.headers)
+            unless token_not_provided
+              mode = 'hybrid'
+              # the extensions must be the same version and arch
+              # than base product
+              base_prod = @system.products.find_by(product_type: :base)
+              if @system.payg? && base_prod.present? && @product.arch == base_prod.arch && @product.version == base_prod.version
```
jesusbv, line 299: announce the `hybrid` system to SCC just the first time

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -58,8 +58,9 @@ def verify_product_activation
...
+            token_not_provided = !params.key?(:token) || params[:token].empty?
+            verify_payg_extension_activation!(product) if token_not_provided
```
digitaltom, line 63: Wouldn't that be: 

```suggestion
          elsif !product.free? && !params[:token].present?
            verify_payg_extension_activation!(product)
```

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -100,6 +100,8 @@ def verify_extension_activation(product)
               'The product is not available for this instance'
             )
           end
+          logger.info "Product #{@product.product_string} available for this instance"
+          InstanceVerification.update_cache(request.remote_ip, @system.login, @product.id)
```
digitaltom, line 104: Isn't this `product` instead of `@product`?

```suggestion
          InstanceVerification.update_cache(request.remote_ip, @system.login, product.id)
```

rjschwei (approved):

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -287,26 +287,23 @@ def scc_activate_product
...
-              request.headers['proxy_byos_mode'] = mode
-              request.headers['scc_login'] = @system.login
-              request.headers['scc_password'] = @system.password
-              request.headers['instance_data'] = params[:instance_data]
-              request.headers['hwinfo'] = params[:hwinfo]
-              response = SccProxy.announce_system_scc(auth, request.headers)
+            unless token_not_provided
+              mode = 'hybrid'
+              # the extensions must be the same version and arch
+              # than base product
```
digitaltom, line 297: So if the version + arch need to match, why is this combined with the condition that determines if an SCC announce is needed? 
How about: 
`raise 'Incompatible extension product' unless @product.arch == base_prod.arch && @product.version == base_prod.version`

engines/scc_proxy/lib/scc_proxy/engine.rb
digitaltom: So the condition to announce the system to SCC is `@system.payg? && base_prod.present?`?

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -287,26 +287,23 @@ def scc_activate_product
...
-            # than base product
-            base_prod = @system.products.find_by(product_type: :base)
-            if base_prod.present? && @product.arch == base_prod.arch && @product.version == base_prod.version
-              request.headers['proxy_byos_mode'] = mode
-              request.headers['scc_login'] = @system.login
-              request.headers['scc_password'] = @system.password
-              request.headers['instance_data'] = params[:instance_data]
-              request.headers['hwinfo'] = params[:hwinfo]
-              response = SccProxy.announce_system_scc(auth, request.headers)
+            unless token_not_provided
```
digitaltom, line 294: How about: 

```
elsif !@product.free? && @product.extension? && params[:token].present?
```

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -58,8 +58,8 @@ def verify_product_activation
 
           if product.base?
             verify_base_product_activation(product)
-          else
-            verify_extension_activation(product)
+          elsif !product.free? && (!params.key?(:token) || params[:token].empty?)
```
digitaltom, line 61:
```suggestion
          elsif !product.free? && params[:token].blank?
```

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: if it is the fake announcement but yes, in the subsequent PR to de-register/de-activate I handle the removal of this announcement if activation failed, as expected

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -287,26 +287,23 @@ def scc_activate_product
...
+              # the extensions must be the same version and arch
+              # than base product
+              base_prod = @system.products.find_by(product_type: :base)
+              if @system.payg? && base_prod.present? && @product.arch == base_prod.arch && @product.version == base_prod.version
+                request.headers['proxy_byos_mode'] = mode
+                request.headers['scc_login'] = @system.login
+                request.headers['scc_password'] = @system.password
+                request.headers['instance_data'] = params[:instance_data]
+                request.headers['hwinfo'] = params[:hwinfo]
+                response = SccProxy.announce_system_scc(auth, request.headers)
```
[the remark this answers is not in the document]
jesusbv: would [4ad0c34](https://github.com/SUSE/rmt/pull/1203/commits/4ad0c3442213506fe0e4fe61ea89390ba93a588e) + [this](https://github.com/SUSE/rmt/blob/master/engines/scc_proxy/lib/scc_proxy/engine.rb#L95) be enough ? I think it would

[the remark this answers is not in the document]
digitaltom: I don't think so. We're intercepting an activation call here, and sending an announce call to SCC. They have different auth methods, that's why we cannot copy the `request.headers['HTTP_AUTHORIZATION']` into `auth`. As described above, the `auth` needs to get build with the regcode. 
Also, I don't know if `params[:instance_data]` and `params[:hwinfo]` are actually part of the client's request payload. Maybe those values need to get taken from `@system` instead.

bear454 (approved):

engines/scc_proxy/lib/scc_proxy/engine.rb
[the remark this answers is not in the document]
jesusbv: @digitaltom I think that is addressed in [75ee5b0](https://github.com/SUSE/rmt/pull/1203/commits/75ee5b0e977cb5d987a7e1b23e55071e24d10ac5)

digitaltom (approved):

PR #1204, opened 2024-08-15
De registrer hybrid system product

jesusbv:
## Description

De-register a hybrid system from SCC

De-activate a hybrid system extension from SCC

When activating an extension via SCC for a PAYG system,
the system must be announced to SCC then the activation of the extension happens

If that activation failed, the system should be removed from SCC

Refactor tests to include PAYG and separate BYOS from PAYG

## How to test 
running `--clean` on the client on a hybrid system should work
running `SUSEConnect -d -p <extension>` in a hybrid system should remove the extension from SCC
and the client

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -58,7 +58,7 @@ module InstanceVerification
 
           if product.base?
             verify_base_product_activation(product)
-          else
+          elsif !product.free? && params[:token].blank?
             verify_extension_activation(product)
           end
         rescue InstanceVerification::Exception => e
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -59,7 +59,7 @@ module InstanceVerification
           if product.base?
             verify_base_product_activation(product)
           elsif !product.free? && params[:token].blank?
-            verify_extension_activation(product)
+            verify_payg_extension_activation!(product)
           end
         rescue InstanceVerification::Exception => e
           unless @system.byos?
@@ -79,11 +79,10 @@ module InstanceVerification
           raise ActionController::TranslatedError.new('Unexpected instance verification error has occurred')
         end
 
-        def verify_extension_activation(product)
+        def verify_payg_extension_activation!(product)
           return if product.free?
 
           base_product = @system.products.find_by(product_type: :base)
-
           subscription = Subscription.joins(:product_classes).find_by(
             subscription_product_classes: {
               product_class: base_product.product_class
@@ -91,7 +90,7 @@ module InstanceVerification
           )
 
           # This error would occur only if there's a problem with subscription setup on SCC side
-          raise "Can't find a subscription for base product #{base_product.product_string}" unless subscription
+          raise InstanceVerification::Exception, "Can't find a subscription for base product #{base_product.product_string}" unless subscription
 
           allowed_product_classes = subscription.product_classes.pluck(:product_class)
 
@@ -100,6 +99,8 @@ module InstanceVerification
               'The product is not available for this instance'
             )
           end
+          logger.info "Product #{@product.product_string} available for this instance"
+          InstanceVerification.update_cache(request.remote_ip, @system.login, product.id)
         end
 
         def verify_base_product_activation(product)
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -324,7 +324,7 @@ module SccProxy
             if @system.payg? && base_prod.present?
               raise 'Incompatible extension product' unless @product.arch == base_prod.arch && @product.version == base_prod.version
 
-              params['hostname'] = @system.system_information['hostname']
+              params['hostname'] = @system.hostname
               params['proxy_byos_mode'] = mode
               params['scc_login'] = @system.login
               params['scc_password'] = @system.password
@@ -391,7 +391,7 @@ module SccProxy
           elsif @system.hybrid? && @product.extension?
             # check if product is on SCC and
             # if it is -> de-activate it
-            scc_systems_activations = find_hybrid_product_on_scc(headers)
+            scc_systems_activations = find_hybrid_product_on_scc(request.headers)
             result = SccProxy.product_class_access(scc_systems_activations, @product.product_class)
             if result[:is_active] || (!result[:is_active] && result[:message].downcase.include?('expired'))
               # if product is active on SCC or
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -165,7 +165,6 @@ module SccProxy
       error_message
     end
 
-    # this method is also in PR#1202
     def get_scc_activations(headers, system_token, mode)
       auth = headers['HTTP_AUTHORIZATION'] if headers && headers.include?('HTTP_AUTHORIZATION')
       uri = URI.parse(SYSTEMS_ACTIVATIONS_URL)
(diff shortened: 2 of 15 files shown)


engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -58,7 +58,7 @@ def verify_product_activation
 
           if product.base?
             verify_base_product_activation(product)
-          else
+          elsif !product.free? && params[:token].blank?
```
bear454, line 61: How does this implement de-registration?

jesusbv: it is not directly related to the de-registration but it is needed for the tests

The change is implemented [here](https://github.com/SUSE/rmt/pull/1203/files#diff-7edbe8015a11c2bffc38e12ec3aa8e97c68e66201ef3bf64947d3bad213ea1e1R61) being part of #1203

once #1203 gets merged, it will get merged here and that change will be gone from the diff

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -318,6 +352,12 @@ def scc_activate_product
               error = JSON.parse(response.body)
               logger.info "Could not activate #{@product.product_string}, error: #{error['error']} #{response.code}"
               error['error'] = SccProxy.parse_error(error['error']) if error['error'].include? 'json'
+              if @system.payg?
+                # if trying to activate first product on a hybrid system
+                # it means the system was "just" announced on this call
+                # if product activation failed, system should get de-register from SCC
+                de_register_hybrid(request.headers['HTTP_AUTHORIZATION'])
```
rjschwei, line 359: s/de_register/deregister/

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -329,6 +369,17 @@ def scc_activate_product
             InstanceVerification.update_cache(request.remote_ip, @system.login, @product.id)
           end
         end
+
+        def de_register_hybrid(auth)
```
rjschwei, line 373: s/de_register/deregister/

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -329,6 +369,17 @@ def scc_activate_product
...
+          response = SccProxy.deregister_system_scc(auth, @system.system_token)
+          unless response.code_type == Net::HTTPNoContent
+            error = JSON.parse(response.body)
+            logger.info "Could not de-activate system #{@system.login}, error: #{error['error']} #{response.code}"
+            error['error'] = SccProxy.parse_error(error['error'], params[:token], params[:email]) if error['error'].include? 'json'
+            raise ActionController::TranslatedError.new(error['error'])
+          end
+          logger.info 'System successfully deregistered from SCC'
+        end
```
rjschwei, line 382: Why do we need a separate implementation? I would have expected that we simply check if a system that is getting removed is hybrid and if that is the case call the existing `deregister_system_scc()` method. What is special about the hybrid system that requires this extra wrapper?

rjschwei (changes requested):

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: > Why do we need a separate implementation? 

To not get `Avoid more than 3 levels of block nesting.` 
If the code inside ~de_register_hybrid~ `deregister_hybrid` method goes where that method is right now,
we would get  that message and rubocop would not be happy, making the linter fail

> if a system that is getting removed is hybrid

No system is getting removed on that call, a product of a system is getting **activated**, the condition is
if the system - whose product is getting activated - was just announced to SCC because it is `hybrid` scenario

>  if that is the case call the existing deregister_system_scc() method

Yes, but we need to handle the error, if any

> What is special about the hybrid system that requires this extra wrapper ?

Nothing special, just to avoid rubocop errors

We could move the error handling inside `SccProxy.deregister_system_scc`  which, if no one objects, I will do

engines/instance_verification/lib/instance_verification/engine.rb
digitaltom: Let's merge #1203 then to avoid confusion and duplication in this pr.

jesusbv: merged

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -355,6 +362,17 @@ def scc_activate_product
...
+
+        def de_register_hybrid(auth)
+          response = SccProxy.deregister_system_scc(auth, @system.system_token)
+          unless response.code_type == Net::HTTPNoContent
+            error = JSON.parse(response.body)
+            logger.info "Could not de-activate system #{@system.login}, error: #{error['error']} #{response.code}"
+            error['error'] = SccProxy.parse_error(error['error'], params[:token], params[:email]) if error['error'].include? 'json'
+            raise ActionController::TranslatedError.new(error['error'])
+          end
+          logger.info 'System successfully deregistered from SCC'
```
bear454, line 374: How can the system be treated as "successfully deregistered" regardless of the response from SCC?

Shouldn't this be in a condition based on some positive response from SCC?

engines/scc_proxy/lib/scc_proxy/engine.rb
rjschwei: I do not quite understand why this was resolved. Everywhere else in the RMT code base functions/methods related to de-registration of a system use `deregister`. A name change to `de_register`  makes it harder to find all the places in the code base where a de-registration happens. As such I'd urge you to reconsider this name change to help everyone else that may need to look for de-registration actions in the RMT code base.

bear454: Resolved in 6fa9a9ea.

engines/scc_proxy/lib/scc_proxy/engine.rb
bear454: Resolved in 6fa9a9ea.

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: according to the [API documentation](https://scc.suse.com/connect/v4/documentation#/systems/delete_systems), that's the response 
- code 204
- No content

so checking [Net::HTTPNoContent](https://docs.ruby-lang.org/en/master/Net/HTTPNoContent.html) seems appropiate to me

bear454 (approved):

digitaltom:
Please change engines/scc_proxy/lib/scc_proxy/engine.rb:327

```
- params['hostname'] = @system.system_information['hostname']
+ params['hostname'] = @system.hostname
```

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -369,13 +387,40 @@ def scc_deactivate_product
...
-            unless response.code_type == Net::HTTPOK
-              error = JSON.parse(response.body)
-              error['error'] = SccProxy.parse_error(error['error'], params[:token], params[:email]) if error['error'].include? 'json'
-              logger.info "Could not de-activate product '#{@product.friendly_name}', error: #{error['error']} #{response.code}"
-              raise ActionController::TranslatedError.new(error['error'])
+            handle_response(response)
+          elsif @system.hybrid? && @product.extension?
+            # check if product is on SCC and
+            # if it is -> de-activate it
+            scc_systems_activations = find_hybrid_product_on_scc(headers)
```
digitaltom, line 394: For me, this call to SCC fails with 'Unauthorized 401'. I assume that is, because the headers here are generated without providing the `Authorization: ` header from the original request, and the instance id.

digitaltom (changes requested):
See inline comment about authentication failure, and comment about fixing hostname parameter.

digitaltom (approved):

rjschwei (approved):

PR #1207, opened 2024-08-19
Raise an exception for unknown CSPs

jesusbv:
## Description

Instead of saving `nil` or empty value on the DB for `system_token` field, raise an exception if the CSP value from the metadata is unknown

## How to test 

Create a system with an unknown CSP, request should fail

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -68,7 +68,13 @@ module SccProxy
         nil,
         params['instance_data']
       )
-      instance_id_key = INSTANCE_ID_KEYS[params['hwinfo']['cloud_provider'].downcase.to_sym]
+      id_key = params['hwinfo']['cloud_provider'].downcase.to_sym
+      unless INSTANCE_ID_KEYS.key?(id_key)
+        error_message = 'Unknown cloud provider'
+        Rails.logger.error error_message
+        raise ActionController::TranslatedError.new(error_message)
+      end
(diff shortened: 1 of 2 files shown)


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -68,7 +68,13 @@ def get_instance_id(params)
         nil,
         params['instance_data']
       )
-      instance_id_key = INSTANCE_ID_KEYS[params['hwinfo']['cloud_provider'].downcase.to_sym]
+      id_key = params['hwinfo']['cloud_provider'].downcase.to_sym
+      unless INSTANCE_ID_KEYS.key?(id_key)
+        error_message = 'Unknown cloud provider'
+        Rails.logger.error error_message
+        raise ActionController::TranslatedError.new(error_message)
```
rjschwei, line 75: Why? I am unaware of any conversation at the business level where a decision was made that 3rd party CSPs that run their own update infrastructure are not allowed to offer BYOS pass through.

jesusbv: if the CSP is unknown we would not have a valid value for `system_token`

do we not care about that ?

rjschwei: I do not understand what the `system_token` has to do with this change. Nor what `system_token` has to do with now breaking functionality for 3rd party CSPs that previously probably worked when `disable_system_tokens: true` is set in the rmt.conf file.

PR #1208, opened 2024-08-20
Handle the empty credentials from the container runtime tools on the client

jesusbv:
## Description

When container runtime tools on the client do not send the proper credentials or an exception happens during the catalog search, handle the exception

## How to test 

Put wrong credentials on `/etc/containers/config.json` 
and run `podman search foo` and autentication required message is shown

```bash
ec2-user@ip-172-31-21-235:~> sudo /usr/bin/cloudguestregistryauth 
Credentials refreshed
ec2-user@ip-172-31-21-235:~> podman search foo
ERRO[0000] error getting search results from v2 endpoint "registry-ec2.susecloud.net": authentication required 
Error: 1 error occurred:
	* couldn't search registry "registry-ec2.susecloud.net": authentication required


ec2-user@ip-172-31-21-235:~> 
```

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -19,12 +19,23 @@ module Registry
     # Returns a Distribution Registry HTTP API V2 - compatible repository catalog as defined in
     # https://distribution.github.io/distribution/spec/api/#listing-repositories
     def catalog
+      raise 'Could not find system with current credentials' if @client&.account.nil?
+
       access_scope = AccessScope.parse('registry:catalog:*')
       repos = access_scope.allowed_paths(System.find_by(login: @client&.account))
       logger.debug("Returning #{repos.size} repos for client #{@client}")
 
       response.set_header('Docker-Distribution-Api-Version', REGISTRY_API_VERSION)
       render json: { repositories: repos }, status: :ok
+    rescue StandardError => e
+      logger.error("Could not search in catalog: #{e.message}")
+      error = if e.message.include?('credentials')
+                ActionController::TranslatedError.new(N_('Please, re-authenticate'))
+              else
+                ActionController::TranslatedError.new(N_(e.message))
+              end
+      error.status = :unauthorized
+      render json: { error: error.message }.to_json, status: :unauthorized
     end
 
     private
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -1,4 +1,6 @@
 module Registry
+  class RegistryAuthError < ArgumentError; end
+
   class RegistryController < Registry::ApplicationController
     REGISTRY_SERVICE = 'SUSE Linux OCI Registry'.freeze
     REGISTRY_API_VERSION = 'registry/2.0'.freeze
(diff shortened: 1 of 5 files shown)


rjschwei (approved):

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -19,12 +19,23 @@ def authorize
     # Returns a Distribution Registry HTTP API V2 - compatible repository catalog as defined in
     # https://distribution.github.io/distribution/spec/api/#listing-repositories
     def catalog
+      raise 'Could not find system with current credentials' if @client&.account.nil?
```
bear454, line 22: Can we classify this as a specific exception, instead of handling with the all-encompassing StandardError? I'm worried your handler is going to mask other exceptions.

bear454: The conditional could be simplified as `unless @client&.account`, or you could go for readability/consistency with `unless @client && @client.account`... but the `.nil?` seems superfluous.

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -19,12 +19,23 @@ def authorize
...
+
       access_scope = AccessScope.parse('registry:catalog:*')
       repos = access_scope.allowed_paths(System.find_by(login: @client&.account))
       logger.debug("Returning #{repos.size} repos for client #{@client}")
 
       response.set_header('Docker-Distribution-Api-Version', REGISTRY_API_VERSION)
       render json: { repositories: repos }, status: :ok
+    rescue StandardError => e
```
bear454, line 30: This would also catch any exception in the interim code - like a failure in JSON rendering. Doesn't guarantee the catalog search failed due to lack of authentication.

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -22,7 +22,6 @@ def authenticate_by_system_credentials(login, password)
       @account = login
       @auth_strategy = :system_credentials
     end
-    @auth_strategy
```
bear454, line 25: Why use the class variable at all?

The condition in line 9-10 could be changed to 

```
if authenticate_by_system_credentials(login, password)
```

and then you don't need to instantiate and pass around the class variable.

bear454 (changes requested):

engines/registry/app/controllers/registry/registry_controller.rb
bear454: Resolved in 3b832353.

engines/registry/app/controllers/registry/registry_controller.rb
bear454: Resolved in 3b832353.

engines/registry/app/models/registry/authenticated_client.rb
bear454: Resolved in 06cd9719

bear454 (approved):

engines/registry/app/models/registry/authenticated_client.rb
```diff
@@ -20,7 +17,6 @@ def authenticate_by_system_credentials(login, password)
     @systems = System.get_by_credentials(login, password)
     if @systems.present?
       @account = login
-      @auth_strategy = :system_credentials
```
jesusbv, line 23: good, now how do we know if `authenticate_by_system_credentials` succeeded or not ?

digitaltom (approved):
I don't have a test setup for this, but code looks ok.

PR #1209, opened 2024-08-20
Add header to SCC endpoint

jesusbv:
## Description

Add test to check the header is sent

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/registry/app/models/access_scope.rb
+++ b/engines/registry/app/models/access_scope.rb
@@ -83,21 +83,21 @@ class AccessScope
     access_policies_yml = YAML.safe_load(
       File.read(Rails.application.config.access_policies)
     )
-    active_products = system.activations.includes(:product).pluck(:product_class)
+    active_product_classes = system.activations.includes(:product).pluck(:product_class)
 
-    allowed_products = (active_products & access_policies_yml.keys)
-    allowed_glob_paths = access_policies_yml.values_at(*allowed_products).flatten
+    allowed_product_classes = (active_product_classes & access_policies_yml.keys)
+    allowed_glob_paths = access_policies_yml.values_at(*allowed_product_classes).flatten
 
-    if system && system.hybrid? && allowed_products.any? { |s| !Product.find_by(product_class: s).free? }.present?
+    if system && system.hybrid? && allowed_product_classes.any? { |s| !Product.find_by(product_class: s).free? }.present?
       auth_header = {
         Authorization: ActionController::HttpAuthentication::Basic.encode_credentials(system.login, system.password)
       }
       activation_state = SccProxy.scc_check_subscription_expiration(
-        auth_header, system.login, system.system_token, Rails.logger, system.proxy_byos_mode, allowed_products.first
-        )
+        auth_header, system.login, system.system_token, Rails.logger, system.proxy_byos_mode, allowed_product_classes.first
+      )
       unless activation_state[:is_active]
         Rails.logger.info(
-          "Access to #{allowed_products.first} from system #{system.login} denied: #{activation_state[:message]}"
+          "Access to #{allowed_product_classes.first} from system #{system.login} denied: #{activation_state[:message]}"
           )
         @allowed_paths = []
         return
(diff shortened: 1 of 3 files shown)


bear454 (approved):

rjschwei (approved):

rjschwei:
@digitaltom @felixsch if this looks good to you please merge and create a RMT 2.19 package for us to test.

jesusbv:
> @digitaltom @felixsch if this looks good to you please merge and create a RMT 2.19 package for us to test.

I understand the testing package should include #1204 (hopefully, once merged) as well

engines/registry/app/models/access_scope.rb
```diff
@@ -91,8 +91,11 @@ def allowed_paths(system = nil)
     if system && system.hybrid?
       allowed_product_class = allowed_products.detect { |s| s.downcase.include?('ltss') }
```
digitaltom, line 92: It looks weird to hardcode 'ltss' here. Why is this not compared with the actual product class?

jesusbv: I want to check if the product needs to check subscription with SCC
the product class is, i.e., `SLES15-SP4-LTSS-X86`  if Im not mistaken
so, checking for `ltss` to include any version or arch made sense to me

jesusbv: using a check for non free product

engines/registry/app/models/access_scope.rb
```diff
@@ -88,19 +88,19 @@ def allowed_paths(system = nil)
     allowed_products = (active_products & access_policies_yml.keys)
```
digitaltom, line 88:
```suggestion
    allowed_product_classes = (active_products & access_policies_yml.keys)
```

also above, it should read `active_product_classes`

engines/registry/app/models/access_scope.rb
```diff
@@ -88,19 +88,19 @@ def allowed_paths(system = nil)
...
     allowed_glob_paths = access_policies_yml.values_at(*allowed_products).flatten
 
-    if system && system.hybrid?
-      allowed_product_class = allowed_products.detect { |s| s.downcase.include?('ltss') }
-      if allowed_product_class.present?
-        activation_state = SccProxy.scc_check_subscription_expiration(
-          {}, system.login, system.system_token, Rails.logger, system.proxy_byos_mode, allowed_product_class
+    if system && system.hybrid? && allowed_products.any? { |s| !Product.find_by(product_class: s).free? }.present?
```
digitaltom, line 91: We'll need a comment here what this condition is doing. I guess it's something like `# if the system is hybrid, check if it's subscription is still valid for accessing images`
Why is this only checking for hybrid systems and not BYOS?

engines/registry/app/models/access_scope.rb
```diff
@@ -88,19 +88,19 @@ def allowed_paths(system = nil)
...
+      auth_header = {
+        Authorization: ActionController::HttpAuthentication::Basic.encode_credentials(system.login, system.password)
+      }
+      activation_state = SccProxy.scc_check_subscription_expiration(
+        auth_header, system.login, system.system_token, Rails.logger, system.proxy_byos_mode, allowed_products.first
```
digitaltom: It feels random to use the `.first` product here. Are we sure that the first product is actually the product that was activated with SCC?

jesusbv: we have a product class, one product class

in `allowed_product_classes` (which should be `allowed_product_class` will be always one product class, correct ?
like access policy has `- FOO: /a/b/c/**` and the paths can be 18 but `FOO` is one, yes it comes in an array but it is an array of size 1

jesusbv: Update the approach, to remove the non active non free extensions from the allowed paths

rjschwei:
> I cannot test run this, but code-wise looks ok now. Please also check how often that gets called and the performance of it, because the inline call to SCC can be expensive. Didn't we have a cache for this?

No there is no cache yet for checking if a system is registered and valid with SCC. That is a follow on project we can start once 2.19 is out.

PR #1213, opened 2024-08-23
Harden our shellout commands to work with paths that contain special characters

digitaltom:
Harden our shellout commands to work with paths that contain special characters. This fixes a regression that came up with 2.18, after moving the temp. download location into the repository directory. We have some repositories that contain '$' in their path (eg.: '/repo/$RCE/RES7-SUSE-Manager-Tools/x86_64/')

Those get correctly escaped now.

How to reproduce: 

```
bin/rmt-cli products enable res-manager-client/7/x86_64
bin/rmt-cli mirror product res-manager-client/7/x86_64
```


josegomezr (approved):
LGTM

PR #1210, opened 2024-08-21
Use text instead of string as it can be very limiting for paths lenght

Adnilson:
## Description

What: RMT is crashing when repo url is too long

Why: customers affected and reported a bug

* Related Issue / Ticket / Trello card: [<link reference>](https://trello.com/c/wClufm0e/3516-bug-rmt-mysql2error-data-too-long-for-column-localpath)

## How to test 

Find a repo url longer than 255 characters and see that no error is found according to path length

## Change Type

*Please select the correct option.*

- [ ] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [ ] I have reviewed my own code and believe that it's ready for an external review.
- [ ] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Review

Please check out our [review guidelines](https://github.com/SUSE/scc-docs/blob/master/team/workflow/code_review.md) 
and get in touch with the author to get a shared understanding of the change.

PR #1214, opened 2024-08-24
Revert "Use text instead of string as it can be very limiting for paths lenght"

digitaltom:
Reverts SUSE/rmt#1210


felixsch (approved):
lgtm :+1: 

Note this is reverted, to not have it in the 2.19 release not because of technical issues so far.

PR #1215, opened 2024-08-26
Use text instead of string as it can be limiting for paths length

digitaltom:


Adnilson:
**Note:** I tried to add this on trello but it did not format correctly.

Below I show some steps in order to see the problem related to the DB.

### Altered the table by optimizing:
  https://medium.com/@ashik.ck/how-to-re-index-mysql-database-after-restored-356788eeba62

  No change

### Used some variations of string (varchar)

1st mirror around 1m
- varchar(255) -> 15s
- varchar(2048), max limit URL ->  22s
- varchar(4096), max linux file path -> 25-34s

### When using text
1st mirror up to 6 minutes
- text -> 40s

These are set on the rails migration file as :string, limit: [limit]

This is a InnoDB limitation:
https://dba.stackexchange.com/a/222182
https://cpanel.net/blog/products/varchar-vs-text-for-mysql-databases/
https://kb.stablepoint.com/docs/choosing-between-varchar-and-text-in-mysql


commands used:
  `bin/rmt-cli products enable sles/15.6/x86_64`
  `bin/rmt-cli mirror product 2609,2626,2683,2618` # mirror SLE 15.6

Around 9GB

Total repositories -> 3358
Total downloaded_files -> 8375

digitaltom:
> 1st mirror around 1m
> 
>     * varchar(255) -> 15s
> 
>     * varchar(2048), max limit URL ->  22s
> 
>     * varchar(4096), max linux file path -> 25-34s

Having paths longer than 255 chars is really an exception, that's why it was only reported now. I'd suggest to set the column to varchar(512), and test for the performance impact. If it's low impact (expected), then I'd change it to 512.

db/schema.rb
```diff
@@ -104,7 +104,7 @@
     t.string "auth_token"
     t.boolean "installer_updates", default: false, null: false
     t.boolean "mirroring_enabled", default: false, null: false
-    t.string "local_path", null: false
+    t.text "local_path", limit: 512, null: false
```
digitaltom, line 107:
```suggestion
    t.string "local_path", limit: 512, null: false
```

digitaltom:
> 1st mirror around 1m
> 
>     * varchar(255) -> 15s
> 
>     * varchar(2048), max limit URL ->  22s
> 
>     * varchar(4096), max linux file path -> 25-34s

Do we have a timing result for varchar(512) to compare?

Adnilson:
> > 1st mirror around 1m
> > ```
> > * varchar(255) -> 15s
> > 
> > * varchar(2048), max limit URL ->  22s
> > 
> > * varchar(4096), max linux file path -> 25-34s
> > ```
> 
> Do we have a timing result for varchar(512) to compare?

@digitaltom :

It is basically the same time. Even 1 second less on average (from 13 to 12)

I tested this on a different network and in a docker container.

PR #1217, opened 2024-08-27
Update system from hybrid to payg if no more non free products active

jesusbv:
## Description

When deactivating a product on a hybrid system, if there are no more non free product active for that system on SCC, that system is no longer hybrid but payg and should get updated on the db

## How to test 

- start a SLES 15.4 PAYG instance
- activate LTSS (`registercloudguest -r FOO`)
- check is active `SUSEConnect -l` and see LTSS being active
- remove LTSS (`SUSEConnect -d -p LTSS`)
- the system must be `payg` / 1 in the db, in the RMT server, access the db, and run
```SQL 
select proxy_byos_mode from systems where login = 'SCC_that_system_login'
```
The value for that field should be 1

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -405,8 +405,8 @@ module SccProxy
               # it is OK to remove it from SCC
               response = SccProxy.deactivate_product_scc(auth, @product, @system.system_token)
               handle_response(response)
-              # if the system does not have more non free products activated on SCC
-              # system should turn to hybrid
+              # if the system does not have more products activated on SCC
+              # switch it back to payg
               @system.payg! if no_more_activations scc_systems_activations
             elsif result[:message].downcase.include?('unexpected error')
               raise ActionController::TranslatedError.new(result[:message])
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -407,7 +407,9 @@ module SccProxy
               handle_response(response)
               # if the system does not have more products activated on SCC
               # switch it back to payg
-              @system.payg! unless system_has_paid_extension? scc_systems_activations
+              # drop the just de-activated activation from the list to avoid another call to SCC
+              # and check if there is any product
+              @system.payg! if scc_systems_activations.reject! { |act| act['service']['product']['id'] == @product.id }.blank?
             elsif result[:message].downcase.include?('unexpected error')
               raise ActionController::TranslatedError.new(result[:message])
             end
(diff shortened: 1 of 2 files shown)


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -429,6 +432,15 @@ def handle_response(response)
             raise ActionController::TranslatedError.new(error['error'])
           end
         end
+
+        def no_more_activations(scc_systems_activations)
```
jesusbv, line 436: maybe rename to `no_more_activations?`

rjschwei: We end up having to to a negative test anyway either via name, as in `no_more_activations` or via a `not` condition in the test. As such I think it makes sense to name the method according to what we are actually looking for which is non-free extension. I'd suggest `product_has_paid_extension` or `product_has_paid_extension_active` or `system_has_paid_extension`, something along those lines.

rjschwei (approved):

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -405,6 +405,9 @@ def scc_deactivate_product
               # it is OK to remove it from SCC
               response = SccProxy.deactivate_product_scc(auth, @product, @system.system_token)
               handle_response(response)
+              # if the system does not have more non free products activated on SCC
+              # system should turn to hybrid
```
digitaltom, line 409:
```suggestion
              # if the system does not have more products activated on SCC
              # switch it back to payg
```

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -405,6 +405,9 @@ def scc_deactivate_product
...
+              @system.payg! if no_more_activations scc_systems_activations
```
digitaltom, line 410: Why is that `no_more_activations'`method needed, can't we just use the existing `find_hybrid_product_on_scc(request.headers).empty?`?

jesusbv: Because that would require a new request to SCC, the information to check is available without the request

The product successfully removed should be skipped during the check of that info, that's the difference with calling SCC again

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: OK, `system_has_paid_extension` as it's the thing object with extensions (products)

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -429,6 +432,15 @@ def handle_response(response)
...
         end
+
+        def system_has_paid_extension?(scc_systems_activations)
+          active_products_classes = scc_systems_activations.map do |act|
+            if act['status'].casecmp('active').zero? && act['service']['product']['product_class'] != @product.product_class
+              act['service']['product']['product_class']
+            end
+          end.flatten.compact
+          active_products_classes.present?
+        end
```
digitaltom, line 443: I don't think there's a need to check product classes here: 

```suggestion
        def system_has_paid_extension?(scc_systems_activations)
          # drop the just de-activated activation from the list to avoid another call to SCC
          scc_systems_activations.reject!{|act| act['service']['product']['id'] == @product.id }
          scc_systems_activations.present?
        end
```

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -392,27 +399,22 @@ def deregister_hybrid(auth)
         def scc_deactivate_product
           auth = request.headers['HTTP_AUTHORIZATION']
           if @system.byos? && @product[:product_type] != 'base'
```
digitaltom, line 401: What if system is byos, and the product is base, do we handle this in some other workflow?

PR #1235, opened 2024-10-03
Move error handling to the proper method 

jesusbv:
## Description

The method that gets the activations from SCC by making a request,
should be responsible to handle if the request failed

Use Rails.logger instead of carrying logger all over the methods

## How to test 

Activate a product on SCC,should work as expected

## Change Type

## Checklist

## Review


bear454 (approved):
It's impressive how much you were able to refactor with such a minor change to the test suite. Keep up the good work :+1:

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -201,14 +201,19 @@ def parse_error(error_message, token = nil, email = nil)
       error_message
     end
 
-    def get_scc_activations(headers, system_token, mode)
+    def get_scc_activations(headers, system_token, mode, login)
```
digitaltom, line 204: Why not using the `system` as a parameter here instead of it's attributes `system_token`, `mode`, `login`? Same for method `scc_check_subscription_expiration()`.

rjschwei (approved):

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: yes, thought about it, it required a bit more refactor that I thought to do in a different PR but
I think it could be done here too

digitaltom (approved):

PR #1241, opened 2024-10-29
Yabeda integration for prom metrics

ngetahun:
## Description

* Related Issue / Ticket / Trello card: https://trello.com/c/Mus496Nz/3617-repct-integrate-yabeda-for-prometheus-collectors

## How to test 

Start the rails server:
`> bin/rails s`
On a separate console, run:
`> curl localhost:4224/metrics`

Should see a text with metrics and values

## Change Type

## Checklist

- [x] I have documented the `MANUAL.md` file with any changes to the user experience.

## Review


Gemfile
```diff
@@ -32,6 +32,13 @@ gem 'repomd_parser', '~> 1.1.0'
 # Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
 # gem 'rack-cors'
 
+# Prometheus Exporter:
+gem 'yabeda'
+gem 'yabeda-rails'
+gem 'yabeda-puma-plugin'
+gem 'yabeda-sidekiq'
```
josegomezr, line 39: This one is not needed, rmt doesn't have sidekiq (yet)

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+return if Settings.scc.try(:metrics) && (Settings.scc.metrics.try(:enabled) && (!!Settings.scc.metrics.enabled  == false))
```
josegomezr, line 3: Maybe `Settings` supports `Hash#dig` ?

It would make it way cleaner to call it like:

```ruby
enabled = Settings.dig(:scc, :metrics, :enabled)
```

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,82 @@
...
+)
+
+# Configure yabeda
+Yabeda.configure do
+  assign_labels = lambda {
+    default_tag :environment, Rails.env
+    default_tag :application, 'rmt'
+  }
+
+  group :sidekiq, &assign_labels
```
josegomezr, line 22: This one is not needed

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,82 @@
...
+  group :rails, &assign_labels
+  group :rake, &assign_labels
```
josegomezr, line 24: For a PoC I'd first leave it until here (line 23). First collect all metrics that the framework gives you, and then incorporate more metrics later.

The reasoning for this is that the metric reporting is still a bit rudimentary, as an implementor you have to shift the context from the place of invocation to the event listener (for example, the one in line 57), and THEN account for the metric. While this is fine, it's not reload-safe, if you're going to rely on reloading this flow won't work.

Among other things this is sadly by design in rails, event subscriber registries are not reloadable.

config/puma.rb
```diff
@@ -36,3 +37,7 @@
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
+
+if Settings.scc.try(:metrics) && (Settings.scc.metrics.try(:enabled) && (!!Settings.scc.metrics.enabled  == false))
+  configure_prometheus!(self)
+end
```
josegomezr, line 43: Puma doesn't know about the `Setttings` object yet, this belongs to the application. Although it may work, this is because of the `Settings` class having default settings to operate.

If you further down in the app code change the default opts then you end up in inconsistent behavior, puma exclusively know about the default behavior but the app sees the altered behavior, although both use the same interface.

On puma config you should not interface with application internals because the application is yet to be loaded

package/obs/rmt.conf
```diff
@@ -12,6 +12,13 @@ scc:
   username:
   password:
   sync_systems: true
+  metrics:
+    enabled: false
+    interval: 60
+    job_name: yabeda
```
josegomezr, line 18:
```suggestion
    job_name: rmt-webserver
```

lib/yabeda/prometheus_patch.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
josegomezr, line 1: This file is not needed. We patch yabeda to allow support for a different groupping key when using prometheus push gateways.

Push gateways are a tool for ephemeral workloads (like cron jobs), RMT does not have this facilities well matured yet to incorporate this.

josegomezr (commented):
You're almost there! Left you some comments to trim a bit what's not used in RMT.

config/puma.rb
ngetahun: This is not true. `lib/rmt/config.rb` sets up `Settings` object.

josegomezr:
```
✗ bex puma 
        15: from /home/scc/.rvm/gems/ruby-2.5.9/bin/ruby_executable_hooks:22:in `<main>'
        14: from /home/scc/.rvm/gems/ruby-2.5.9/bin/ruby_executable_hooks:22:in `eval'
        13: from /home/scc/.rvm/gems/ruby-2.5.9/bin/puma:23:in `<main>'
        12: from /home/scc/.rvm/gems/ruby-2.5.9/bin/puma:23:in `load'
        11: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/bin/puma:8:in `<top (required)>'
        10: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/bin/puma:8:in `new'
         9: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/cli.rb:72:in `initialize'
         8: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/cli.rb:72:in `new'
         7: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/launcher.rb:60:in `initialize'
         6: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/configuration.rb:219:in `load'
         5: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/configuration.rb:219:in `each'
         4: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/configuration.rb:219:in `block in load'
         3: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/dsl.rb:87:in `_load_from'
         2: from /home/scc/.rvm/gems/ruby-2.5.9/gems/puma-5.6.8/lib/puma/dsl.rb:87:in `instance_eval'
         1: from config/puma.rb:1:in `_load_from'
config/puma.rb:1:in `require': cannot load such file -- rmt/config (LoadError)
```

As I said, puma doesn't know about this object yet because the application has not loaded.

When you run the webserver using `bin/rails server` the config files for puma are never loaded. The boot sequence first loads the `rails` framework, then it loads `rack` and then goes over `rackup` to start the server (based on puma or Webrick, nowadays is puma) and then the puma config may be loaded.

ngetahun: I understand what you mean but rmt server is meant to be run with `rails s`, look at the `.service` file, or even the entrypoint script for the docker image. I think that was to make the server options configurable through `rmt.{conf,yml}`.

josegomezr: Ach so... welp, that overrides completely my position then.

.github/workflows/lint-unit.yml
```diff
@@ -82,6 +82,16 @@ jobs:
...
+        
+      - name: Run PubCloud engines tests
+        run: |
+          bundle exec rake test:engines
+
+      - name: Run version checks
+        run: |
+          echo "::group::Version verification checks"
+          ruby ci/check-version-matches.rb
+          echo "::endgroup::"
```
digitaltom, line 94: Looks like these blocks are duplicate now?

.github/workflows/lint-unit.yml
```diff
@@ -92,3 +102,10 @@ jobs:
           echo "::group::Version verification checks"
           ruby ci/check-version-matches.rb
           echo "::endgroup::"
+         
+      - name: Export coverage report
+        uses: actions/upload-artifact@v4
+        if: always()
```
digitaltom, line 108: As @felixsch mentioned issue regarding storage of artifacts, how about storing the coverage report only in case of failure?

ngetahun: This is only for this PR. I will completely remove it before merging it.

felixsch (approved):
lgtm :+1: 

I'm not really sure what is the scope. The metric export now most information, I hope this is fine.

**How I reviewed this merge request:**

- Checked the code changes make sense to me
- Ran the branch and checked `/metrics` is populated (also in error case)

cheers,

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,71 @@
...
+ActiveSupport::Notifications.subscribe 'start_processing.action_controller' do |*args|
+  # Match the same event as Yabeda
+  event = Yabeda::Rails::Event.new(*args)
+
+  Yabeda.rails.started_requests_total.tap do |metric|
+    labels = event.labels.slice(*metric.tags)
+
+    metric.increment(labels, by: 1)
+  end
+end
```
josegomezr, line 55: Please remove this metric, it's a failed attempt to detect unusual behavior. At the end it ended up being closer to an `in-flight` request handler but not really :sweat_smile:

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,71 @@
...
+  if !!(event.payload[:exception] || event.payload[:exception_class])
+    success = 0
+  end
+
+  Yabeda.rake.task_duration_ms.set({ task_name: task_name }, event.duration)
+  Yabeda.rake.task_started_at.set({ task_name: task_name }, event.time / 1000.0)
+  Yabeda.rake.task_finished_at.set({ task_name: task_name }, event.end / 1000.0)
+
+  Yabeda.rake.task_exit_status.set({ task_name: task_name }, success)
+end
```
josegomezr, line 71: This along side the `:rake` group above is not needed for webserver metrics.

The two blocks are used in conjunction with a prometheus push gateway to achieve true instrumentation of ephemeral workloads [Jobs/CronJobs].

ngetahun:
[[REPCT] Integrate Yabeda for prometheus collectors](https://trello.com/c/Mus496Nz/3617-repct-integrate-yabeda-for-prometheus-collectors)

config/initializers/yabeda.rb
ngetahun: I will leave this in for smoke test purposes.

config/initializers/yabeda.rb
ngetahun: done

config/initializers/yabeda.rb
```diff
@@ -0,0 +1,37 @@
+# :nocov:
```
digitaltom, line 1: This should not be needed here, because you added `add_filter('config')` to the simplecov config. 
Also, using `:nocov:` needs to wrap the code, so this would be missing a line at the end of the file: https://github.com/simplecov-ruby/simplecov?tab=readme-ov-file#ignoringskipping-code

lib/rmt/config.rb
```diff
@@ -1,3 +1,4 @@
+# :nocov:
```
digitaltom, line 1: Same comment here, it should be enough to either skip coverage via `add_filter('lib/rmt/config.rb')` or the `:nocov:` lines.

digitaltom (approved):

PR #1247, opened 2024-11-14
Allow SLE Micro system to access free SLES repositories

jesusbv:
## Description

A SLE Micro system must have access to all free SLES repositories

This Fixes bsc#1230419

* Related Issue / Ticket / Trello card: https://bugzilla.suse.com/show_bug.cgi?id=1230419

## How to test 

On a SLE Micro PAYG instance, configured according to the documentation to use Podman,
running `zypper lr` inside the container should show `SLES` repositories

On a SUMA 5.0 instance, activating SUMA 4.3 should also activate the recommended SLES repositories 

## Change Type

## Checklist

## Review

### DIFF
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -12,7 +12,13 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
   end
 
   def show
-    if @system.products.include? @product
+    if @product.identifier.casecmp('sles').zero?
+      # if system has SLE Micro
+      # it should access to SLES products
+      sle_micro = @system.products.any? { |p| p.identifier.downcase.include?('sle-micro') }
+      sle_micro_same_arch = @system.products.pluck(:arch).include?(@product.arch) if sle_micro
+    end
+    if @system.products.include?(@product) || sle_micro_same_arch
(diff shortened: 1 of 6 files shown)


digitaltom:
You are defining, based on the product identifier of the registered base product, to infer which other products / repositories the system can use. That works, but is a lot of manual setup. 

Maybe there is an automatic way to find out that information, in the way how SCC is doing this: 
In SCC, the used subscription of the base product has a list of product classes. The system then has 'permission' to use all products that have a product class of that list. Example: a SUMA subscription includes the product classes of SLES (7261), SLE Micro (MICROOS-X86). 

We have a manual config to build that connection for test subscriptions: https://github.com/SUSE/happy-customer/blob/master/glue/config/multi_product_classes.yml

Just an idea, maybe this is a less manual, and more aligned with SCC way how to define access permission.

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -12,7 +12,13 @@ def activate
   end
 
   def show
-    if @system.products.include? @product
+    if @product.identifier.casecmp('sles').zero?
```
bear454, line 15: Is there no positive test here we can do? It feels inefficient to say "if it's not sles maybe it's sle-micro".

PR #1248, opened 2024-11-21
Azure basic images do not get access to LTSS

jesusbv:
## Description

Azure basic images must not get access to LTSS product

## How to test 

Running an Azure basic image, trying to activate LTSS should fail

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/instance_verification/lib/instance_verification/providers/example.rb
+++ b/engines/instance_verification/lib/instance_verification/providers/example.rb
@@ -50,7 +50,9 @@ class InstanceVerification::Providers::Example < InstanceVerification::ProviderB
     return true if (identifier.casecmp('sles_sap').zero? && SLES4SAP_PRODUCT_IDENTIFIER.include?(instance_billing_info[:marketplace_code]))
   end
 
-  def basic?
-    false
+  def allowed_extension?
+    # method to check if a product (extension) meet the criteria
+    # to be acivated on SCC or not, i.e. LTSS in Azure Basic VM
+    true
   end
 end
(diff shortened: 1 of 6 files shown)


engines/instance_verification/lib/instance_verification/providers/example.rb
```diff
@@ -49,4 +49,8 @@ def payg_billing_code?(iid, identifier)
     return true if (identifier.casecmp('sles').zero? && instance_billing_info[:billing_product] == SLES_PRODUCT_IDENTIFIER)
     return true if (identifier.casecmp('sles_sap').zero? && SLES4SAP_PRODUCT_IDENTIFIER.include?(instance_billing_info[:marketplace_code]))
   end
+
+  def basic?
+    false
+  end
```
rjschwei, line 55: We should probably have some comment here explaining what this is intended to do. The term  `basic` means something to us (SUSE) because we publish so called `basic` images in Azure and those images have no support entitlements but this meaning is no obvious to any 3rd party CSP using the Public Cloud features in  that are part of RMT.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -372,6 +373,15 @@ def has_no_regcode?(auth_header)
         protected
 
         def scc_activate_product
+          if (@system.system_information &&
+              JSON.parse(@system.system_information)['cloud_provider'].casecmp('microsoft').zero? &&
```
rjschwei, line 377: Should not be CSP specific and CSP could have an offer the meets the definitions that make us exclude ltss registration.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -372,6 +373,15 @@ def has_no_regcode?(auth_header)
...
+              @product.product_class.downcase.include?('ltss') &&
+              InstanceVerification.provider.new(logger, request, nil, @system.instance_data).basic?
```
rjschwei, line 379: Thinking ahead this should maybe be more generic, i.e. the verification code should probably have `allows_ltss?` rather than `basic`. If at some point other addons are made available then we can simply extend the logic to be product specific rather than having the function name a specific designation for the listing.

rjschwei (changes requested):

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: Fair point, I understood it should be only for Microsoft or CSP specific

That means I will update the other two verification codebases

rjschwei (approved):

bear454 (approved):

PR #1249, opened 2024-11-25
Send the instance ID in the token for Azure

jesusbv:
## Description

When activating an extension, i.e. LTSS in Azure,
the header must contain the the instance identifier

This Fixes bsc#1233314

* Related Issue / Ticket / Trello card: https://bugzilla.suse.com/show_bug.cgi?id=1233314

## How to test 

on Azure, registering LTSS extension should work

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -25,12 +25,6 @@ NET_HTTP_ERRORS = [
   Net::HTTPRetriableError
 ].freeze
 
-INSTANCE_ID_KEYS = {
-  amazon: 'instanceId',
-  google: 'instance_id',
-  microsoft: 'vmId'
-}.freeze
-
 # rubocop:disable Metrics/ModuleLength
 module SccProxy
   class << self
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -68,9 +68,10 @@ module SccProxy
         nil,
         params['instance_data']
       )
-      instance_id_key = INSTANCE_ID_KEYS[params['hwinfo']['cloud_provider'].downcase.to_sym]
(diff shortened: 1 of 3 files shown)


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -68,9 +68,10 @@ def get_instance_id(params)
         nil,
         params['instance_data']
       )
-      instance_id_key = INSTANCE_ID_KEYS[params['hwinfo']['cloud_provider'].downcase.to_sym]
-      iid = verification_provider.parse_instance_data
-      iid[instance_id_key]
+      csp = params['hwinfo']['cloud_provider'].downcase
+      instance_id_key = INSTANCE_ID_KEYS[csp.to_sym]
+      instance_data = verification_provider.parse_instance_data
+      csp.casecmp('microsoft').zero? ? instance_data['attestedData'][instance_id_key] : instance_data[instance_id_key]
```
digitaltom, line 74: Could this be encapsulated as a method in the cloud provider specific instance verification class?

rjschwei: I think we should keep CSP specific cases out of the generic code. If we need to pass a key that is basically meaningless to SCC but used to allow activation for certain products then that key should be provided by `verification_provider.get_identifier`. With this the generic RMT code does the same thing for every provider and the provider specific implementation about how to get this identifier is hidden. We choose to use the instance identifier, a 3rd party CSP in their implementation of the verification code could choose some other identifier.

rjschwei (changes requested):

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: good point

rjschwei (approved):

digitaltom (approved):

PR #1254, opened 2024-12-02
skip rotation for read Apis

paragjain0910:
## Description

This PR targets to only rotate the token on calls to  PUT/POST/DELETE in system namespace

* Related Issue / Ticket / Trello card: https://trello.com/c/bimNtIHi/3647-repct-skip-system-token-rotation-in-read-only-apis

## How to test 

1. Show that calls to systems/ APIs don't rotate the token, only to [PUT/POST /systems/*](https://scc.suse.com/connect/v4/documentation#/systems/put_systems):

2. PUT /systems , PUT/POST/DELETE /systems/products

3. POST /systems/products/synchronize

4. Show that the current token gets returned in the response header

5. ~~Have it well documented in swagger~~ (RMT has no swagger)

```
> docker run --network=host --rm -ti registry.suse.com/suse/sle15:15.5 /bin/bash
$ zypper rm -y container-suseconnect
$ zypper in -y suseconnect-ng
$ suseconnect --url http://localhost:4224
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token got set
$ suseconnect --keepalive
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token updated
$ suseconnect -d -p sle-module-python3/15.5/x86_64
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token updated
$ suseconnect --rollback
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token updated
$ suseconnect --list-extensions
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token _not_ updated
$ suseconnect --status
$ cat /etc/zypp/credentials.d/SCCcredentials # see that token _not_ updated
 /systems/activations doesn't rotate token: 
>curl -u <credentials from system above> -H "System-Token: <token from system above>" -I "http://localhost:4224/connect/systems/activations"

```

## Change Type

- [ 8] **New Feature** (a non-breaking change which adds new functionality)

## Review

### DIFF
--- a/app/controllers/api/connect/v3/systems/products_controller.rb
+++ b/app/controllers/api/connect/v3/systems/products_controller.rb
@@ -5,6 +5,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
   before_action :check_product_service_and_repositories, only: %i[show activate]
   before_action :load_subscription, only: %i[activate upgrade]
   before_action :check_base_product_dependencies, only: %i[activate upgrade show]
+  after_action :refresh_system_token, only: %i[activate upgrade], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
 
   def activate
     create_product_activation
--- a/app/controllers/api/connect/v4/systems/products_controller.rb
+++ b/app/controllers/api/connect/v4/systems/products_controller.rb
@@ -1,5 +1,6 @@
 class Api::Connect::V4::Systems::ProductsController < Api::Connect::V3::Systems::ProductsController
 
+  after_action :refresh_system_token, only: %i[synchronize destroy], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
   def destroy
     if @product.base?
       raise ActionController::TranslatedError.new(N_('The product "%s" is a base product and cannot be deactivated'), @product.name)
--- a/app/controllers/api/connect/v4/systems/products_controller.rb
+++ b/app/controllers/api/connect/v4/systems/products_controller.rb
@@ -1,6 +1,6 @@
 class Api::Connect::V4::Systems::ProductsController < Api::Connect::V3::Systems::ProductsController
 
-  after_action :refresh_system_token, only: %i[synchronize destroy], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
+  after_action :refresh_system_token, only: %i[activate upgrade synchronize destroy], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
   def destroy
     if @product.base?
       raise ActionController::TranslatedError.new(N_('The product "%s" is a base product and cannot be deactivated'), @product.name)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,7 +24,7 @@ class ApplicationController < ActionController::API
         # that supports this feature.
         if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER)
           @system.update(last_seen_at: Time.zone.now)
-          headers[SYSTEM_TOKEN_HEADER] = @system.system_token
+          system_token_header
         # only update last_seen_at each 3 minutes,
         # so that a system that calls SCC every second doesn't write + lock the database row
         elsif !@system.last_seen_at || @system.last_seen_at < 3.minutes.ago
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,11 +21,11 @@ class ApplicationController < ActionController::API
         update_user_agent
 
         # If SYSTEM_TOKEN_HEADER is present, RMT assumes the client uses a SUSEConnect version
-        # that supports this feature. In this case, refresh the token and include it in the response.
+        # that supports this feature.
         if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER)
-          @system.update(last_seen_at: Time.zone.now, system_token: SecureRandom.uuid)
+          @system.update(last_seen_at: Time.zone.now)
           headers[SYSTEM_TOKEN_HEADER] = @system.system_token
-        # only update last_seen_at each 3 minutes,
+          # only update last_seen_at each 3 minutes,
         # so that a system that calls SCC every second doesn't write + lock the database row
         elsif !@system.last_seen_at || @system.last_seen_at < 3.minutes.ago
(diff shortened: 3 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -21,11 +21,11 @@ def authenticate_system(skip_on_duplicated: false)
         update_user_agent
 
         # If SYSTEM_TOKEN_HEADER is present, RMT assumes the client uses a SUSEConnect version
-        # that supports this feature. In this case, refresh the token and include it in the response.
+        # that supports this feature.
         if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER)
-          @system.update(last_seen_at: Time.zone.now, system_token: SecureRandom.uuid)
+          @system.update(last_seen_at: Time.zone.now)
           headers[SYSTEM_TOKEN_HEADER] = @system.system_token
```
digitaltom, line 27: could use the new method `system_token_header` here?

spec/requests/api/connect/base_controller_spec.rb
```diff
@@ -55,15 +55,6 @@ def require_product
     end
   end
 
-  shared_examples 'updates the system token' do
```
digitaltom, line 58: I'd suggest to move this to 'PUT /systems' for example, to make sure the token gets rotated correctly there.

app/controllers/application_controller.rb
digitaltom: Shouldn't all of this go to `refresh_system_token`, so that the token only gets rotated `if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER) == true`?

paragjain0910: do you mean i should remove the complete if condition from the authenticate_system function from application_controller.rb
change from
   ```
     if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER)
          @system.update(last_seen_at: Time.zone.now)
          headers[SYSTEM_TOKEN_HEADER] = @system.system_token
          # only update last_seen_at each 3 minutes,
        # so that a system that calls SCC every second doesn't write + lock the database row
        elsif !@system.last_seen_at || @system.last_seen_at < 3.minutes.ago
          @system.touch(:last_seen_at)
        end
```
to
```
        if !@system.last_seen_at || @system.last_seen_at < 3.minutes.ago
          @system.touch(:last_seen_at)
        end
```
add 2)  the below in base_controller.rb so that ( only adding the token in the header )
`  after_action :system_token_header, if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
`
Also, I think, the current way will also work which is already implemented in the PR. without adding after-action in base-controller
as we need to send the token in response headers for all the api's ( read + write ) I think. so for the system namespace controllers, we are always calling authenticate_system function in before-action.

    ```
    if system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER)
          @system.update(last_seen_at: Time.zone.now)
          headers[SYSTEM_TOKEN_HEADER] = @system.system_token
          # only update last_seen_at each 3 minutes,
        # so that a system that calls SCC every second doesn't write + lock the database row
        elsif !@system.last_seen_at || @system.last_seen_at < 3.minutes.ago
          @system.touch(:last_seen_at)
        end
```

digitaltom: You're right, we always should return the `SYSTEM_TOKEN_HEADER`, when tokens are enabled.

app/controllers/api/connect/v4/systems/products_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Api::Connect::V4::Systems::ProductsController < Api::Connect::V3::Systems::ProductsController
 
+  after_action :refresh_system_token, only: %i[synchronize destroy], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
```
digitaltom, line 3:
```suggestion
  after_action :refresh_system_token, only: %i[activate upgrade synchronize destroy], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
```

spec/requests/api/connect/base_controller_spec.rb
paragjain0910: I have added test cases for read and write api's both.

app/controllers/api/connect/v4/systems/products_controller.rb
paragjain0910: v4 product controller has only two actions synchronize and destroy and has after_action to refresh for these two actions 
v3 product controller has only activate and upgrade actions. and has after_action to refresh for these two actions 
V4 inherits v3
so I thought, we don't need to add it in v4.

digitaltom: One would think so, but it seems the `after_action` definition in the V4 class overwrites the one from V3 and needs to include `activate upgrade`. I tested this with the test steps from the pr description, and it fails without this change.

app/controllers/api/connect/v3/systems/products_controller.rb
```diff
@@ -5,6 +5,7 @@ class Api::Connect::V3::Systems::ProductsController < Api::Connect::BaseControll
   before_action :check_product_service_and_repositories, only: %i[show activate]
   before_action :load_subscription, only: %i[activate upgrade]
   before_action :check_base_product_dependencies, only: %i[activate upgrade show]
+  after_action :refresh_system_token, only: %i[activate upgrade], if: -> { request.headers.key?(SYSTEM_TOKEN_HEADER) }
```
digitaltom, line 8: All the `if` conditions for the `refresh_system_token` after_action need a check for `system_tokens_enabled?`, like: 

```suggestion
  after_action :refresh_system_token, only: %i[activate upgrade], if: -> { system_tokens_enabled? && request.headers.key?(SYSTEM_TOKEN_HEADER) }
```

or we include the `system_tokens_enabled?` check into `refresh_system_token()`.

paragjain0910: it is better to add the check in the refresh_system_token. so added the check.

app/controllers/api/connect/v4/systems/products_controller.rb
paragjain0910: Got it. Added activate and upgrade actions in the v4 product controller

digitaltom (approved):
Good work! Could you add a line to `package/obs/rmt-server.changes` regarding this change? Like "Skip system token rotation in read-only APIs".

PR #1257, opened 2024-12-11
Refactor SUSE repo cleanup and local path generation logic supporting Akamai CDN

paragjain0910:
## Description

- Refined `make_local_path` to handle URLs from both SUSE domains (`https://updates.suse.com%` and `https://dl.suse.com%`) 
  - The method is called during repository creation and update in rmt.
  - Some repository URLs contain `/repo` in the path, such as `https://updates.suse.com/repo/$RCE/SLES10-SP4-LTSS-Updates/sles-10-x86_64/` in SCC. When synced and mirrored, these URLs would lead to incorrect directory paths locally.
  - For example, the URL `https://updates.suse.com/repo/$RCE/...` would be updated to `https://dl.suse.com/repo/$RCE/...`. Without this fix, mirrored repositories would incorrectly create local paths like `/repo/repo/...`.
  - The local path generation now correctly removes the `/repo` prefix, ensuring the correct local path structure.

- Updated `remove_suse_repos_without_tokens!` to delete repositories with URLs from both the old (`https://updates.suse.com%`) and new (`https://dl.suse.com%`) CDN domains when `auth_token` is `nil`.
  - This method is called during the sync and import workflows. Previously, only repositories with URLs starting with `https://updates.suse.com%` were cleared.
  - After the migration to the new CDN, repositories with URLs starting with `https://dl.suse.com%` are now also deleted when lacking an authentication token.

* Related Issue / Ticket / Trello card:  

https://trello.com/c/YFbHJEkq/3669-edgxit-test-migration-to-new-cdn-domain-auth-token-with-rmt

## Review

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -21,6 +21,11 @@ class Repository < ApplicationRecord
   before_destroy :ensure_destroy_possible
 
   class << self
+
+    def remove_suse_repos_without_tokens!
+      where(auth_token: nil).where("external_url LIKE '%.suse.com%'").where(installer_updates: 0).delete_all
+    end
+
     # Mangles remote repo URL to make a nicer local path, see specs for examples
     def make_local_path(url)
       uri = URI(url)
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -21,18 +21,12 @@ class Repository < ApplicationRecord
   before_destroy :ensure_destroy_possible
 
   class << self
-
-    def remove_suse_repos_without_tokens!
-      where(auth_token: nil)
-        .where("external_url LIKE 'https://updates.suse.com%' OR external_url LIKE 'https://dl.suse.com%'")
-        .delete_all
-    end
-
     # Mangles remote repo URL to make a nicer local path, see specs for examples
     def make_local_path(url)
       uri = URI(url)
       path = uri.path.to_s
-      path.gsub!(%r{^/repo}, '') if (uri.hostname == 'updates.suse.com' || uri.hostname == 'dl.suse.com')
+      # drop '/repo' from SLE11 paths, to avoid double /repo/repo in local storage path.
+      path.gsub!(%r{^/repo/\$RCE/}, '/$RCE/')
       (path == '') ? '/' : path
     end
 
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -23,7 +23,7 @@ class Repository < ApplicationRecord
   class << self
 
     def remove_suse_repos_without_tokens!
-      where(auth_token: nil).where("external_url LIKE '%.suse.com%'").where(installer_updates: 0).delete_all
+      where(auth_token: nil).where("external_url LIKE '%.suse.com%'").where(installer_updates: 0).where.not(scc_id: nil).delete_all
     end
 
     # Mangles remote repo URL to make a nicer local path, see specs for examples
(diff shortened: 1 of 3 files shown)


lagartoflojo:
Do we need to somehow handle the installer-updates.suse.com domain (which will transition to dl-public.suse.com)?
That domain hosts files that don't require an auth token.

digitaltom:
> Do we need to somehow handle the installer-updates.suse.com domain (which will transition to dl-public.suse.com)? That domain hosts files that don't require an auth token.

There's nothing specific about installer-updates.suse.com in RMT, should be good.

app/models/repository.rb
```diff
@@ -23,14 +23,16 @@ class Repository < ApplicationRecord
...
+        .where("external_url LIKE 'https://updates.suse.com%' OR external_url LIKE 'https://dl.suse.com%'")
+        .delete_all
     end
 
     # Mangles remote repo URL to make a nicer local path, see specs for examples
     def make_local_path(url)
       uri = URI(url)
       path = uri.path.to_s
-      path.gsub!(%r{^/repo}, '') if (uri.hostname == 'updates.suse.com')
+      path.gsub!(%r{^/repo}, '') if (uri.hostname == 'updates.suse.com' || uri.hostname == 'dl.suse.com')
```
digitaltom, line 35: I'd suggest to do

```suggestion
      path.gsub!(%r{^/repo/\$RCE/}, '/\$RCE/') # drop '/repo' from SLE11 paths, to avoid double /repo/repo in local storage path.
```
to avoid further hostname hardcoding.

app/models/repository.rb
```diff
@@ -23,14 +23,16 @@ class Repository < ApplicationRecord
   class << self
 
     def remove_suse_repos_without_tokens!
-      where(auth_token: nil).where('external_url LIKE ?', 'https://updates.suse.com%').delete_all
+      where(auth_token: nil)
+        .where("external_url LIKE 'https://updates.suse.com%' OR external_url LIKE 'https://dl.suse.com%'")
+        .delete_all
```
digitaltom, line 28: Let's research a bit why this gets done at all here and if it could possibly get dropped.

paragjain0910: Removed the function as rmt would not have any repos from updates.suse.com without auth-token

app/models/repository.rb
paragjain0910: path.gsub!(%r{^/repo/\$RCE/}, '/$RCE/')

added the above.

ngetahun:
[[EDGXIT] Test migration to new CDN domain + auth token with RMT](https://trello.com/c/YFbHJEkq/3669-edgxit-test-migration-to-new-cdn-domain-auth-token-with-rmt)

paragjain0910:
The following are the different kind of repos that exists in scc.suse.com.

| **Source URL**                        | **Number of Repos** |
|---------------------------------------|---------------------|
| https://updates.suse.com              | 6025                |
| https://download.opensuse.org/        | 7                   |
| http://download.opensuse.org/         | 30                  |
| http://www.broadcom.com               | 1                   |
| https://public.dhe.ibm.com            | 4                   |
| http://www.lsi.com                    | 1                   |
| https://developer.download.nvidia.com | 2                   |
| http://download.nvidia.com            | 8                   |
| https://download.nvidia.com           | 10                  |
| https://cdn.opensuse.org              | 3                   |
| https://installer-updates.suse.com    | 54                  |
| **Total**                             | **6145**            |

app/models/repository.rb
paragjain0910: done and removed dependency from domain names

digitaltom (approved):
Please add a changelog entry, stating that RMT is enabled to handle the new dl.suse.com CDN domain.

PR #1261, opened 2024-12-17
Rmt cleanup old metadata

paragjain0910:
## Description

This task is for cleaning up old metadata files after repository updates in the CDN.

https://trello.com/c/SznwZfr3/3662-l3-regression-rmt-doesnt-clean-up-metadata

## How to test ( locally )

```
1) Enable a repository using the command:  ` rmt-cli repos enable <repo-id>`
2) Run the mirror command:  ` rmt-cli mirror`
3) Rename any metadata file in the repodata directory (e.g., rmt/public/repo/SUSE/Products/SLE-Product-SLES/15-SP6/ppc64le/product/repodata).
4) Run the mirror command again: `   rmt-cli mirror`

Check whether it will remove the renamed metadata file from the repodata directory and download the updated one from the CDN.
```
## Change Type

- [* ] **Bug Fix** (a non-breaking change which fixes an issue)

### DIFF
--- a/lib/rmt/mirror/base.rb
+++ b/lib/rmt/mirror/base.rb
@@ -155,9 +155,9 @@ class RMT::Mirror::Base
     logger.debug("Can not remove stale metadata directory: #{e}")
   end
 
-  def move_files(glob:, destination:)
+  def move_files(glob:, destination:, clean_before: true)
     FileUtils.mkpath(destination) unless Dir.exist?(destination)
-    FileUtils.rm_rf(Dir.glob(File.join(destination, '*')))
+    FileUtils.rm_rf(Dir.glob(File.join(destination, '*'))) if clean_before
     FileUtils.mv(Dir.glob(glob), destination, force: true)
   rescue StandardError => e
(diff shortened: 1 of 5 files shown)


lib/rmt/mirror/base.rb
```diff
@@ -157,6 +157,7 @@ def cleanup_stale_metadata
 
   def move_files(glob:, destination:)
     FileUtils.mkpath(destination) unless Dir.exist?(destination)
+    FileUtils.rm_rf(Dir.glob(File.join(destination, '*')))
```
felixsch, line 160: Can we make this a conditional `clean_before`, so when running we need to set it explicitly. This would make the behavior clear, that the directory gets cleaned before moving.

paragjain0910: added the conditional parameter clean_before with default value to true.

Adnilson (approved):

felixsch (approved):
lgtm :+1:

**How I reviewed this pull request:**

- Checked for changes I want to suggest
- Ran the branch and mirrored repository:
  - Renamed a file to simulate updated metadata
  - Checked the files get cleaned
  - Remirrored and checked again

PR #1264, opened 2024-12-23
Configure SUMA product tree base URL

lcaparroz:
## Description

Allow user to configure the SUMA product tree base URL in RMT's config file `mirroring.suma_product_tree_base_url` attribute. The configured value takes precedence over the default (https://scc.suse.com/suma/).

* Related Issue / Ticket / Trello card: https://trello.com/c/Fv5zbDoj/3694-rgs-configure-suma-product-tree-base-url-in-rmt

## How to test 

On your local setup, change the attribute `mirroring.suma_product_tree_base_url` on `config/rmt.local.yml` to point to a custom URL (the URL must be reachable and respond to `<custom URL>/product_tree.json`, e.g.:

```yaml
mirroring:
  mirror_src: false
  verify_rpm_checksums: false
  dedup_method: hardlink
  suma_product_tree_base_url: <custom URL>/suma/
```

Then, run `bin/rmt-cli mirror`, and verify that the log output shows that the mirroring process has finished without any errors regarding the SUMA product tree.

## Change Type

## Checklist

## Review

### DIFF
--- a/lib/rmt/mirror/suma_product_tree.rb
+++ b/lib/rmt/mirror/suma_product_tree.rb
@@ -3,9 +3,9 @@ class RMT::Mirror::SumaProductTree
 
   attr_reader :mirroring_base_dir, :url, :logger
 
-  def initialize(logger:, mirroring_base_dir:, url: FILE_URL)
+  def initialize(logger:, mirroring_base_dir:, url: nil)
     @mirroring_base_dir = mirroring_base_dir
-    @url = url
+    @url = url || Settings.try(:mirroring).try(:suma_product_tree_base_url) || FILE_URL
     @logger = logger
   end
 
--- a/config/rmt.yml
+++ b/config/rmt.yml
@@ -31,6 +31,7 @@ scc:
 mirroring:
   mirror_src: false
   dedup_method: hardlink
+  suma_product_tree_base_url: <%= ENV['SUMA_PRODUCT_TREE_BASE_URL'] %>
 
 http_client:
   verbose: false
--- a/lib/rmt.rb
+++ b/lib/rmt.rb
@@ -1,5 +1,5 @@
 module RMT
-  VERSION ||= '2.20'.freeze
+  VERSION ||= '2.21'.freeze
 
   DEFAULT_USER = '_rmt'.freeze
   DEFAULT_GROUP = 'nginx'.freeze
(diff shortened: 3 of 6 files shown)


Adnilson (approved):

lib/rmt/mirror/suma_product_tree.rb
```diff
@@ -3,9 +3,9 @@ class RMT::Mirror::SumaProductTree
 
   attr_reader :mirroring_base_dir, :url, :logger
 
-  def initialize(logger:, mirroring_base_dir:, url: FILE_URL)
+  def initialize(logger:, mirroring_base_dir:, url: nil)
     @mirroring_base_dir = mirroring_base_dir
-    @url = url
+    @url = url || Settings.try(:mirroring).try(:suma_product_tree_base_url) || FILE_URL
```
digitaltom, line 8: Is the url to load the product tree from different than the scc host from the config?

Adnilson: That is a good question.

Wouldn't the `config/rmt.yml` file have something like this below?
`
mirroring:
  suma_product_tree_base_url: <%= ENV['SUMA_URL'] %>
`

digitaltom:
@lcaparroz @Adnilson My comment was resolved here without proper discussion I think. 
What I was asking is: Why do we need an additional config value for the `SUMA_PRODUCT_TREE_BASE_URL`, when we already have `SCC_HOST`? I think the problem we want to solve here is, the hardcoded `FILE_URL` in `lib/rmt/mirror/suma_product_tree.rb` not using `SCC_HOST` (see also https://bugzilla.suse.com/show_bug.cgi?id=1234844). 
What's the use case of configuring an extra host in `SUMA_PRODUCT_TREE_BASE_URL`?

Adnilson:
> @lcaparroz @Adnilson My comment was resolved here without proper discussion I think. What I was asking is: Why do we need an additional config value for the `SUMA_PRODUCT_TREE_BASE_URL`, when we already have `SCC_HOST`? I think the problem we want to solve here is, the hardcoded `FILE_URL` in `lib/rmt/mirror/suma_product_tree.rb` not using `SCC_HOST` (see also https://bugzilla.suse.com/show_bug.cgi?id=1234844). What's the use case of configuring an extra host in `SUMA_PRODUCT_TREE_BASE_URL`?

My bad, was in a rush to get this delivered. Thought it was a good idea to have this variable there.

PR #1270, opened 2025-01-09
Update Zypper auth check

jesusbv:
## Description

When a hybrid system runs zypper command, the check
needs to take care of the subscription verification

Currently, the check lacks paid extension ID, resulting in
an unsuccessful verification

This change fixes that

If system is hybrid, and the path being accessed belongs to a paid extension,
we check that the paid extension subscription is active

if the repository path belongs to a free repository or no matches with paid extensions,
no need to check the subscription

if ZypperAuth handle_scc_subscription method gets called without a product id,
the method will check _all_ non free products suscriptions to be active, if any

This fixes bsc#1230157
* Related Issue / Ticket / Trello card: https://bugzilla.suse.com/show_bug.cgi?id=1230157

## How to test 

Start an Azure instance, register LTSS, wait > 20 min for the cache to get expired
run `zypper ref` should work

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/zypper_auth/lib/zypper_auth/engine.rb
+++ b/engines/zypper_auth/lib/zypper_auth/engine.rb
@@ -6,7 +6,7 @@ module ZypperAuth
       Thread.current[:logger]
     end
 
-    def verify_instance(request, logger, system, params_product_id = nil)
+    def verify_instance(request, logger, system)
       return false unless request.headers['X-Instance-Data']
 
       base_product = system.products.find_by(product_type: 'base')
--- a/engines/zypper_auth/lib/zypper_auth/engine.rb
+++ b/engines/zypper_auth/lib/zypper_auth/engine.rb
@@ -51,9 +51,22 @@ module ZypperAuth
     end
 
     def handle_scc_subscription(request, system, verification_provider, params_product_id = nil)
-      product_class = Product.find_by(id: params_product_id).product_class if params_product_id.present?
-      result = SccProxy.scc_check_subscription_expiration(request.headers, system, product_class)
-      return true if result[:is_active]
+      if params_product_id.present?
+        product_class = Product.find_by(id: params_product_id).product_class
+        result = SccProxy.scc_check_subscription_expiration(request.headers, system, product_class)
+        return true if result[:is_active]
+      else
+        # no product id provided
+        # check all non free extensions subscriptions with SCC, if any
+        paid_extensions = system.products.select { |prod| prod if !prod.free && prod.product_type == 'extension' }
+        return true if paid_extensions.empty?
+
+        all_active = paid_extensions.all? do |paid_extension|
+          result = SccProxy.scc_check_subscription_expiration(request.headers, system, paid_extension.product_class)
+          result[:is_active]
+        end
+        return true if all_active
+      end
 
       ZypperAuth.zypper_auth_message(request, system, verification_provider, result[:message])
       false
@@ -140,8 +153,23 @@ module ZypperAuth
         alias_method :original_path_allowed?, :path_allowed?
 
         # additional validation for strict_authentication auth subrequest
-        def path_allowed?(path)
-          return false unless original_path_allowed?(path)
+        def path_allowed?(headers)
+          return false unless original_path_allowed?(headers)
+
+          if @system.hybrid?
+            paid_extensions = @system.products.select { |prod| prod if !prod.free && prod.product_type == 'extension' }
+            paid_extensions.each do |paid_extension|
+              repos_paths = paid_extension.repositories.pluck(:local_path)
+              repos_paths.each do |repo_path|
+                if headers.fetch('X-Original-URI', '').include? repo_path
+                  logger.info "verifying paid extension #{paid_extension.identifier}"
+                  return ZypperAuth.verify_instance(request, logger, @system, paid_extension.id)
+                end
+              end
+            end
+          end
+          # either no hybrid system, no need to check for paid extensions
+          # or path not found on paid extensions and system is hybrid
           ZypperAuth.verify_instance(request, logger, @system)
         end
       end
--- a/engines/zypper_auth/lib/zypper_auth/engine.rb
+++ b/engines/zypper_auth/lib/zypper_auth/engine.rb
@@ -35,7 +35,10 @@ module ZypperAuth
     rescue InstanceVerification::Exception => e
       if system.byos?
         result = SccProxy.scc_check_subscription_expiration(request.headers, system, base_product.product_class)
-        return true if result[:is_active]
+        if result[:is_active]
+          InstanceVerification.update_cache(request.remote_ip, system.login, base_product.id)
+          return true
+        end
       end
 
       ZypperAuth.zypper_auth_message(request, system, verification_provider, e.message)
--- a/engines/zypper_auth/lib/zypper_auth/engine.rb
+++ b/engines/zypper_auth/lib/zypper_auth/engine.rb
@@ -6,7 +6,7 @@ module ZypperAuth
       Thread.current[:logger]
     end
 
-    def verify_instance(request, logger, system)
+    def verify_instance(request, logger, system, _params_product_id = nil)
       return false unless request.headers['X-Instance-Data']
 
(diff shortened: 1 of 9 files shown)


rjschwei (approved):

jesusbv:
I will add the tests and coverage but it can be reviewed in the meantime

engines/zypper_auth/lib/zypper_auth/engine.rb
```diff
@@ -51,9 +51,22 @@ def verify_instance(request, logger, system, params_product_id = nil)
     end
 
     def handle_scc_subscription(request, system, verification_provider, params_product_id = nil)
-      product_class = Product.find_by(id: params_product_id).product_class if params_product_id.present?
-      result = SccProxy.scc_check_subscription_expiration(request.headers, system, product_class)
-      return true if result[:is_active]
+      if params_product_id.present?
+        product_class = Product.find_by(id: params_product_id).product_class
+        result = SccProxy.scc_check_subscription_expiration(request.headers, system, product_class)
+        return true if result[:is_active]
```
digitaltom, line 57: If `result[:is_active]` is not true here, the method would return the value from the line before, which is `result`. I suggest to always return true or false here: 

```suggestion
        return result[:is_active]
```

engines/zypper_auth/lib/zypper_auth/engine.rb
```diff
@@ -51,9 +51,22 @@ def verify_instance(request, logger, system, params_product_id = nil)
...
+        # no product id provided
+        # check all non free extensions subscriptions with SCC, if any
+        paid_extensions = system.products.select { |prod| prod if !prod.free && prod.product_type == 'extension' }
+        return true if paid_extensions.empty?
+
+        all_active = paid_extensions.all? do |paid_extension|
+          result = SccProxy.scc_check_subscription_expiration(request.headers, system, paid_extension.product_class)
+          result[:is_active]
+        end
+        return true if all_active
```
digitaltom, line 68:
```suggestion
        return all_active
```

engines/zypper_auth/lib/zypper_auth/engine.rb
```diff
@@ -140,8 +153,23 @@ def verify_instance
         alias_method :original_path_allowed?, :path_allowed?
 
         # additional validation for strict_authentication auth subrequest
-        def path_allowed?(path)
-          return false unless original_path_allowed?(path)
+        def path_allowed?(headers)
+          return false unless original_path_allowed?(headers)
```
digitaltom, line 157: What's going on here, `original_path_allowed?` is an alias for `path_allowed?`, and then we call `original_path_allowed?` from inside `path_allowed?`?

jesusbv: correct

engines/zypper_auth/lib/zypper_auth/engine.rb
```diff
@@ -140,8 +153,23 @@ def verify_instance
...
+
+          if @system.hybrid?
```
rjschwei, line 159: But shouldn't this apply to proxy systems as well? A BYOS system that uses RMT as a proxy to SCC should have the same verification right? OR do we not hit this code path for proxy clients?

engines/zypper_auth/lib/zypper_auth/engine.rb
```diff
@@ -140,8 +153,23 @@ def verify_instance
...
+            paid_extensions = @system.products.select { |prod| prod if !prod.free && prod.product_type == 'extension' }
+            paid_extensions.each do |paid_extension|
+              repos_paths = paid_extension.repositories.pluck(:local_path)
+              repos_paths.each do |repo_path|
+                if headers.fetch('X-Original-URI', '').include? repo_path
+                  logger.info "verifying paid extension #{paid_extension.identifier}"
+                  return ZypperAuth.verify_instance(request, logger, @system, paid_extension.id)
```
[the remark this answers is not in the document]
rjschwei: This creates a loop `ZypperAuth.verify_instance` calls `handle_scc_subscription` in 2 places. In one of those places we have ```if is_valid && system.hybrid? && !handle_scc_subscription(request, system, verification_provider, params_product_id)``` So if the instance verification passed and the system is hybrid we end up in `handle_scc_subscription` and then here, since the system is a hybrid system we turn around and call `verify_instance` again. I do not understand how this is going to end well.

engines/zypper_auth/lib/zypper_auth/engine.rb
jesusbv: we do, and I think `byos?` system should be included, unless there are BYOS subscriptions to register the system that include LTSS, meaning one regcode including system + LTSS instead of two regcodes

I think the workflow is two reg codes

engines/zypper_auth/lib/zypper_auth/engine.rb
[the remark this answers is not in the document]
jesusbv: > and then here, since the system is a hybrid system we turn around and call verify_instance again

I do not know what you mean, if you mean that we call `ZypperAuth.verify_instance` twice, 
- one in line 166 (which we return the returned value of that call)
- one in line 173

the call in line 173 happens if the system is not hybrid (and I would add byos) or if it is hybrid/byos and the path trying to be accessed was not found in that system products

is that what you meant, @rjschwei  ?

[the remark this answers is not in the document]
jesusbv: my comment of moving the handle_scc_subscription call to `path_allowed?` is more of "whose responsibility is it ?"

For example, if zypper is trying to access path A, it makes sense that the method checking if it is allowed
`path_allowed?` checks that the path is a non free repo and then checks that the subscription is valid/not expired

while `verify_instance` maybe should verify that the instance metadata is valid

It was just an open question than a suggestion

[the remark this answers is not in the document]
jesusbv: @rjschwei I understand what you meant, will adjust the PR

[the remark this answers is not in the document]
jesusbv: @rjschwei 

if you mean in the current implementation, we check if the system is `hybrid` or `byos` so only once would be checked

if you mean in the suggested implementation, moving the `has_active_subscription?` method to `path_allowed?`
we would add logic to avoid the loop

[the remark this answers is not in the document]
rjschwei: I am not sure how to explain this differently. Basically we have the following implementation, I think
```
def A(foo):
   if foo:
     B(foo)

def B(foo):
   if foo:
     A(foo)
```
that creates a circle I do not see the exit condition in the code.

And yes I mean the current implementation, i.e. the diff as it is presented in GitHub, that's what I am reviewing.

engines/zypper_auth/lib/zypper_auth/engine.rb
digitaltom: Well, 2 questions here: 
* Why is that alias needed? 
* Why is that method recursively calling into itself?

jesusbv:
Apologies for the big commit

I will add the remaining needed tests (to check cache refresh)

digitaltom:
Unfortunately I don't have a good way to verify & test this. So when Robert or somebody else from the pubcloud team approved, and tests on your side show it's fixing the hybrid scenarios, it's ok for me.

jesusbv:
> Unfortunately I don't have a good way to verify & test this. So when Robert or somebody else from the pubcloud team approved, and tests on your side show it's fixing the hybrid scenarios, it's ok for me.

I have tested this changes in
`SLES12-SP5-BYOS` 
- refresh cache works OK (cloudregistruauth)
- zypper ref works OK
- clean registration and re-register works OK

`SLES4SAP12-SP5-BYOS` 
- refresh cache works OK (cloudregistruauth)
- zypper ref works OK
- clean registration and re-register works OK

`SLES4SAP15-SP5-PAYG` 
- refresh cache works OK (cloudregistruauth)
- zypper ref works OK
- clean registration and re-register works OK

rjschwei (approved):

PR #1273, opened 2025-01-15
Check if repository metadata already exists (is cached)

felixsch:
Instead of re-downloading repomd metadata, check if the files are already exists in the "cache" (repository_path).

**How to review:**

```
$ git checkout master
$ bundle exec bin/rmt-cli repos enable 6301
$ bundle exec bin/rmt-cli mirror repository 6301
$ bundle exec bin/rmt-cli mirror repository 6301
# expect: See that the metadata is re-downloaded (↓ instead of →)

$ git checkout fix-non-cached-metadata
$ bundle exec bin/rmt-cli mirror repository 6301
# expect: See that the metadata is NOT re-downloaded (→ instead of ↓)
```

**Note:** Why is there no test for this? The logic of when to download what is in `downloader.rb` rather then `repomd.rb`. If you insist I add a test checking if `cache_dir:` is set? Needed?

**Thank you for reviewing!**

cheers,

### DIFF
--- a/lib/rmt/mirror/repomd.rb
+++ b/lib/rmt/mirror/repomd.rb
@@ -28,7 +28,7 @@ class RMT::Mirror::Repomd < RMT::Mirror::Base
 
     metadata_files = RepomdParser::RepomdXmlParser.new.parse_file(repomd_xml.local_path)
       .map do |reference|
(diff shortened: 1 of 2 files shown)


mssola (approved):
Works as advertised 👍🏻

mssola:
> If you insist I add a test checking if cache_dir: is set? Needed

Not needed from my side.

PR #1276, opened 2025-01-21
Update cache scrubber

jesusbv:
## Description

Add the new cache directories and their expiration times to the cache scrubber 

This is a step in the list of steps to reduce queries to SCC

* Related Issue / Ticket / Trello card: <link reference>

## How to test 

with this changes in place, restarting RMT should show the new cache directories in /var/lib/rmt/rmt-cache-trim.sh

## Change Type

*Please select the correct option.*

- [ ] **Bug Fix** (a non-breaking change which fixes an issue)
- [X] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [X] I have reviewed my own code and believe that it's ready for an external review.
- [ ] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [X] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Review

Please check out our [review guidelines](https://github.com/SUSE/scc-docs/blob/master/team/workflow/code_review.md) 
and get in touch with the author to get a shared understanding of the change.

PR #1278, opened 2025-01-24
Add cache check when activating a product with SCC

jesusbv:
## Description

When a BYOS or HYBRID system activates a product with SCC we check first if the registration code use for that activation is in the cache for that product

If there is a match in the cache and the registration code is active we do not reach to SCC, update the cache and update the system pubcloud_reg_code if reg code is new or different

If there is no match in the cache we call SCC, update the cache and update the system pubcloud_reg_code if it's new or different

If the registration code is inactive, we do not follow with the activation

## Change Type

## Checklist

## Review


jesusbv:
This PR needs the changes of #1275  and #1276 merged and rebased, in the meantime, will work on the tests for this PR

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -17,9 +18,40 @@ def self.update_cache(remote_ip, system_login, product_id, registry: false)
     )
   end
 
+  def self.update_cache_not_payg(cache_key, mode)
```
rjschwei, line 21: Why do we have to differentiate between payg and other flavors for the cache update? Appears to me to be a fully generic mechanism, we write a given key to a given location.

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -17,9 +18,40 @@ def self.update_cache(remote_ip, system_login, product_id, registry: false)
...
+      Rails.application.config.repo_hybrid_cache_dir
+    else
+      # payg
+      Rails.application.config.repo_payg_cache_dir
+    end
+  end
+
+  def self.get_cache_entries(mode)
+    cache_path = InstanceVerification.get_cache_path(mode)
+    FileUtils.mkdir_p(cache_dir)
```
rjschwei, line 45: We should not create a directory as a side effect of trying to get the entries in that directory.

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -17,9 +18,40 @@ def self.update_cache(remote_ip, system_login, product_id, registry: false)
...
+    Dir.children(cache_path)
+  end
+
+  def self.reg_code_in_cache?(reg_code, product_hash, mode)
```
rjschwei, line 49: Generic method name but specific implementation to BYOS.

rjschwei (changes requested):

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: `BYOS` and `HYBRID`, I understand based on the comments we want to include `PAYG`

rjschwei (approved):

PR #1280, opened 2025-01-31
filter by the whole key

jesusbv:
## Description

When creating or updating a system during regsharing, use the full unique combination

This Fixes bsc#1236600

* Related Issue / Ticket / Trello card: https://bugzilla.suse.com/show_bug.cgi?id=1236600

## Change Type

## Checklist

## Review


digitaltom (commented):
Looks good in general. But this needs testing, because I could imagine that when the system token rotates on the source RMT, the system would not get found on the target RMT anymore because it has a different system token. How do we handle that case?

rjschwei (approved):

PR #1281, opened 2025-02-05
When migrating compare product classes

jesusbv:
## Description

Because of the change from SLE Micro < 6.0 to SL Micro 6+ comparing the identifier no longer works when migrating

This fixes bsc#1236816

* Related Issue / Ticket / Trello card: [bsc#1236816](https://bugzilla.suse.com/show_bug.cgi?id=1236816)

## How to test 

Start a Micro 5.5 instance, run `transactional-update migration` , choose any of the available options, i.e.

```bash
Available migrations:

    1 |SUSE Linux Micro 6.1 x86_64

    2 |SUSE Linux Micro 6.0 x86_64


[num/q]: 1
```
Migration should succeed

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -122,7 +122,7 @@ module InstanceVerification
 
           activated_bases = @system.products.where(product_type: 'base')
           activated_bases.each do |base_product|
-            return true if (base_product.identifier == upgrade_product.identifier)
+            return true if base_product.product_class == upgrade_product.product_class
           end
 
           raise ActionController::TranslatedError.new('Migration target not allowed on this instance type')
--- a/engines/instance_verification/spec/requests/api/connect/v3/systems/products_controller_spec.rb
+++ b/engines/instance_verification/spec/requests/api/connect/v3/systems/products_controller_spec.rb
@@ -624,6 +624,7 @@ describe Api::Connect::V3::Systems::ProductsController, type: :request do
           let(:new_product) do
             FactoryBot.create(
               :product, :with_mirrored_repositories, identifier: old_product.identifier,
+              product_class: old_product.product_class,
               version: '999', predecessors: [ old_product ]
               )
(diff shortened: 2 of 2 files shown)


engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -122,7 +122,7 @@ def verify_base_product_upgrade
 
           activated_bases = @system.products.where(product_type: 'base')
           activated_bases.each do |base_product|
-            return true if (base_product.identifier == upgrade_product.identifier)
+            return true if base_product.product_class == upgrade_product.product_class
```
digitaltom, line 125: Maybe `base_product.identifier == upgrade_product.identifier || base_product.product_class == upgrade_product.product_class` is more safe.

jesusbv: I considered that option but isn't  `product_identifer` matching a subset of `product_class` matching ?

digitaltom: The product identifier can change across versions (as happened with Micro), but also the product class can change. So the suggested condition would be more safe. 
In the end what we do here is a hack because it's not based on the subscription. In SCC, the migration would check if the target product is included in the system's subscription.

jesusbv: OK, so, similar to the issue with SUMA ARM64, a better check on subscription is recommended/encouraged 

I will do that instead of the double hack (product identifier + product class)

Thoughts ?

digitaltom: In general, it would be useful if you have a way also for PAYG to find out the used subscription, to use it's settings to determine which products a system should have access to.

jesusbv: yes, the fix/new check would be for PAYG as well

jesusbv:
closing in favour of #1283

PR #1287, opened 2025-02-20
Upgrade to ruby 3.4.1 and Rails 7.2

mssola:
## Description

This PR does not just upgrade to Ruby 3.4.1, but it also upgrades a lot of gems underneath. This has been accidental by the way I needed to manage bundler. Anyways, I have split the Gemfile into both versions, and the `Gemfile` and `Gemfile.lock` files are merely a symlink.

There are still tests to be worked on, and the building surely needs a rebump (e.g. picking the right symlink?). Hence the `wip` label.

This PR also brings some improvements to the sqlite3 setup, which I was only able to reproduce while performing the upgrade.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,1 +1,1 @@
-Gemfile-3.4.lock
+Gemfile-3.4.1.lock


mssola:
I believe this is being handled somewhere else. Closing it...

PR #1306, opened 2025-03-25
Handle null client

### LINKED ISSUE #1302: 500 Internal Server Error on registry pull leads to Rails NoMethod FATAL error
encountering 500 Internal Server errors on public cloud team production infra RMT servers as shown in the following systemd journal entries:

https://etherpad.nue.suse.com/p/500-ISE-Registry-NoMethodError
https://etherpad.nue.suse.com/p/500-ISE-Registry-NoMethodError-multiple

jesusbv:
## Description

Do not raise/log a 500 error when there is no client

This Fixes #1302

* Related Issue / Ticket / Trello card: #1302 

## How to test 

using the registry should not log 500 errors

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/registry/app/models/access_scope.rb
+++ b/engines/registry/app/models/access_scope.rb
@@ -81,8 +81,8 @@ class AccessScope
     raise Registry::Exceptions::InvalidScope.new('Invalid scope format') unless %r{^[a-z0-9\-_/:*(),.]+$}i.match?(scope)
   end
 
-  def allowed_paths(system = nil)
-    repo_list = RegistryCatalogService.new.repos(reload: false, system: system)
+  def allowed_paths(system)
+    repo_list = RegistryCatalogService.new.repos(system, reload: false)
     access_policies_yml = YAML.safe_load(
       File.read(Rails.application.config.access_policies)
     )
--- a/engines/registry/app/models/access_scope.rb
+++ b/engines/registry/app/models/access_scope.rb
@@ -61,7 +61,10 @@ class AccessScope
     if @namespace.nil?
       @image == 'catalog' ? @actions : AUTHORIZED_ACTION
     else
-      allowed_paths(client.systems.first)
+      @allowed_paths = []
+      allowed_paths(client.systems.first) if client.present?
+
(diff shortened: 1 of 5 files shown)


engines/registry/app/models/access_scope.rb
```diff
@@ -61,7 +61,10 @@ def authorized_actions(client = nil)
     if @namespace.nil?
       @image == 'catalog' ? @actions : AUTHORIZED_ACTION
     else
-      allowed_paths(client.systems.first)
+      @allowed_paths = []
+      allowed_paths(client.systems.first) if client.present?
```
digitaltom, line 65: It's unexpected that the `allowed_paths` method internally sets the `@allowed_paths` variable. Also, the `allowed_paths` method has a default of `nil` for the system parameter, so why do we need to skip it when the client is nil? 

I'd suggest to write this as: 

`@allowed_paths = allowed_paths(client&.systems&.first)`

By skipping setting the allowed paths entirely when there is no client set, wouldn't that break accessing free images for anonymous clients?

jesusbv: that was the first option I did, but inside of `allowed_paths` method we use the [system](https://github.com/SUSE/rmt/blob/master/engines/registry/app/models/access_scope.rb#L86) that would also break/raise exception

jesusbv: happy to refactor the system use chain, just chose the smaller change to fix the issue

digitaltom: If it breaks without a system, then the `system = nil` default is misleading. Are there no paths allowed for unauthenticated clients?

jesusbv: yes, IIRC, the catalog check does a query without a system

jesusbv: Done

jesusbv: Done

digitaltom (approved):

rjschwei (approved):

Issue #1288, opened 2025-02-21
Do not mirror delta rpms by default


dirkmueller:
Checking the SUSE CDN logs, we see that RMT / SMT servers are downloading delta rpms. zypper(1) is falling back gracefully if drpms are not available, and downloading them to store locally seems like a waste of bandwith and storage, as the RMT setup is close / high bandwidth to the actual machines that want updates, and delta rpms make no sense there. 

So much they don't make sense that there is a discussion to turn it off client side even: 

https://github.com/openSUSE/libzypp/pull/620?email_source=slack

digitaltom:
Disabling the download on delta RPMs is likely changed here: https://github.com/SUSE/rmt/blob/master/lib/rmt/mirror/repomd.rb#L64
We would need to investigate if there can be issues on the client side, when those delta RPMs are not available on RMT, because the repo metadata still references them, we cannot rewrite the metadata.

dirkmueller:
libzypp clients will silently ignore failures on delta rpms and fall back to the full rpm. dunno about other clients though.

ngetahun:
Closing this as it has been fixed by #1308

PR #1308, opened 2025-03-26
[WIP] Disable `.drpm` packages mirroring by default

lcaparroz:
## Description

* Related Issue: #1288 

## How to test 

TBD

## Change Type

## Checklist

## Review


lib/rmt/cli/clean.rb
```diff
@@ -121,18 +121,15 @@ def dangling_packages_list(repo_base_dir, repomd_file)
     packages = (actual_packages - expected_packages).sort
 
     packages.map do |file|
-      next nil unless File.exist?(file)
-
-      # Only remove files if they were not recently created
-      file_stat = File.stat(file)
-      next nil if (Time.current - file_stat.mtime) < DANGLING_FILE_MINIMUM_AGE
+      file_status = File.stat(file)
```
wstephenson, line 124: Do you need to rename this variable? `file_stat` makes it clear that it's a Stat object whereas `file_status` is far less definite.

wstephenson, line 124: Won't `File.stat('/path/doesnt/exist')` fail here, which is unhandled?  And it would render the `File.exist?` call at line 160 inside `eligible_for_cleaning?` moot - File.exist is just a wrapper around stat() anyway.

lib/rmt/config.rb
```diff
@@ -38,6 +37,10 @@ def mirror_src_files?
       ActiveModel::Type::Boolean.new.cast(Settings.try(:mirroring).try(:mirror_src))
     end
 
+    def mirror_drpm_files?
+      ActiveModel::Type::Boolean.new.cast(Settings.try(:mirroring).try(:mirror_drpm)) || false
```
digitaltom, line 41:
```suggestion
      !!Settings.try(:mirroring).try(:mirror_drpm)
```

lcaparroz: The `!!` "method" won't convert a string like `"'false'"` to `false`, but rather to `true`. I considered this scenario in case the user surrounds a boolean with quotes, making it a string in the YAML file.

There is an example in the `RMT::Config` [spec, line 89](https://github.com/SUSE/rmt/pull/1308/files#diff-9e0b1824898f095a1821711cea640a8250ed44318d34f15fd3d452adab3f6eb1R89-R93).

lib/rmt/cli/clean.rb
lcaparroz: Reverted it to the original naming on 020f16f.

lcaparroz: You actually make an excellent point. I changed the behavior on 020f16f.

lib/rmt/mirror/repomd.rb
```diff
@@ -65,9 +65,16 @@ def parse_packages_metadata(metadata_references)
...
       next unless xml_parsers.key? file.type
 
       xml_parsers[file.type].new.parse_file(file.local_path)
     end.flatten.compact
   end
+
+  def need_to_download?(ref)
+    return false if !RMT::Config.mirror_drpm_files? && ref.local_path.match?(/\.drpm$/)
+
+    super(ref)
```
felixsch, line 78: I'm wondering, is this really needed since `next if file.type == :deltainfo && !RMT::Config.mirror_drpm_files?` should not even add the references of the list of potential packages to check, no?

lcaparroz: I was sincerely thinking about this too.

This part is an additional guard rail, *if it happens that files other than `type == :deltainfo` also contain `.drpm` packages*. I mean, what would happen if a `primary` file also contains `.drpm`?

I'm not a specialist in *repomd* repositories, and I wanted to play safe here, but, of course, we can review it if we feel assured enough that only `deltainfo` files references `.drpm`.

felixsch: mhm good point. Then let it stay as it is. I think `.drpm` definitions should not in `primary.xml` since it will list the packages available and drpm is only a diff of existing packages but you never know how strange certain repositories are setup.

lib/rmt/cli/clean.rb
```diff
@@ -72,7 +72,7 @@ def run_package_clean(repomd_files)
     partial_reports =
       dangling_packages_list_by_dir(repomd_files).map do |repo_dir, dangling_files|
         unless options.dry_run
-          FileUtils.rm(dangling_files.map(&:path))
+          FileUtils.rm_f(dangling_files.map(&:path))
```
digitaltom, line 75: Did you discover a case where the '-f' was needed?

spec/spec_helper.rb
```diff
@@ -106,4 +106,13 @@
       example.run
     end
   end
+
+  # Reload configuration only from "default" RMT config file to avoid
+  # interference from '/etc/resolv.conf' or 'config/rmt.local.yml' when running
+  # tests randomly.
```
digitaltom, line 112: :heart: I wanted to do the same in my pr. Can you add a comment about this to https://github.com/SUSE/rmt/blob/disable-drpm-mirror-default/lib/rmt/config.rb#L9 ?

spec/spec_helper.rb
```diff
@@ -106,4 +106,13 @@
       example.run
     end
   end
+
+  # Reload configuration only from "default" RMT config file to avoid
+  # interference from '/etc/resolv.conf' or 'config/rmt.local.yml' when running
```
digitaltom, line 111:
```suggestion
  # interference from '/etc/rmt.conf' or 'config/rmt.local.yml' when running
```

lib/rmt/cli/clean.rb
digitaltom: Is this in reference to spec/lib/rmt/cli/clean_spec.rb:428?

digitaltom (approved):

lib/rmt/cli/clean.rb
lcaparroz: > Is this in reference to spec/lib/rmt/cli/clean_spec.rb:428?
Yes, I thought about the same scenario: if a file was previously identified as dangling, and before deletion by the command, it got removed. I changed it after feedback from @wstephenson. It's a possible scenario I didn't consider before.

`File.rm` fails if any of the files doesn't exist anymore. `File.rm_f` will just remove whatever files exist without failing or providing any feedback about non-existing paths.

I wondered if we needed to have more granularity here, e.g., removing each file individually so we can report they were actually removed by the command. However, I considered the idea to be "over-engineered," since, ideally, mirrored files should only be managed by RMT itself.

The whole idea is to assert the command will not fail if some concurrent process removes any dangling files simultaneously.

Do you think we should improve the code here, or let an explicit comment?

digitaltom: I think it's fine, looks good to me

spec/spec_helper.rb
lcaparroz: Added on 5e4f7be.

spec/spec_helper.rb
lcaparroz: Fixed on 5e4f7be.

PR #1309, opened 2025-03-26
Option to not parse repodata file when it's unchanged

digitaltom:
## Description

Parsing metadata files can be slow and cpu intense on large update repos (> 300MB uncompressed primary.xml). 
With this patch, the parsing of such files, and subsequent validation / download of referenced files 
gets skipped when the metadata file is unchanged compared to the one on disk. 

This speeds up the mirroring process significantly, but brings the (theoretical?) issue, 
that missing files on disk will not get re-downloaded automatically on the next mirror run. 

## How to test 

With the config of `revalidate_repodata: false`, mirror a specific repo multiple times with `--debug` and see that only the first time the repodata gets parsed. Eg.: `bin/rmt-cli mirror repository 5710 --debug`

When changing the setting back to default `revalidate_repodata: true`, see that repodata gets parsed every time. 

## Change Type

## Checklist

## Review


jesusbv (approved):

lib/rmt/mirror/repomd.rb
```diff
@@ -21,26 +21,27 @@ def mirror_implementation
...
 
-    metadata_files = RepomdParser::RepomdXmlParser.new.parse_file(repomd_xml.local_path)
+    updated_metadata_files = RepomdParser::RepomdXmlParser.new.parse_file(repomd_xml.local_path)
       .map do |reference|
         ref = RMT::Mirror::FileReference.build_from_metadata(reference, base_dir: temp(:metadata), base_url: repomd_xml.base_url, cache_dir: repository_path)
         enqueue ref
-        ref
-      end
+        (RMT::Config.revalidate_repodata? || metadata_updated?(ref)) ? ref : nil
+      end.compact
```
lcaparroz, line 35: lI was thinking about the behavior here:

- When RMT considers that a file hasn't been updated, it still enqueues it for download.
- Late on, the downloader class will make a `HEAD` request to also verify if a file has been updated.
- Unless the timestamp of the files doesn't match, it should not re-download the file (this scenario isn't expected at all, unless a repodata file was updated and still has the same checksum).l

Then, I wondered if we could avoid the hassle of making requests if RMT already considered the file hasn't been updated. This means we could potentially skip the `enqueue` method here if no changes are detected.

However, I tried it locally, and unfortunately, it resulted in a strange behavior were every other run of the `mirror` command would re-downloaded the files even with `mirroring.revalidate_repodata == false`:

- If the files already exist and have not changed, RMT wouldn't enqueue it for download.
- Then, before completing the `mirror` command, the files would be moved (`move_files` on line 18), but this actually would delete the files, since RMT clears the destination path before moving, and no files have been copied around (the downloader class copy the files if they are cached, thus avoiding this issue).
- On the next run, the file wouldn't exist and then RMT re-downloads it.

I guess we could do some improvements to this behavior specifically, but then we'd need to rework this `move_files` logic to avoid this issue, but it sounds a bit too much work for now.

What do you think?

MANUAL.md
```diff
@@ -262,6 +262,8 @@ The `mirroring` section lets you adjust mirroring behavior.
     to `false` if absent.
   * `mirroring.mirror_src`:
     Whether to mirror source (arch = `src`) RPM packages or not.
+  * `mirroring.revalidate_repodata`:
+    Whether to re-validate repodata content and packages when the repodata did not change
```
lcaparroz, line 266: It would be nice to have a statement about the default behavior when the key is absent in the configuration file (as in lines 261~262), as defined by the new method in `RMT::Config`:

```suggestion
    Whether to re-validate repodata content and packages when the repodata did not change. Defaults to `true` if absent.
```

lcaparroz (commented):
I tested it locally, and the implementation works as described. Great work :clap: 

I left a few remarks I think could be addressed before merging.

lib/rmt/mirror/repomd.rb
digitaltom: True, looks like there is potential for an additional optimization there. But I'd rather break it down into a separate issue, not to make thing too complex here.

lib/rmt/config.rb
```diff
@@ -34,6 +34,14 @@ def deduplication_by_hardlink?
       Settings.try(:mirroring).try(:dedup_method).to_s.to_sym != :copy
     end
 
+    # This method checks whether to re-validate metadata content and packages
+    # when the metadata did not change (default=true)
+    def revalidate_repodata?
+      return true if Settings.try(:mirroring).try(:revalidate_repodata).nil?
+
+      !!ActiveModel::Type::Boolean.new.cast(Settings.mirroring.revalidate_repodata)
```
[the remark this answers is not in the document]
lcaparroz: You could keep line 40 as it is and then just remove the `!!`, or make it all one line, such as:

```suggestion
      ActiveModel::Type::Boolean.new.cast(Settings.mirroring.revalidate_repodata) || true
```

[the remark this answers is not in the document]
lcaparroz: You could keep the line 40, as it is, and then just remove the `!!`:

```suggestion
      return true if Settings.try(:mirroring).try(:revalidate_repodata).nil?

      ActiveModel::Type::Boolean.new.cast(Settings.mirroring.revalidate_repodata)
```

lib/rmt/mirror/repomd.rb
Adnilson: Can you make a task for it?

felixsch:
Can you add this feature also to the Debian repository implementation? I think this should be consistent across all repositories types.

digitaltom:
> Can you add this feature also to the Debian repository implementation? I think this should be consistent across all repositories types.

Yes, I'll look into this.

lib/rmt/mirror/repomd.rb
digitaltom: @lcaparroz I'm addressing your described case in https://github.com/SUSE/rmt/pull/1309/commits/ebe94010e515289e4f0aae66ed272838a617355d, could you check?

digitaltom:
@felixsch added in https://github.com/SUSE/rmt/pull/1309/commits/0f05f32cbd5f3ca361afe137ae0b544cf3cf058b

package/obs/rmt-server.changes
```diff
@@ -5,8 +5,10 @@ Tue Jan 21 10:49:54 UTC 2025 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
   * rmt-server-pubcloud:
     * Add pubcloud_reg_code column to systems table
     * Add cache directories and their expiration times to the scrubber
+  * Update Puma webserver
```
felixsch, line 8: did we?

felixsch (approved):
lgtm :+1: 

**How I reviewed this merge request:**

- Checked the code changes and if I want to suggest something
- Mirrored products repos (repomd) and checked with settings
- Mirrored nginx ppa debian repository and checked

cheers,

package/obs/rmt-server.changes
digitaltom: Yes, but forgot to add it to changelog: https://github.com/SUSE/rmt/pull/1312/files

PR #1313, opened 2025-03-28
Build the cache without active

jesusbv:
## Description

When building the cache '-active' should not be part of it, otherwise, we would never find '-inactive' entries in the cache

## How to test 

Behaviour should remain the same

## Change Type

## Checklist

## Review


jesusbv:
I will add a `set_cache_active` method as well

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -24,7 +24,7 @@ def self.build_cache_entry(remote_ip, system_login, encoded_reg_code, mode, prod
       # for byos or hybrid cache
       product_hash = product.attributes.symbolize_keys.slice(:identifier, :version, :arch)
       product_triplet = "#{product_hash[:identifier]}_#{product_hash[:version]}_#{product_hash[:arch]}"
-      "#{encoded_reg_code}-#{product_triplet}-active"
+      "#{encoded_reg_code}-#{product_triplet}"
```
rjschwei, line 27: We came to the conclusion that the cache key needs to be different, would it not make sense to have 1 PR that changes the cache key to the latest format that we agreed upon?

jesusbv: I can combined both changes, yes, could not do it before as this PR is prior agreeing on the new format

jesusbv: added in #1314

jesusbv:
closing in favour of #1314, based on @rjschwei request to join the two changes in the same PR

PR #1317, opened 2025-04-04
Add upgrade cache check

jesusbv:
## Description

Check the cache before making a request to SCC

Refactor the cache check, so activate and upgrade product use the same code

Update registration code to the test

Update the test to allow cache check

* Related Issue / Ticket / Trello card: [reduce SCC queries](https://trello.com/c/62VdOel1)  [jira ticket](https://jira.suse.com/browse/PCT-473)

## How to test 

Upgrade a product present on the cache, it should not query the SCC

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -304,8 +304,6 @@ module SccProxy
 
         protected
 
-        # rubocop:disable Metrics/PerceivedComplexity
-        # rubocop:disable Metrics/CyclomaticComplexity
         def scc_activate_product
           product_hash = @product.attributes.symbolize_keys.slice(:identifier, :version, :arch)
           unless InstanceVerification.provider.new(logger, request, product_hash, @system.instance_data).allowed_extension?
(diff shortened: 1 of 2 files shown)


digitaltom:
As described [here](https://suse.slack.com/archives/C02DNAQ3BAP/p1743623482297529), I don't think you can skip the calls `SccProxy.scc_activate_product` and `SccProxy.scc_upgrade` when a system is in the cache. Because then SCC does not know about the activated products on the system and will be out of sync with the current state of the system. This will lead to errors on subsequent calls to acivate/upgrade/migrate products.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -389,6 +375,12 @@ def scc_upgrade
           auth = request.headers.fetch('HTTP_AUTHORIZATION', '')
           SccProxy.scc_upgrade(auth, @product, @system, logger)
           logger.info "System #{@system.login} successfully upgraded with SCC"
+          InstanceVerification.update_cache(
+            InstanceVerification.build_cache_entry(
+              request.remote_ip, @system.login, @system.pubcloud_reg_code, @system.proxy_byos_mode, @product
+            ),
```
rjschwei, line 381: Here we are just passing the product object, which is incompatible with the change in #1318 . I think passing the product is sufficient, see comments in #1318

jesusbv: it is not, this PR only pass the product because the change to fix that is implemented in a different PR
not this PR responsibility

PR #1318, opened 2025-04-07
Add subscription check cache

jesusbv:
## Description

Update the SCC subscription check method to first look for the subscription
in the cache

Update the call of the method

Update build cache as scc subscription check from registry uses product class

## How to test 

running zypper or registry catalog should use the cache

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -245,7 +245,7 @@ module InstanceVerification
               'The product is not available for this instance'
             )
           end
-          logger.info "Product #{@product.product_string} available for this instance"
+          logger.info "Product #{product.product_string} available for this instance"
           cache_key = InstanceVerification.build_cache_entry(request.remote_ip, @system.login, nil, 'payg', product)
           InstanceVerification.update_cache(cache_key, 'payg')
         end
--- a/engines/instance_verification/lib/instance_verification/engine.rb
+++ b/engines/instance_verification/lib/instance_verification/engine.rb
@@ -30,9 +30,7 @@ module InstanceVerification
               ''
             end
       encoded_reg_code = Base64.strict_encode64(params.fetch(:token, ''))
-      product_hash = product.attributes.symbolize_keys.slice(:identifier, :version, :arch)
-      product_triplet = "#{product_hash[:identifier]}_#{product_hash[:version]}_#{product_hash[:arch]}"
-      "#{encoded_reg_code}-#{iid}-#{product_triplet}"
+      "#{encoded_reg_code}-#{iid}-#{product.product_class}"
     end
   end
 
@@ -123,7 +121,9 @@ module InstanceVerification
     is_valid
   rescue InstanceVerification::Exception => e
     if system.byos?
-      result = SccProxy.scc_check_subscription_expiration(request.headers, system, base_product.product_class)
+      result = SccProxy.scc_check_subscription_expiration(
+        request.headers, system, request.remote_ip, false, base_product
+      )
       if result[:is_active]
         # update the cache for the base product
         InstanceVerification.set_cache_active(cache_key, 'byos')
@@ -247,7 +247,7 @@ module InstanceVerification
           end
           logger.info "Product #{@product.product_string} available for this instance"
           cache_key = InstanceVerification.build_cache_entry(request.remote_ip, @system.login, nil, 'payg', product)
-          InstanceVerification.set_cache_active(cache_key, 'payg')
+          InstanceVerification.update_cache(cache_key, 'payg')
         end
 
         def verify_base_product_activation(product)
--- a/engines/instance_verification/spec/requests/api/connect/v3/systems/products_controller_spec.rb
+++ b/engines/instance_verification/spec/requests/api/connect/v3/systems/products_controller_spec.rb
@@ -113,6 +113,7 @@ describe Api::Connect::V3::Systems::ProductsController, type: :request do
               headers: {}
               )
             allow(InstanceVerification::Providers::Example).to receive(:new).and_return(plugin_double)
+            allow(plugin_double).to receive(:instance_identifier).and_return('foo')
             allow(plugin_double).to receive(:allowed_extension?).and_return(true)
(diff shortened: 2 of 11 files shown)


jesusbv:
This is the first change of a series of changes that combine will implement SCC reduce queries effort

The next changes, in subsequent PRs will focus on undoing the changes for
- announce system/activate product
- upgrade product

Those calls need to reach SCC independently of the cache, otherwise the system and SCC would be out of sync

The dev plan will also be updated accordingly

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -15,16 +15,15 @@ def self.update_cache(cache_entry, mode, registry: false)
     )
   end
 
-  def self.build_cache_entry(remote_ip, system_login, encoded_reg_code, mode, product)
+  def self.build_cache_entry(remote_ip, system_login, encoded_reg_code, mode, product, product_class) # rubocop:disable Metrics/ParameterLists
```
digitaltom, line 18: Can we move `product_class = product.product_class` into this method, to not make the parameter list even longer?

jesusbv: no, because the `registry` engine uses `product_class` and does not have a product as it is now, see the [call](https://github.com/SUSE/rmt/blob/master/engines/registry/app/models/access_scope.rb#L100)

could we get the product with a DB query ? we are trying to reduce DB queries too

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -15,16 +15,15 @@ def self.update_cache(cache_entry, mode, registry: false)
...
       [remote_ip, system_login, product.id].join('-')
     elsif mode == 'registry'
       [remote_ip, system_login].join('-')
     else
       # for byos or hybrid cache
-      product_hash = product.attributes.symbolize_keys.slice(:identifier, :version, :arch)
-      product_triplet = "#{product_hash[:identifier]}_#{product_hash[:version]}_#{product_hash[:arch]}"
-      "#{encoded_reg_code}-#{product_triplet}-active"
+      # product is the product_class here
+      "#{encoded_reg_code}-#{product_class}-active"
```
rjschwei, line 26: I thought we were adding the system identifier to the cache entry, instanceid for EC2, vmid for Azure machineid for GCE?

engines/instance_verification/lib/instance_verification/engine.rb
rjschwei: Sorry, I do not quite understand. The parameter list already includes the product object and product_class is a member of that object. Why pass the object and a member of the object when inside the method we could access the member explicitly. Or is the `product` argument misnamed?

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -79,7 +78,10 @@ def self.verify_instance(request, logger, system)
     return false unless base_product
 
     # check the cache for the system
-    cache_key = InstanceVerification.build_cache_entry(request.remote_ip, system.login, system.pubcloud_reg_code, system.proxy_byos_mode, base_product)
+    cache_key = InstanceVerification.build_cache_entry(
+      request.remote_ip, system.login, system.pubcloud_reg_code,
+      system.proxy_byos_mode, base_product, base_product.product_class
```
rjschwei, line 83: In this instance of the call it is evident that the `build_cache_entry` does not need a `product_class` argument because inside of `build_cache_entry` we could just as well access the member of the `product` Object. What am I missing?

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: that is done in a different PR

This PR adds a subscription cache check

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: see comment above

engines/instance_verification/lib/instance_verification/engine.rb
```diff
@@ -98,7 +100,9 @@ def self.verify_instance(request, logger, system)
     is_valid
   rescue InstanceVerification::Exception => e
     if system.byos?
-      result = SccProxy.scc_check_subscription_expiration(request.headers, system, base_product.product_class)
+      result = SccProxy.scc_check_subscription_expiration(
+        request.headers, system, request.remote_ip, false, base_product, base_product.product_class
```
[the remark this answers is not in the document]
jesusbv: see comment above

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: we could change the logic in the registry method
```python
allowed_non_free_product_classes = allowed_product_classes.map { |s| s unless Product.find_by(product_class: s, product_type: 'extension').free? }.compact
```
to use the whole Product instead of its product class

engines/instance_verification/lib/instance_verification/engine.rb
rjschwei: I read it 3 times and I still don't understand sorry.

here we call `InstanceVerification.build_cache_entry(......, base_product, base_product.product_class)` which matches
`def self.build_cache_entry(......, product, product_class)`. Based on this signature match it appears that

`InstanceVerification.build_cache_entry(......, base_product)`
````def self.build_cache_entry(......, product)
product_class = product.product_class
```
is equivalent but does not require a signature change of the method.

jesusbv: OK, if we do not use the product class as a parameter, 
[this call](https://github.com/SUSE/rmt/blob/master/engines/registry/app/models/access_scope.rb#L98-L101) does not work, as mentioned, we could query the DB for the product object instead of using the product class,
however, that would add a bunch of DB queries, so using a param instead to avoid the extra DB queries

engines/instance_verification/lib/instance_verification/engine.rb
jesusbv: instead of using the extra param and not touch the registry allow path code,refactored that part and use product only

rjschwei (approved):

PR #1322, opened 2025-04-11
Export data when system activates or upgrades a product

jesusbv:
## Description

Send the information to data warehouse whenever a system activates or upgrades a product

* Related Issue / Ticket / Trello card: https://trello.com/c/n8vD9v1b

## How to test 

With the DataExport engine on, the activation of a product should trigger the data export method

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/data_export/lib/data_export/engine.rb
+++ b/engines/data_export/lib/data_export/engine.rb
@@ -27,16 +27,12 @@ module DataExport
       end
 
       Api::Connect::V3::Systems::SystemsController.class_eval do
-        after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? }
+        after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? && !@system.products.empty? }
 
         def export_rmt_data
-          if @system.products.empty?
+          @system.products.pluck(:name).each do |prod_name|
+            params[:dw_product_name] = prod_name
             send_data
-          else
-            @system.products.pluck(:name).each do |prod_name|
-              params[:dw_product_name] = prod_name
-              send_data
-            end
           end
         rescue StandardError => e
           logger.error('Unexpected data export error has occurred:')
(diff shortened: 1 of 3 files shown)


digitaltom:
@jesusbv @rjschwei Wouldn't it be better to decouple the telemetry data export to it's own asynchronous queue? 
With the current implementation it can slow down product activations, or in error case even break it.

jcomello (approved):

engines/data_export/lib/data_export/engine.rb
```diff
@@ -30,14 +30,49 @@ def export_rmt_data
         after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? }
 
         def export_rmt_data
+          if @system.products.empty?
+            send_data
```
rjschwei, line 34: I don't think we want to send anything if there is no product data. I don't quite see how data without the association to a given product would be useful.

jesusbv: only sending the data when systems controller performs an update if the system has products fixed in https://github.com/SUSE/rmt/pull/1322/commits/1f5fc18f1942ae85a63e47a84d29319120e7cd3b

engines/data_export/lib/data_export/engine.rb
```diff
@@ -27,16 +27,12 @@ def export_rmt_data
       end
 
       Api::Connect::V3::Systems::SystemsController.class_eval do
-        after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? }
+        after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? && !@system.products.empty? }
```
jcomello, line 30: You could use the `Object#present?` method here instead of `!@system.products.empty?`
But it's authors choice.
```suggestion
        after_action :export_rmt_data, only: %i[update], if: -> { response.successful? && !@system.byos? && @system.products.present? }
```

PR #1325, opened 2025-04-16
Make RMT docker compose setup more simple and robust

felixsch:
## Description

In the effort to keep development environments simple I want to update the `docker-compose` setup for RMT. This is the first step to eventually have a `ruby-2.5` and a `ruby-3.1` build when we need to split for SLE16.

## How to test 

```
$ cp .env.example .env
# Fill out the missing pieces in .env

$ make server
# expect: startes the rails server

# then in another terminal
$ make shell
> rmt-cli sync
# expect: works

> rmt-cli products enable SLES/15.6/x86_64
# expect: enable the product

> rmt-cli mirror
# expect: mirrors to your local public/ directory 

# then on the host in another terminal again after mirroring
$ docker run --rm --privileged -ti --network host registry.suse.com/bci/golang:1.21-openssl
> zypper in -y suseconnect-ng 
> suseconnect-ng --url http://localhost:8080
# expect: registers to the local RMT
```

**As always, if you have questions please do not hesitate to approach me!** :rocket:

thank you :heart:


Dockerfile
```diff
@@ -8,6 +8,9 @@ RUN zypper --non-interactive install --no-recommends \
     update-alternatives --install /usr/bin/bundle bundle /usr/bin/bundle.ruby2.5 5 && \
     update-alternatives --install /usr/bin/bundler bundler /usr/bin/bundler.ruby2.5 5
 
+RUN zypper --non-interactive install --no-recommends \
+        vim jq curl
```
lcaparroz, line 12: `vim` is already installed on line 6. Do we need a separated step for installing these `jq` and `curl`? Could we include those in the first `zypper install` call?

It would unnecessarily create another layer during the build process.

lcaparroz: Additionally, I was following the instructions on `PACKAGE.md` on a Docker container, the process breaks when building the `tar.bz2` file because the image doesn't have `bzip2`. Could you include `bzip2` on the `zypper install` statement?

docker-compose.yml
```diff
@@ -1,37 +1,23 @@
...
-    environment:
-      - MYSQL_HOST=db
-      - SECRET_KEY_BASE=${SECRET_KEY_BASE}
-      - SCC_USERNAME=${SCC_USERNAME}
-      - SCC_PASSWORD=${SCC_PASSWORD}
+    env_file: '.env'
     volumes:
       - .:/srv/www/rmt
-      - public:/srv/www/repo/public/
+      - ./config/rmt.yml:/etc/rmt.conf
```
lcaparroz, line 15: What is the case for removing the `public` volume? I see it as necessary if we want to have consistency between the database and the disk for downloaded files.

felixsch: I thought its easier to comprehend wehn it takes it from your local `public/` directory instead of having a separate volume. This way it is not hidden in a volume and you can access it from outside.

Do you see a downside to this?

Dockerfile
felixsch: yep will do

docker-compose.yml
lcaparroz: Maybe I'm confusing things here, but this volume should deny you access from outside the container, it will keep downloaded files on your local `public/`, but actually mounted as another volume on the Docker container.

I'm uncertain about the motivation for the volume, but I'd guess it's to keep the environment closer to an actual RMT installation, as I suppose `/srv/www/repo/public/` would be the path used on production systems.

felixsch: After revisting this PR: The directory `/srv/www/rmt` is mounted to `.` this also includes the `public` directory from the local directory. Do you think the extra `volume` brings any extra value here? Maybe that its easier to transfer?

Dockerfile
felixsch: done

docker-compose.yml
felixsch: We talked about this in our pairing session. `/srv/www/rmt/public` is mounted by `- .:/srv/www/rmt`. I do not see the point why this needs to be an extra volume. Mounting the `public` directory from the host system to somewhere different works fine.

Makefile
```diff
@@ -85,3 +85,18 @@ build-tarball: clean man
 	find $(NAME)-$(VERSION) -name \*~ -exec rm {} \;
 	tar cfvj package/obs/$(NAME)-$(VERSION).tar.bz2 $(NAME)-$(VERSION)/
 	rm -rf $(NAME)-$(VERSION)/
+
+database-up:
+	docker-compose up db -d
```
lagartoflojo, line 90: I can't comment on the rest of this PR, but I saw some conversation in our chat channels that `docker-compose` was deprecated in favor for `docker compose` =)

felixsch: aha old habits die hard, will update!

felixsch: done

felixsch:
I thought it makes sense to have a dedicated `uuid` for the development RMTs which would allow us on SCC side to have all the test systems in one place rather than scattered in multiple RMT instances. Do you see a problem with this?

Adnilson:
I meant to generate when we build the images in our own computers.

Adnilson (approved):

PR #1327, opened 2025-04-28
Bump bundler to 2.2.34 to build on SP6

felixsch:
## Description

Bundler had a major upgrade on SP6 due to CVEs which renders the 1.x branch unusable. This needs to be taken into account in the CI and additionally also in the spec file.

Here try to fix to get CI working again and make building RMT a little bit more easy.

The PR does:

- Fix the CI running the new version
- Move `make dist` into a container rather running on the development maschine
- Replace bundler version on rpm build time with the available from the build environment

NOTE: The updated bundler (2.2.34) version is now available on almost all service packs

card: https://trello.com/c/FpA0HZsY/3900-fix-rmt-package-building-on-obs

**How to test this merge request:**

- RMT CI is working again
- make dist:
```
 $ make dist
# expect: To run a container and build the tarball. The result is available in `package/obs`
```

- build against SLESSP7 (which not yet has the 2.2.34 update)
```
 $ make dist
 $ cp -r package/obs/* <obs project checkout>
 $ cd <obs project checkout>
 $ osc build SLE_15_SP7
```
HINT: To checkout a new copy of the the devel project from obs: `osc co systemsmanagement:SCC:RMT rmt-server`

**As always, if you think something is missing or having trouble to understand, please reach out!** :rocket:

thank you :rocket:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 NAME          = rmt-server
 VERSION       = $(shell ruby -e 'require "./lib/rmt.rb"; print RMT::VERSION')
 
+.PHONY = all clean man dist build-tarball
+
 all:
 	@:
 
@@ -13,7 +15,10 @@ man:
 	ronn --roff --pipe --manual RMT MANUAL.md > rmt-cli.8 && gzip -f rmt-cli.8
 	mv rmt-cli.8.gz package/obs
 
-dist: clean man
+dist:
+	docker-compose run -it -v $(PWD):/srv/www/rmt --rm rmt make build-tarball
+
+build-tarball: clean man
 	@mkdir -p $(NAME)-$(VERSION)/
 
 	@cp -r app $(NAME)-$(VERSION)/
--- a/ci/rmt-run-feature-tests
+++ b/ci/rmt-run-feature-tests
@@ -11,6 +11,12 @@ group "synchronize SCC product/repositories"
   rmt-cli sync
 groupend
 
+group "install feature test dependencies"
+pushd "$SOURCE"
+  bundle install --with test
+popd
+groupend
+
 group "run feature tests"
 pushd "$SOURCE"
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -13,7 +15,10 @@ man:
 	ronn --roff --pipe --manual RMT MANUAL.md > rmt-cli.8 && gzip -f rmt-cli.8
 	mv rmt-cli.8.gz package/obs
 
-dist: clean man
+dist:
+	docker-compose run -it -v $(PWD):/srv/www/rmt --rm rmt make build-tarball
```
digitaltom, line 19: What's the reason why we need to build the  dist tarball with docker-compose?

felixsch: The reason was that multiple people have problems with their ruby2.5 setup and this decouples the packaging process from the development setup. So you get a tarball no matter what your development setup looks like. Do you see a problem?

digitaltom: I see. Well for me it's inconvenient because I need to build containers now to build the tarball (took 5m). But I see that I can also use `make build-tarball`.

ci/rmt-run-feature-tests
```diff
@@ -11,6 +11,12 @@ group "synchronize SCC product/repositories"
   rmt-cli sync
 groupend
 
+group "install feature test dependencies"
+pushd "$SOURCE"
+  bundle install --with test
```
digitaltom, line 16: Why did this work before when we did not bundle the test gems? Maybe this already worked because of `RAILS_ENV=test`?

digitaltom (approved):

ci/rmt-run-feature-tests
felixsch: This is a good question where I have no answer for it, I was looking briefly into it why this is no necessary which only spawned the question, how was this even working before

PR #1346, opened 2025-07-18
Use instance id as a login on the public cloud environment

jesusbv:
## Description

Instance id is guarantee to be unique across the three CSPs, thus we do not need to generate a pseudo unique login value

This would help also with duplicated systems detection

## How to test 

On a public cloud environment, with this changed applied the system should have the instance id as a login:

Screenshot from 2025-07-18 15-48-53

Screenshot from 2025-07-18 15-49-18

## Change Type

## Checklist

## Review

### DIFF
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -20,11 +20,6 @@ class System < ApplicationRecord
   alias_attribute :scc_synced_at, :scc_registered_at
 
   def init
-    # if we are in the public cloud
-    # system_token should be not nil
-    self.login ||= system_token if system_token.presence
-    # if we are not in the public cloud
-    # use the regular SCC_UUID generated value
     self.login ||= System.generate_secure_login
     self.password ||= System.generate_secure_password
     self.registered_at ||= Time.zone.now
--- a/app/models/system.rb
+++ b/app/models/system.rb
@@ -20,6 +20,11 @@ class System < ApplicationRecord
   alias_attribute :scc_synced_at, :scc_registered_at
 
   def init
+    # if we are in the public cloud
+    # system_token should be not nil
+    self.login ||= system_token if system_token.presence
+    # if we are not in the public cloud
+    # use the regular SCC_UUID generated value
     self.login ||= System.generate_secure_login
     self.password ||= System.generate_secure_password
     self.registered_at ||= Time.zone.now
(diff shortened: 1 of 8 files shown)


jesusbv:
I will adapt the [example](https://github.com/SUSE/rmt/blob/master/engines/instance_verification/lib/instance_verification/providers/example.rb#L60) of the `InstanceVerification` provider to return `nil` following this approach and will adapt the tests too

app/models/system.rb
```diff
@@ -20,7 +20,16 @@ class System < ApplicationRecord
   alias_attribute :scc_synced_at, :scc_registered_at
 
   def init
-    self.login ||= System.generate_secure_login
+    self.login ||= if system_token.presence
+                     # if we are in the public cloud
+                     # system_token should be not nil
+                     # obfuscate the instance id
+                     Base64.strict_encode64(system_token)
```
rjschwei, line 27: I do not really see a reason to encode the instance ID

digitaltom:
I'm not sure I understand the benefits of this change. 
Current login is a random uuid4 value, which I wouldn't call pseudo random, likelihood of a collision here is astronomically small. 
Afaik, in the public cloud, we already use the instance ID as a static system_token, so it's already used to detect systems that were cloned.

app/models/system.rb
```diff
@@ -20,7 +20,16 @@ class System < ApplicationRecord
   alias_attribute :scc_synced_at, :scc_registered_at
 
   def init
-    self.login ||= System.generate_secure_login
+    self.login ||= if system_token.presence
```
digitaltom, line 23: I'd suggest to set the login in https://github.com/SUSE/rmt/blob/master/engines/scc_proxy/lib/scc_proxy/engine.rb#L291, then this is encapsulated to public cloud code.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -293,15 +293,16 @@ def announce_system
               system_information: system_information,
               proxy_byos_mode: :payg,
               instance_data: instance_data,
-              system_token: system_token
+              system_token: system_token,
+              login: system_token
```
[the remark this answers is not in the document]
jesusbv: > variable would be better called instance_identifier instead of system_token.

totally agree, noticed it while implementing this PR
the plan was to do the renaming on a different PR
but I guess we can start the renaming here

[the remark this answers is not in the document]

digitaltom (approved):

app/models/system.rb
rjschwei: It is the same, guessing another instances instance id is going to be about as successful as guessing someone else' generate user name. In any event the user name alone is useless and the underlying system will still fail access when it gets re-verified when the cache expires. I cannot think of a way to exploit the knowledge that the user name is the instance id.

Issue #1352, opened 2025-08-08
make build fails due to missing public/repo directory


rtamalin:
Attempting to run `make build` per the [development guide's docker-compose workflow](DEVELOPMENT.md#development-setup---docker-compose) is a freshly checked out repo fails as follows:

> ...
> #12 [7/8] RUN sed -i 's/#!\/usr\/bin\/env ruby/#!\/usr\/bin\/ruby.ruby2.5/g' /srv/www/rmt/bin/rmt-cli &&     ln -s /srv/www/rmt/bin/rmt-cli /usr/bin &&     mkdir /var/lib/rmt/ &&     groupadd -r nginx &&     useradd -g nginx -s /bin/false -r -c "user for RMT" _rmt &&     chown _rmt /srv/www/rmt/public/repo &&     chown _rmt /srv/www/rmt/public/suma
> #12 0.250 chown: cannot access '/srv/www/rmt/public/repo': No such file or directory
> #12 ERROR: process "/bin/sh -c sed -i 's/#!\\/usr\\/bin\\/env ruby/#!\\/usr\\/bin\\/ruby.ruby2.5/g' /srv/www/rmt/bin/rmt-cli &&     ln -s /srv/www/rmt/bin/rmt-cli /usr/bin &&     mkdir /var/lib/rmt/ &&     groupadd -r nginx &&     useradd -g nginx -s /bin/false -r -c \"user for RMT\" _rmt &&     chown _rmt /srv/www/rmt/public/repo &&     chown _rmt /srv/www/rmt/public/suma" did not complete successfully: exit code: 1
> ------
>  > [7/8] RUN sed -i 's/#!\/usr\/bin\/env ruby/#!\/usr\/bin\/ruby.ruby2.5/g' /srv/www/rmt/bin/rmt-cli &&     ln -s /srv/www/rmt/bin/rmt-cli /usr/bin &&     mkdir /var/lib/rmt/ &&     groupadd -r nginx &&     useradd -g nginx -s /bin/false -r -c "user for RMT" _rmt &&     chown _rmt /srv/www/rmt/public/repo &&     chown _rmt /srv/www/rmt/public/suma:
> 0.250 chown: cannot access '/srv/www/rmt/public/repo': No such file or directory
> ------
> Dockerfile:20
> 
> --------------------
> 
>   19 |     
> 
[...]

PR #1353, opened 2025-08-08
Fix missing public/repo path

rtamalin:
## Description

With the removal of the public/repo/.gitignore file the public/repo directory is no longer included in the git repo, causing make build to fail due a missing public/repo directory.

To avoid problems cause by this issue, and additionally simplify the use of the docker compose development environment, this patch adds new make rules to ensure that both the public/repo path is created and then public hierarchy is updated with appropriate permissions. These new rules will be triggered as preconditions for the build rule.

Update the DEVELOPMENT.md file to reflect these changes, and also explicitly indicate that running the server rule leaves the server running in the foreground of the user's shell session, and the same is true for the shell rule.

Also update the toplevel .gitignore to explicitly ignore just the public/repo directory, rather than any occurrence of a directory or file with the name repo.

* Related Issue: #1352 

## How to test 

Run `make build` in a freshly checked out repo; without the change it will fail, with the change it will succeed, and will also automatically set the public repo to have 0777 permissions.

## Change Type

This a development environment setup change, which doesn't fall into any of the above categories.

## Checklist

## Review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,7 @@ package/obs/.osc
 package/obs/rmt-cli.8.gz
 package/obs/_link
 package/systemsmanagement:SCC:RMT
-repo
+/public/repo/
 registry
 
 vendor/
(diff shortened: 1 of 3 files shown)


mssola (approved):

rtamalin:
Closing in favour of #1354 which is based upon a branch in the main repo rather than a fork, which should allow the feature tests to run successfully.

PR #1356, opened 2025-08-12
Find system only necessary

jesusbv:
## Description

Handle find_systems in two ways

- when there is instance identifier
- when there is not instance identifier

in the case that there is an instance identifier
in the request headers, it means we can check an exact match

if there is not an instance identifier in the headers,
we need to check for duplicates

## How to test 

for an existing system in the DB with login as instance identifier:
  auth an instance with the same login, pass and token it should not update the token
  
 for an existing system in the DB with login not as instance identifier and token present:
   auth an instance with the same login pass and token it should not update the token
   
for an existing system in the DB with login not  as instance identifier and token not present (nil):
  auth an instance with the same login, pass and a new token, it should update the token  

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -510,7 +510,17 @@ module SccProxy
               # this situation.
               return true if skip_on_duplicated && @systems.size > 1
 
-              @system = find_system
+              instance_data = ''
+              instance_data = Base64.decode64(request.headers['X-Instance-Data'].to_s) if request.headers['X-Instance-Data'].present?
+              @system = find_system(login, instance_data)
+              if @system.blank?
+                error = ActionController::TranslatedError.new(N_(
+                    "Can not find system with present credentials #{login} #{instance_data}"
+                ))
+                error.status = :unauthorized
+                raise error
+              end
+
               if system_tokens_enabled? && request.headers.key?(ApplicationController::SYSTEM_TOKEN_HEADER)
                 @system.update(last_seen_at: Time.zone.now)
                 headers[ApplicationController::SYSTEM_TOKEN_HEADER] = @system.system_token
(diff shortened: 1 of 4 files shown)


digitaltom:
I don't fully understand which problem this pr is solving. 
Is it about not setting a system token on systems that use the instance id as login username? 
The system token is set statically to the instance id in this case afaik.

jesusbv:
> I don't fully understand which problem this pr is solving. Is it about not setting a system token on systems that use the instance id as login username? The system token is set statically to the instance id in this case afaik.

before using `instance id` in the `login` column, the `login` has `SCC_UUID`, that made possible duplication (low, but non zero) so for the public cloud case, we replicated `find_system` now [find_system_by_token_header](https://github.com/SUSE/rmt/blob/master/app/controllers/application_controller.rb#L64) 

when all the systems in the DB use `instance id` in the login, find_system should not be necessary but we do not know
if the system is using that as a login, so the check is updated to take instance identifier into account and handle the cases

rjschwei:
Yes, I am also a bit confused why we need changes to system retrieval. I understand that we have collisions with the user name, 2 independent RMT servers generating the same user name and then when the data gets shared we have a problem to add a new user. But that both the generated user name and the generated password match appears not likely even at the scale of 40 million systems. As such I do not quite understand why we need changes on the system lookup.

jesusbv:
> Yes, I am also a bit confused why we need changes to system retrieval. I understand that we have collisions with the user name, 2 independent RMT servers generating the same user name and then when the data gets shared we have a problem to add a new user. But that both the generated user name and the generated password match appears not likely even at the scale of 40 million systems. As such I do not quite understand why we need changes on the system lookup.

the change is not about collisions, the change as mentioned [here](https://github.com/SUSE/rmt/pull/1356#issuecomment-3183396933) is  

> to take instance identifier into account and handle the cases

before this change, we had 

- a loop for all the systems found with username + pass
- check if the system token is present or not
  - if no systems with a system token
    - we update the value of the first payg/hybrid system we can find
  
now, we can use the instance data coming as a parameter in the request header, and find the exact system
or update

- if the system is using iid as a login -> direct match
- if the system is using old username SCC_UUID` and has a token -> direct match (unless same username and pass and different token, more information with this change)
- if system is using old username and no token -> same case as before this change, I think this change makes the `find_system` scenarios more accurate but Im open to feedback

digitaltom:
Sorry, but I still don't understand the benefit here. Introducing a custom login value and system lookup introduces additional complexity, that needs to get weighted against the benefit. Is the goal to save database queries? 

> before this change, we had
> 
>     a loop for all the systems found with username + pass
>     check if the system token is present or not
>         if no systems with a system token
>             we update the value of the first payg/hybrid system we can find

But this loop is only a loop in the case the system got cloned including the system credentials (which as far as I understood is not a case happening in the public cloud), or the extremely rare case of a uuid collision.

PR #1357, opened 2025-08-18
Ignore errors when setting public permissions

rtamalin:
Once public/repo has been populated by a running RMT it may not be possible to update the permissions on files due to ownership changes. Also quieten the associated error messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,4 +107,4 @@ public_repo:
 
 public_perm: public_repo
 	@echo ensure public permission is 0777
-	@chmod -R 0777 public
+	@chmod -fR 0777 public || true


Makefile
```diff
@@ -107,4 +107,4 @@ public_repo:
 
 public_perm: public_repo
 	@echo ensure public permission is 0777
-	@chmod -R 0777 public
```
digitaltom, line 110: Do we need to open the permissions that widely? I thought it'd be enough to have that directory writable by the Rails process. But maybe it doesn't matter when this is just used as a dev env.

digitaltom (approved):

Makefile
rtamalin: I was only following the original instructions in the DEVELOPING.md which specified chmod -R 0777 public.

If we only need to ensure that public/repo exists with a specific permission, we can refine this further.

PR #1366, opened 2025-09-15
docker compose development environment enhancements

rtamalin:
As it appears that only the public/repo directory requires modified permissions, enhance the Makefile build rules to ensure that those permissions are being applied to the directory after ensuring that it exists, and avoid changing the permissions for other files and directories under public.

Update the docker-compose.yml's MariaDB version to match the version found in SLE 15 SP7, namely version 10.11, and switch to using the SUSE built version, available from registry.suse.com.

Updated db/schema.rb:
  * Updated charset and collation settings triggered by running the rails db:migrate against the SLE 15 SP7 MariaDB 10.11 DB.
  * Re-orders the proxy_byos_mode and system_information fields in the systems table, adding a collation setting for the latter.
  * Adds the missing pubcloud_reg_code field to the db/schema.rb to align with db/migrate/20250121104357_add_regcode_to_systems.rb.

Updated DEVELOPMENT.md to explicitly call out that the make server and make shell commands may need to be run in separate user shell sessions/windows.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 2025_03_19_155325) do
 
-  create_table "activations", charset: "utf8", force: :cascade do |t|
+  create_table "activations", charset: "utf8mb3", collation: "utf8mb3_general_ci", force: :cascade do |t|
     t.bigint "service_id", null: false
     t.bigint "system_id", null: false
     t.datetime "created_at", null: false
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 2025_03_19_155325) do
 
-  create_table "activations", charset: "utf8", force: :cascade do |t|
+  create_table "activations", charset: "utf8mb3", collation: "utf8mb3_general_ci", force: :cascade do |t|
```
mssola, line 15: One thing I'm unsure about is that for all these changes, the charset and the collation varies for each table. I'd prefer to have just one charset for all tables, kind of like we had with `utf8`. I'm honestly not an expert on charsets on MySQL/MariaDB, but if we update this, we ought to be consistent.

Thus, would it be possible to check which charset/collation should be the one and simply have it everywhere. Also note that this charset is supposedly set in `config.rmt.yml`, so I do wonder why that is not honored.

rtamalin: As I said earlier, the db/schema.rb changes are not hand crafted, but rather the results of the interactions between `rails db:migrate` and the DB backend, i.e. the SUSE built MariaDB version 10.11 container.

So these values are automagically chosen by some means during those interactions. And as also referred to in my earlier response, and in previous discussions with @digitaltom , I investigated if explicitly setting overrides for the default character set and collation values that the MariaDB instance, via mapping in a custom cnf file under /etc/mysql/conf.d, had any affect, and even when I would set the character set to `utf8` (the same as the encoding value in `config/rmt.yml`) and the collation to `utf8mb3_general_ci`, I would see tweaks to some of the updated db/schema.rb's table and field settings to values other than those two setting.

I'm neither an expert in Ruby on Rails nor in MariaDB, so I'm happy for someone with more knowledge to provide suggestions for how to it better.

> Also note that this charset is supposedly set in config.rmt.yml, so I do wonder why that is not honored.
I think it is being honoured somewhat, in the sense that it is using utf8 variants, possibly chosen for optimisation purposes, e.g. the mb3 variant requires only 3 bytes to store characters, though could have issues with specific utf8 characters... When using the upstream MariaDB variants, which appear to be built with latin1 as the default, you will see latin1 variants appear in some cases.

Potentially we may want to be more explicit and specify utf8mb4 in the config/rmt.yml to ensure that it uses that variant?

PR #1368, opened 2025-09-23
Fix only one handler for data exporter

jesusbv:
## Description

if there are more than one handler for the data exporter, RMT will fail

Under the scenario of someone implementing their own handler, i.e. foo.rb and placing it there, they may rename or move example.rb and that would be OK

but if RMT gets updated it will place example.rb in the same location, making two handlers for the data exporter when it should be only one

This fixes that scenario and fixes bsc#1248869

* Related Issue / Ticket / Trello card: [bsc#1248869](https://bugzilla.suse.com/show_bug.cgi?id=1248869
)
## How to test 

On a server with RMT and data export engine with a handler, i.e. handler.rb, example.rb should be move to example in the same directory

## Change Type

## Checklist

## Review

### DIFF
--- a/package/obs/rmt-server.spec
+++ b/package/obs/rmt-server.spec
@@ -392,6 +392,9 @@ fi
 
 %post pubcloud
 %service_add_post rmt-server-regsharing.service rmt-server-trim-cache.service
+if [ $"(ls %{data_dir}/engines/data_export/lib/data_export/handlers/ | grep '*.rb' | wc -l > 1)" ]; then
+    mv %{data_dir}/engines/data_export/lib/data_export/handlers/example.rb %{data_dir}/engines/data_export/lib/data_export/handlers/example
+fi
 
 %preun pubcloud
 %service_del_preun rmt-server-regsharing.service rmt-server-trim-cache.service
--- a/package/obs/rmt-server.spec
+++ b/package/obs/rmt-server.spec
@@ -392,9 +392,6 @@ fi
 
 %post pubcloud
 %service_add_post rmt-server-regsharing.service rmt-server-trim-cache.service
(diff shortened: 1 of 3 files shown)


package/obs/rmt-server.spec
```diff
@@ -392,6 +392,9 @@ fi
 
 %post pubcloud
 %service_add_post rmt-server-regsharing.service rmt-server-trim-cache.service
+if [ $"(ls %{data_dir}/engines/data_export/lib/data_export/handlers/ | grep '*.rb' | wc -l > 1)" ]; then
+    mv %{data_dir}/engines/data_export/lib/data_export/handlers/example.rb %{data_dir}/engines/data_export/lib/data_export/handlers/example
+fi
```
digitaltom, line 397: We could also just ship the example handler as `handler.example`, so it's not picked up by default? Moving files that got installed by rpm before seems not optimal to me, because the `example` file might not get removed on rpm removal, and a file that was installed by the package (`example.rb`) is suddenly missing.

jesusbv: I like the idea

That approach has a downside, if one installs RMT with `handler.example`, RMT will start successfully 
but it will pollute the logs with "missing handler" error

jesusbv: I can update the README of that engine for users that do not use/implement data export engine mentioning to chaneg the name to `handler.rb` or some name

Thoughts ?

digitaltom: How about writing a warning only once, when the engine starts and there is no data export handler defined. 
After that we could just not try to do any data export.

digitaltom (approved):

rjschwei (approved):

PR #1396, opened 2025-11-25
Do not add example data exporter handler

jesusbv:
## Description

Do not place a 'example.rb' for the handler,
place 'example' without extension and the same content

the engine has a check in place to only trigger the data export if a
handler is present, so if a customer does not have a handler, they are
not affected or impacted and if they have a handler, then the checks
in place guarantee that must be only one handler and would trigger the
data export

## Change Type

## Checklist

## Review


package/obs/rmt-server.spec
```diff
@@ -343,6 +343,11 @@ getent passwd %{rmt_user} >/dev/null || \
 	%{_sbindir}/useradd -g %{rmt_group} -s /bin/false -r \
 	-c "user for RMT" %{rmt_user}
 %service_add_pre rmt-server.target rmt-server.service rmt-server-migration.service rmt-server-mirror.service rmt-server-sync.service rmt-server-systems-scc-sync.service rmt-uptime-cleanup.service
+# pick data exporter handler
+if [ -e "%{app_dir}/engines/data_export/lib/data_export/handlers/*.rb" ]; then
+  mv %{app_dir}/engines/data_export/lib/data_export/handlers/example.rb %{app_dir}/engines/data_export/lib/data_export/handlers/example
+fi
+
```
rjschwei: Hmm I wonder how this is supposed to work. `%pre` scriptlet runs "....before a package is installed..." Before a package is installed no files from the package are on the filesystem yet. Makes me wonder what is being moved and how this was tested.......

Probably works on update if there is an `example.rb` file from a previous install.

Also the test will be true if only `example.rb` is installed because `%{app_dir}/engines/data_export/lib/data_export/handlers/*.rb` will have content. Then if only `example.rb` exists, again at this stage it can only exist from a previous install,  it will be moved away and when the new package gets installed a new `example.rb` package will be put in place.

So this has some serious logic errors and a timing issue as to when the operation is executed.

rjschwei (approved):

PR #1405, opened 2026-01-08
Fix large rmt cli

jesusbv:
## Description

Process the list and purge rmt-cli commands by batches
    
There are a lots of systems in the DB,
processing them all with the current approach
takes too long or command does not respond
    
This change fixes that
    
Purge command interface remains the same while it removes the systems,
accordingly, the same fir the list command if listing the systems has
the --all option, it does show all the systems by batches
    
This fixes bsc#1253146 and bsc#1253147

* Related Issue / Ticket / Trello card: 

  - [bsc#1253146](https://bugzilla.suse.com/show_bug.cgi?id=1253146) 
  - [bsc#1253147](https://bugzilla.suse.com/show_bug.cgi?id=1253147)

## How to test 

With these changes applied, in any test infra, i.e. AWS (~800K systems), run the command
```bash
rmt-cli systems list --all
```
or
```bash
rmt-cli systems purge before <date>
```
and select `n` so no systems are destroyed

## Change Type

## Checklist

## Review

### DIFF
--- a/package/obs/rmt-server.changes
+++ b/package/obs/rmt-server.changes
@@ -3,6 +3,7 @@ Tue Jan 13 17:14:09 UTC 2026 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
 
 - Version 2.25
   * fix rmt-cli list and purge commands for large data (bsc#1253146 and bsc#1253147)
+  * Fix mirroring of SLE16 NVIDIA-GPU-Compute-Toolkit-CUDA repo (bsc#1256826)
 
 -------------------------------------------------------------------
 Tue Nov 11 11:03:50 UTC 2025 - Natnael Getahun <natnael.getahun@suse.com>
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -20,11 +20,6 @@ class Repository < ApplicationRecord
 
   class << self
 
-    def remove_suse_repos_without_tokens!
-      where(auth_token: nil).where("external_url LIKE '%.suse.com%' OR external_url LIKE '%.ranchergovernment.com%'")
-        .where(installer_updates: 0).where.not(scc_id: nil).delete_all
-    end
-
     # Mangles remote repo URL to make a nicer local path, see specs for examples
     def make_local_path(url)
       uri = URI(url)
--- a/lib/rmt.rb
+++ b/lib/rmt.rb
@@ -1,5 +1,5 @@
 module RMT
-  VERSION ||= '2.24'.freeze
+  VERSION ||= '2.25'.freeze
 
   DEFAULT_USER = '_rmt'.freeze
   DEFAULT_GROUP = 'nginx'.freeze
--- a/lib/rmt/mirror.rb
+++ b/lib/rmt/mirror.rb
@@ -64,9 +64,20 @@ class RMT::Mirror
       uri.query = @repository.auth_token if @repository.auth_token
 
       request = RMT::HttpRequest.new(uri, method: :head, followlocation: true)
-      request.on_success do
-        return key
+
+      request.on_complete do |response|
+        # 404 is an expected result here, we just continue to the next repo type
+        if response.success?
+          return key
+        elsif response.timed_out?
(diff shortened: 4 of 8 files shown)


lib/rmt/cli/systems.rb
```diff
@@ -4,6 +4,7 @@ class RMT::CLI::Systems < RMT::CLI::Base
   # we maintain on scc-docs for more information
   # (https://github.com/SUSE/scc-docs/blob/master/projects/scc/architecture/glossary.md#inactive).
   INACTIVE = 3.months
+  BATCH_SIZE = 20
```
rjschwei: That appears like a low number, how was it determined?

lib/rmt/cli/systems.rb
```diff
@@ -84,25 +85,54 @@ def remove(target)
...
+      border_bottom = sliced_systems_ids.last == systems_last
+      first_row = sliced_systems_ids.first == systems_first
+      style[:border_bottom] = border_bottom
+      style[:border_top] = first_row
+      puts decorator_systems.to_table(add_headers: first_row, style: style)
+      sliced_systems.destroy_all unless keep
+    end
+  end
+
+  def max_table_widths
```
rjschwei: `widths` -> `width`

lib/rmt/cli/systems.rb
```diff
@@ -84,25 +85,54 @@ def remove(target)
...
 
-    systems.destroy_all
+    System.where(id: systems).in_batches(&:destroy_all)
     puts "Purged systems that have not contacted this RMT since #{before}."
   end
 
   protected
 
+  def handle_rows(systems, keep: true)
+    maxs_widths = max_table_widths
```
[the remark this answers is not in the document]
jesusbv: happy to change it for a better name, the variable represent the max width of multiple values
it is an array where each element is a max width

lib/rmt/cli/systems.rb
digitaltom: I'm not clear on why this max width calculation is necessary, and how it solves the performance of listing the systems. 
The only rendered column with custom data is the hostname, and we could simply enforce that to 20 characters and go with a static table layout?

db/migrate/20260108092033_add_index_to_systems_hostname.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToSystemsHostname < ActiveRecord::Migration[5.1]
+  def change
+    add_index :systems, :hostname, unique: false
```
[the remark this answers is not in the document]
jesusbv: [here](https://github.com/SUSE/rmt/pull/1405/files#diff-10975e2473aab93b67bd13473f920034f1e7265445058d20718d8d08738cd62cR127)

tested in test infra AWS which has ~800K systems without the index it takes ~48 seconds to get the max value for that field

with the index, in the same set, ~3.2 seconds 

if we do not want the field to be indexed, it is going to take a lot longer in prod with millions of rows

lib/rmt/cli/systems.rb
jesusbv: if the max width is not calculated, the table printed by slices will have different sizes per column and in total
as currently one can not set the width of the cell but the whole table

if we are fine with different sizes like


Screenshot from 2026-01-09 19-45-24

of course, it can look worse that that,
then we do not need to calculate the max width for the table (and we would not need the `hostname` index)
I assumed we would want the table to not look like that

package/obs/rmt-server.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 13 17:14:09 UTC 2026 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
+
+- Version 2.25
+  * rmt-server-pubcloud:
+    * fix rmt-cli list and purge commands for large data (bsc#1253146 and bsc#1253147)
```
rjschwei: This is not Public Cloud specific. Anyone that has a very large number of systems in the table will run into this issue.

digitaltom (approved):

PR #1466, opened 2026-05-07
Refactor registry generalize

jesusbv:
## Description

The implementation of the registry was done in a SUSE specific way

we should remove any SUSE specific concepts in the registry setup that is part of the update infrastructure and make it generic, that 3rd party cloud providers that run their own update infrastructure can provide the same service to their PAYG customers as we do

* Related Issue / Ticket / Trello card: [PCT-1126](https://jira.suse.com/browse/PCT-1126)

## How to test 

registry access should work the same as before 

## Change Type

## Checklist

## Review

### DIFF
--- a/config/rmt.yml
+++ b/config/rmt.yml
@@ -57,7 +57,7 @@ connect_api:
 redis:
   url: <%= ENV.fetch('REDIS_URL', nil) %>
 
-# Optional registry setup. If used the values cannot be empty
+# Optional registry engine setup. If used the values cannot be empty
 # registry:
 #   service:
 #   realm:
--- a/config/rmt.yml
+++ b/config/rmt.yml
@@ -56,6 +56,8 @@ connect_api:
 
 redis:
   url: <%= ENV.fetch('REDIS_URL', nil) %>
-registry:
-  service:
-  realm:
+
+# Optional registry setup. If used the values cannot be empty
+# registry:
+#   service:
+#   realm:
--- a/config/rmt.yml
+++ b/config/rmt.yml
@@ -56,3 +56,6 @@ connect_api:
 
 redis:
   url: <%= ENV.fetch('REDIS_URL', nil) %>
+registry:
+  service:
+  realm:
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -2,7 +2,6 @@ module Registry
   class RegistryAuthError < ArgumentError; end
 
   class RegistryController < Registry::ApplicationController
-    REGISTRY_SERVICE = 'SUSE Linux OCI Registry'.freeze
     REGISTRY_API_VERSION = 'registry/2.0'.freeze
 
     before_action :set_requested_scopes, except: [ :catalog ]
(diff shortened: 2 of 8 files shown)


config/rmt.yml
```diff
@@ -56,3 +56,6 @@ connect_api:
 
 redis:
   url: <%= ENV.fetch('REDIS_URL', nil) %>
+registry:
+  service: 'SUSE Linux OCI Registry'
```
jesusbv: we shall probably remove the `service` and `realm` values from that file and populate it through SRE

jesusbv: done

jcomello (approved):

engines/registry/config/environments/production.rb
```diff
@@ -16,6 +16,8 @@ class Application < Rails::Application
       config.registry_public_key = config.registry_private_key.public_key
     end
     config.access_policies = ACCESS_POLICIES_PATH
+    config.registry_service = Settings[:registry].try(:service)
+    config.registry_realm = Settings[:registry].try(:realm)
```
digitaltom: Shouldn't this go to `engines/registry/lib/registry/engine.rb`? It's engine specific code, and should be set for all environments.

engines/registry/config/environments/test.rb
```diff
@@ -8,4 +8,6 @@
   config.registry_private_key = OpenSSL::PKey::RSA.new(2048)
   config.registry_public_key = config.registry_private_key.public_key
   config.autoload_paths << Rails.root.join('engines/registry/spec/support/')
+  config.registry_service = 'SUSE Linux OCI Registry'
+  config.registry_realm = 'SUSE Registry Authentication'
```
digitaltom: Is this needed? I think you're reading the values from Settings now.

jesusbv: no, I removed the one in prod, left that one behind

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -69,7 +68,8 @@ def basic_auth
       # skip authentication if this is not a login request
       return unless request.authorization
 
-      authenticate_or_request_with_http_basic('SUSE Registry Authentication') do |login, password|
+      realm = Settings.try[:registry].try(:realm) rescue ''
```
rjschwei: I think I understand the idea, i.e. have the `registry` setting in the config by default. However, that forces us to deal with empty values and not knowing if a user of RMT wanted to run a registry or if the entry is simply there because it is in the config by default and empty.

It makes me wonder whether or not we should have this as a default in the config. Maybe we should have the registry config commented out, i.e.

```
# Optional registry setup. If used the values cannot be empty
# registry:
#   service:
#   realm:
``` 
in rmt.yml and then generate an error here if the value is there but is empty.

jesusbv: should we update also the [documentation](https://documentation.suse.com/sles/15-SP6/html/SLES-all/rmt-preface.html) or with the optional commented block is enough ?

rjschwei: Hmm, the config is now optional but we still catch the error, with `rescue ''` which means we then call `authenticate_or_request_with_http_basic('') do |login, password|` which will then trigger some unrelated error that will be hard to debug.

I think what we want is:

```
raise "registry configured in rmt.conf but realm not set"
```

rjschwei (approved):

config/rmt.yml
```diff
@@ -56,3 +56,8 @@ connect_api:
 
 redis:
   url: <%= ENV.fetch('REDIS_URL', nil) %>
+
+# Optional registry setup. If used the values cannot be empty
```
digitaltom:
```suggestion
# Optional registry engine setup. If used the values cannot be empty
```

digitaltom (approved):

PR #1470, opened 2026-05-21
Add Ansible playbook for RMT deployment

ngetahun:
## Description

This PR adds an Ansible playbook for automated RMT server deployment on SLES 16+ and openSUSE Leap 16+.

### What's New

- **New Package**: `ansible-rmt-server` (noarch, SLES/Leap 16+ only)
- Automated installation: RMT server, MariaDB, Nginx, SSL certificates
- Configurable via `group_vars/all.yml` (SCC credentials, repo URL, database password)
- Integrated with Makefile: `make install`, `make lint`, `make test`, `make deploy`

* Related Issue / Ticket / Trello card: N/A

### 1. Build and verify packaging
```bash
make build-tarball
tar -tjf package/obs/rmt-server-3.0.alpha.tar.bz2 | grep ansible/
```

### 2. Build RPM (OBS)
```bash
cd ~/Documents/obs_workspace/systemsmanagement:SCC:RMT/rmt-server-3
cp package/obs/* .
osc build SLE_16.0 x86_64 --no-verify -k .
```
Expected: `ansible-rmt-server-3.0.alpha-0.noarch.rpm` (~65 KB, noarch)

### 3. Test from source
```bash
make install  # Install Ansible dependencies
make lint     # Validate playbooks
make test     # Run tests

cd ansible
cp group_vars/all.yml.example group_vars/all.yml
vim group_vars/all.yml  # Add SCC credentials

cd ..
make check-ansible  # Dry run
make deploy         # Deploy RMT
```

### 4. Verify deployment
```bash
sudo systemctl status rmt-server nginx mariadb
sudo ls -la /etc/rmt/ssl/
sudo rmt-cli sync
sudo rmt-cli products list --all | head
```

### 5. Test configurability
Edit `ansible/group_vars/all.yml`:
```yaml
rmt_repo_url: https://download.opensuse.org/repositories/systemsmanagement:/SCC:/RMT/SLE_16.0/
rmt_db_password: "custom-password"
```
Re-run: `ansible-playbook site.yml`

### 6. Test idempotency
```bash
ansible-playbook site.yml  # First run
ansible-playbook site.yml  # Second run (should show changed=0)
```

## Change Type

## Checklist

## Review


ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,122 @@
...
+    #password: rmt
+    #priv: 'rmt.*:ALL'
+    #state: present
+  #no_log: true
+
+# Restart RMT
+- name: Restart rmt-server service
+  service:
+    name: rmt-server.service
+    state: restarted
```
deveshsharma63: I'm not entirely sure why we're using handlers above when rmt-server is also being explicitly restarted at the end of this task file. One thing to note is that the explicit restart task like this will execute at the end of this particular .yml file, whereas handlers will run at the end of the entire Ansible role/play execution.

If the intention is to always restart rmt-server immediately after this task file finishes, we could remove the handler notifications. Otherwise, if we're okay restarting the service only when a task reports a changed state & towards the end of the role execution, we could simply remove the final explicit restart task instead.

ansible/roles/rmt/vars/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# RMT role variables (not meant to be overridden)
+
+rmt_packages:
+  - rmt-server-3
+  - mariadb
+  - nginx
+
+rmt_public_dir: /var/lib/rmt/public
```
deveshsharma63: I'm not entirely certain how you've differentiated the variables in the main.yml files between the built-in role directories defaults/ & vars/.

ansible/ansible.cfg
```diff
@@ -0,0 +1,18 @@
+[defaults]
+inventory = localhost,
```
deveshsharma63: I hope you're treating localhost here as an inline inventory host list.

package/obs/rmt-server.spec
```diff
@@ -210,6 +225,16 @@ install -D -m 644 package/files/nginx-pubcloud/nginx-https.conf %{buildroot}%{_s
...
 install -D -m 644 package/files/nginx-pubcloud/auth-location.conf %{buildroot}%{_sysconfdir}/nginx/rmt-auth.d/auth-location.conf
 
+# ansible playbook (SLES/Leap 16+ only)
+%if (0%{?sle_version} >= 160000) || (0%{?is_opensuse} && 0%{?suse_version} >= 1600)
+mkdir -p %{buildroot}%{_datadir}/ansible/rmt
+cp -r ansible/* %{buildroot}%{_datadir}/ansible/rmt/
+# Remove build artifacts (tests already excluded from tarball)
+find %{buildroot}%{_datadir}/ansible/rmt -type d -name "__pycache__" -exec rm -rf {} + 2>/dev/null || true
+find %{buildroot}%{_datadir}/ansible/rmt -name "*.pyc" -delete 2>/dev/null || true
+find %{buildroot}%{_datadir}/ansible/rmt -name "*.retry" -delete 2>/dev/null || true
```
felixsch: Is this generated when running the build process? If not, I would move this cleanup into the `Makefile` `dist` target. Then it never is part of the tarball, which does not pollute the source build.

ngetahun: noted

Makefile
```diff
@@ -1,24 +1,89 @@
...
+	@echo ''
+
+# =============================================================================
+# Installation
+# =============================================================================
+
+install: install-ansible ## Install all dependencies (Ansible)
+
+install-ansible: ## Install Ansible dependencies and collections
+	cd ansible && pip install ansible ansible-lint
```
rtamalin: If this is intended to just be installed for the current user to use, then potentially add the --user flag here to both indicate that facft, and to avoid potential for problematic side effects if ever run as root on a real system where installing Python packages via pip over distribution package manager installations can be problematic.

Also this will "succeed" so long as a version the specified Python packages is available on the system, which could be the package manager provided ones... If specific minimum versions are required, it may be worth adding appropriate version requirements to the package specifications.

Makefile
```diff
@@ -1,24 +1,89 @@
...
+# =============================================================================
+# Testing & Validation
+# =============================================================================
+
+test: ## Run all tests (Ansible playbook tests)
+	@echo "==> Running Ansible playbook tests..."
+	cd ansible && ansible-playbook tests/test_playbook.yml
+	@echo "==> All tests passed!"
+
+lint: ## Validate and lint all code (Ansible playbooks)
```
rtamalin: Minor observation, which will depend on the SLE 16.x stream commitment to keeping Ansible up-to-date...

In the long run we may want to control the target ecosystem under which these commands are run, e.g. the rmt container based upon the target ecosystem version...

I suggest this because the version of ansible, ansible-core and ansible-lint you will have installed is somewhat dependent upon both the version of Python you have installed, and the distribution ecosystem you are currently running on...

E.g. if you are currently running this on Leap 16.0 you would be working with Ansible 11.x and ansible-core 2.18.x, while if you are currently running on Tumbleweed, you might be working with Ansible 13.x and ansible-core 2.20.x, and the linting results for the latter may be different, and stricter than the former, and may even run the risk of requiring deprecation changes to be applied for the latest version...

ansible/roles/rmt/defaults/main.yml
```diff
@@ -0,0 +1,60 @@
...
+rmt_ca_cert_path: /etc/rmt/ssl/rmt-ca.crt
+rmt_ssl_cert_validity_days: 3650  # 10 years
+rmt_ssl_key_size: 4096
+
+# Service configuration
+rmt_enable_http: true
+rmt_enable_https: true
+
+# Python pip executable for dependencies
+rmt_pip_executable: /usr/bin/pip3.11
```
rtamalin: This is not valid for a SLE 16.1 or SLE 16.0 system; the Python 3.13 ecosystem is what is installed in SLE 16.1 (or SLE 16.0) so the correct specific path would be `/usr/bin/pip3.13`...

However, that is release specific, and probably should just be /usr/bin/pip3 to avoid problems like this...

But on a good note this fact `rmt_pip_executable` also doesn't appear to be used/referenced, so potentially can be dropped...

rtamalin (changes requested):
rmt_pip_executable path is invalid for SLE 16.x ecosystems... But also doesn't appear to be used...

ansible/roles/rmt/vars/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# RMT role variables (not meant to be overridden)
```
rtamalin: Not sure what the logic here but `vars` file defined variables are usually considered the ones that are "expected" to be superceded/overridden at run time... As @deveshsharma63 alludes to below...

The usual logical concept is that `defaults` holds the equivalent of the compiled in default values for things, and `vars` entries are frequently defined in terms of a possibly undefined variable defaulting back to a value defined in `defaults`, e.g. a simplistic example might be

```
# in group_vars/all.yml
rmt_var_system: "system_default_value"

# in roles/rmt/defaults/main.yml
rmt_var_default: "some_default_value"

# in roles/rmt/vars/main.yml
rmt_var: "{{ rmt_var_system | default(rmt_var_default) }}"
```

This allows for the default value to be defined and used if the potential system variable hasn't been defined... Or it could allow for a `group_vars/xxxx.yml` group specific or `host_vars/yyyy.yml` host specific override to be configured...

ansible/roles/rmt/vars/main.yml
```diff
@@ -0,0 +1,13 @@
...
+
+rmt_packages:
```
rtamalin: Doesn't appear to be referenced in the role implementation...

In fact none of the variables specified in this file are referenced anywhere else in the role, or the entire ansible tree...

```
% grep -rI -e "rmt_packages" -e "rmt_public_dir" -e "rmt_config_path" -e "rmt_nginx_vhost_dir" -e "rmt_systemd_dir" ansible
ansible/roles/rmt/vars/main.yml:rmt_packages:
ansible/roles/rmt/vars/main.yml:rmt_public_dir: /var/lib/rmt/public
ansible/roles/rmt/vars/main.yml:rmt_config_path: /etc/rmt.conf
ansible/roles/rmt/vars/main.yml:rmt_nginx_vhost_dir: /etc/nginx/vhosts.d
ansible/roles/rmt/vars/main.yml:rmt_systemd_dir: /etc/systemd/system
```

As such you can probably just drop this file entirely ;-)

package/obs/rmt-server.spec
```diff
@@ -116,6 +116,21 @@ Conflicts:      rmt-server-configuration
...
+Group:          Development/Tools/Other
+BuildArch:      noarch
+Requires:       ansible-core >= 2.9
+Recommends:     rmt-server = %version
+
+%description -n ansible-rmt-server
+Ansible playbook and roles for automated deployment and configuration
+of RMT server on localhost. Includes SSL certificate generation,
+database setup, and nginx configuration.
+%endif
```
rtamalin, line 144: There should be a an appropriate Requires on the `python313-PyMySQL` (or ecosystem equivalent) package defined in this section using standard SLE Python RPM packaging macros, which will ensure that the PyMySQL module is available for the Ansible playbooks to leverage.

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+    name:
+      - rmt-server
+      - mariadb
+      - nginx
+    state: latest
+
+- name: Ensure PyMySQL is installed for Python 3.11
+  ansible.builtin.pip:
+    name: PyMySQL
+    executable: /usr/bin/pip3.11
```
rtamalin: Another python version specific reference... This should either match the correct version available for SLE 16.x systems, or use `/usr/bin/pip3` to avoid Python3 version dependencies...

Also are we sure we want to be pip installing anything given that we have a zypper installable package available, python313-PyMySQL, that should satisfy this need?

My recommendation would be to drop this action in favour of adding the appropriate package specifier to the above package list, and an appropriate Requires directive to the spec files definition of the ansible playbooks subpackage.

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+    owner: root
+    group: root
+  notify: Reload nginx
+
+- name: Install rmt & dependencies
+  zypper:
+    name:
+      - rmt-server
+      - mariadb
+      - nginx
```
rtamalin: This list should also include a package specifier that pulls in the Python ecosystem appropriate version of the PyMySQL module, e.g.


```suggestion
      - nginx
      - python3-PyMySQL
```

Also this action should really only be relevant for "developer" mode installs using these Ansible playbooks... If installing rmt-server as a customer would, via package manager installs, the spec file definition for the ansible playbooks subpackage should have a Requires: dependency that pulls in the appropriate package to provide PyMySQL.

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+    state: latest
+
+- name: Ensure PyMySQL is installed for Python 3.11
+  ansible.builtin.pip:
+    name: PyMySQL
+    executable: /usr/bin/pip3.11
+# Uncomment this when creating fresh setup with empty database
```
rtamalin: If the intent is to retain this code in the long run it would be better to create a block section that is guarded by a when condition that disables it by default, and then override that when condition's check inputs to enable the block if we want to trigger a fresh setup with an empty DB...

ansible/roles/rmt/tasks/main.yml
rtamalin: Further this dependency is only required if we want to support the commented out fresh setup section below... If not we can drop it, and the need for the spec file requires dependency below...

ansible/requirements.yml
```diff
@@ -0,0 +1,13 @@
+---
+# External Ansible roles and collections required for RMT deployment
+collections:
+  - name: ansible.posix
+    version: ">=1.5.0"
+  - name: community.general
+    version: ">=7.0.0"
+  - name: community.mysql
```
rtamalin, line 18: Potentially this requirement is unnecessary if we aren't using  any `community.mysql` actions, both of which are currently commented out...

ansible/roles/rmt/tasks/main.yml
rtamalin: Any we only would need this if we actual enable the next 2 actions that are currently commented out below... If not then we don't appear to need PyMySQL at  all... And we wouldn't need the `community.mysql` collection in the `requirements.yml` either, or the test_playbook's check that the `community.mysql` collection is installed...

package/obs/rmt-server.spec
rtamalin: Note that this may not be needed if we don't care about the 2 associated actions which are currently commented out at the end of the rmt role's tasks/main.yml

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+
+  tasks:
+    - name: Verify Ansible version
+      ansible.builtin.assert:
+        that:
+          - ansible_version.full is version('2.9', '>=')
+        fail_msg: "Ansible version must be 2.9 or higher"
+        success_msg: "Ansible version check passed"
+
+    - name: Check if community.mysql collection is installed
```
rtamalin: This check is potentially not needed since the actions that leverage this collection are currently commented out.

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+  vars:
+    rmt:
+      RMT_ORGANIZATION_USERNAME: "test-user"
+      RMT_ORGANIZATION_PASSWORD: "test-pass"
+
+  tasks:
+    - name: Verify Ansible version
+      ansible.builtin.assert:
+        that:
+          - ansible_version.full is version('2.9', '>=')
```
rtamalin: FYI this is a fairly ancient version of Ansible, and may not not actually be compatible with the Python 3 ecosystems available on SLE 16.x... This was the last version before the re-org of Ansible into the modern style of the `ansible` package being akin to a "meta" package that depends on specific `ansible-core` version stream package + supported community collections...

ansible/roles/rmt/meta/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+galaxy_info:
+  author: SUSE RMT Team
+  description: Ansible role to install and configure SUSE RMT (Repository Mirroring Tool) server
+  company: SUSE
+  license: MIT
+  min_ansible_version: "2.9"
```
rtamalin: FYI this is a fairly ancient version of Ansible, and may not not actually be compatible with the Python 3 ecosystems available on SLE 16.x... This was the last version before the re-org of Ansible into the modern style of the `ansible` package being akin to a "meta" package that depends on specific `ansible-core` version stream package + supported community collections...

ansible/tests/inventory.yml
```diff
@@ -0,0 +1,6 @@
+---
```
rtamalin: Not sure that this file is needed since it doesn't appear to be referenced anywhere...

The Makefile target rules below don't reference it...

Makefile
```diff
@@ -1,24 +1,89 @@
...
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-z-]+:.*?## / {printf "  %-20s %s\n", $$1, $$2}' $(MAKEFILE_LIST) | grep -E "^  (dist|build-tarball|man|public_repo) "
+	@echo ''
+
+# =============================================================================
+# Installation
+# =============================================================================
+
+install: install-ansible ## Install all dependencies (Ansible)
+
+install-ansible: ## Install Ansible dependencies and collections
```
rtamalin: After having reviewed more of the overall PR I'm not sure that I understand the real relevance/benefit of the `install-ansible` target... Yes it ensures a viable ansible is available/installed locally, but that is for the "local" distribution ecosystem, and won't test things in the context of the target distribution ecosystem, e.g. SLE 16.x.

Similarly for the `deploy-ansible` target...

It seems to me that we would be better off crafting target rules that install the desired Ansible ecosystem into the `rmt` container environment, and run the test playbooks there, and similarly deploy the ansible playbooks into the `rmt` container environment, and run the site.yml --check there, and test running the site.yml within the rmt container...

That would exercise things in an environment that is likely to correspond to the actual customer install environment...

ansible/requirements.yml
rtamalin: Also note that this dependency is satisfied by the Ansible 13 or Ansible 11 that are currently available in SLE 16.1 BETA and SLE 16.0, which come with versions 4.0.1 and 3.12.0 respectively.

ansible/requirements.yml
```diff
@@ -0,0 +1,13 @@
...
+    version: ">=3.5.0"
+  - name: community.crypto
```
rtamalin, line 20: This dependency is satisfied by the Ansible 13 or Ansible 11 that are currently available in SLE 16.1 BETA and SLE 16.0, which come with versions 3.1.0 and 2.25.0 respectively.

ansible/requirements.yml
```diff
@@ -0,0 +1,13 @@
+---
+# External Ansible roles and collections required for RMT deployment
+collections:
+  - name: ansible.posix
+    version: ">=1.5.0"
+  - name: community.general
```
rtamalin, line 16: This dependency is satisfied by the Ansible 13 or Ansible 11 that are currently available in SLE 16.1 BETA and SLE 16.0, which come with versions 12.3.0 and 10.4.0 respectively.

ansible/requirements.yml
```diff
@@ -0,0 +1,13 @@
+---
+# External Ansible roles and collections required for RMT deployment
+collections:
+  - name: ansible.posix
```
rtamalin, line 14: This dependency is satisfied by the Ansible 13 or Ansible 11 that are currently available in SLE 16.1 BETA and SLE 16.0, which come with versions 2.1.0 and 1.6.2 respectively.

ansible/requirements.yml
```diff
@@ -0,0 +1,13 @@
+---
```
rtamalin, line 1: Technically this dependencies are satisfied by depending up Ansible >=11 (which is what SLE 16.0 comes with currently) or Ansible >=13 (which is what SLE 16.1 BETA currently comes with...

rtamalin: As such there technically shouldn't be a need to run an `ansible-galaxy collections install -r requirements.yml` if installing rmt-server within a SLE 16.x ecosystem ;-)

Makefile
```diff
@@ -74,6 +140,12 @@ build-tarball: clean man
 	# don't package example data export handler
 	@rm -rf $(NAME)-$(VERSION)/engines/data_export/lib/data_export/handlers/example.rb
 
+	# don't package ansible tests and Python artifacts
+	@rm -rf $(NAME)-$(VERSION)/ansible/tests
```
rtamalin: There may actually be a benefit to packaging the tests so that they are available to the build service to run a check step that confirms that the required ansible collections are available with acceptable versions for the target stream being built...

Basically if add an appropriate BuildRequires, e.g.

```
BuildRequires: ansible >= 11
```
to the subpackage definition, and then add a %check section that runs the the test playbook to confirm that the appropriate collections are available with compatible versions, we can make the package build sanity check that the desired collections are provided by the target streams Ansible ecosystem.

And if we don't care about packaging the test playbook we can just have the %install section drop them.

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+        fail_msg: "Ansible version must be 2.9 or higher"
+        success_msg: "Ansible version check passed"
+
+    - name: Check if community.mysql collection is installed
+      ansible.builtin.command: ansible-galaxy collection list community.mysql
```
rtamalin: We could enhance these collection checks to include a step that extracts the associated versions and then verify that the available versions are compatible with the desired/target versions...

ansible/roles/rmt/defaults/main.yml
rtamalin: Yeah as my further digging identified below we likely don't need to care about running pip given that:
1. the relevant Python dependency, PyMySQL, can be satisfied by an available distribution package
2. we probably shouldn't be doing any pip installing anyway so as part of a package based installation workflow
3. the relevant Ansible actions that depend on the PyMySQL dependency are commented out

ansible/roles/rmt/meta/main.yml
```diff
@@ -0,0 +1,24 @@
...
+
+  platforms:
+    - name: openSUSE
+      versions:
+        - Leap16
```
rtamalin: The version of ansible available in Leap/SLE 16.0 is currently 11.x

ansible/roles/rmt/meta/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: SLES
+      versions:
+        - 16
```
rtamalin: May want to add 16.1 to this list...

ansible/roles/rmt/meta/main.yml
rtamalin: And probably should add 16.1 to this list...

ansible/roles/rmt/templates/rmt.conf.j2
```diff
@@ -0,0 +1,41 @@
+database:
```
rtamalin, line 1: There probably should be a comment block included at the start of this file saying it is a generated file and that any modifications will be overwritten when running the ansible site.yml playbook, and that instead the relevant input settings the the group_vars/all.yml should be modified to make persistent changes...

ansible/roles/rmt/files/rmt_host_http_ssl.conf
```diff
@@ -0,0 +1,51 @@
+server {
```
rtamalin, line 1: Not sure why this is a template rather than just a fixed file since there don't appear to be any templating actions present the file?

Also, there probably should be a comment block included at the start of this file saying it is a generated file and that any modifications will be overwritten when running the ansible site.yml playbook, and that instead the relevant input settings the the group_vars/all.yml should be modified to make persistent changes...

ansible/roles/rmt/files/rmt_host_plain_text.conf
```diff
@@ -0,0 +1,51 @@
+server {
```
rtamalin, line 1: Same points as above for the http_ssl variant of this...

ansible/roles/rmt/files/rmt_upstreams.conf
```diff
@@ -0,0 +1,3 @@
+upstream rmt {
```
rtamalin, line 1: Again, not sure why this is a template at all since there is no templating happening, and it's name doesn't even follow the same naming scheme as the other files...

And again it should include a comment at the start that it is a generated file and that any changes will be lost when the file is regenerated by the playbook run.

ansible/group_vars/all.yml
```diff
@@ -0,0 +1,25 @@
...
+# Configuration for RMT server on localhost
+
+# SCC credentials - REQUIRED
+# Set your SUSE Customer Center organization credentials
+rmt:
+  RMT_ORGANIZATION_USERNAME: "your-scc-organization-username"
+  RMT_ORGANIZATION_PASSWORD: "your-scc-organization-password"
+
+# Override default RMT configuration if needed
+# See group_vars/all.yml.example for all available options
```
rtamalin: Since you have moved all of the variable definitions to the roles/rmt/defaults/main.yml, this should probably refer to that file instead of the group_vars/all.yml.example.

And the group_vars/all.yml.example should probably align with this file now, servicing as a reference copy in case this file gets trashed...

ansible/group_vars/all.yml.example
```diff
@@ -0,0 +1,55 @@
+---
```
rtamalin, line 1: This file's content should probably align with the companion all.yml, right?

ansible/roles/rmt/meta/main.yml
rtamalin: Good spot it is the ansible-core version that needs to be specified, not the ansible community package version.

ansible/roles/rmt/tasks/main.yml
rtamalin: Underlying code has been addressed - resolving

ansible/roles/rmt/tasks/main.yml
rtamalin: Resolving as underlying issue has been addressed.

ansible/roles/rmt/tasks/main.yml
rtamalin: Resolving as underlying code issue has been addressed.

ansible/roles/rmt/vars/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# RMT role variables (not meant to be overridden)
+# rmt_public_dir: /var/lib/rmt/public
+# rmt_config_path: /etc/rmt.conf
+# rmt_nginx_vhost_dir: /etc/nginx/vhosts.d
+# rmt_systemd_dir: /etc/systemd/system
+
+rmt_repo_url: https://download.opensuse.org/repositories/systemsmanagement:/SCC:/RMT/SLE_15_SP7/
```
rtamalin: These two values are the same as those specified in the defaults/main.yml, so there no real benefit to specifying it here, and more likely the potential for maintenance pain and confusion going forward... E.g. if we decide we want to change the default rmt_repo_url or rmt_repo_name, we have to change it in both places, but only this one will actually count/have an effect...

You don't need to specify a variable here to allow someone to be able to override it through other means; they can override it whether it is listed here or not.

As previously mentioned the use of the `vars/main.yml` is more to provide a place where a role can define an important fact that it wants to use in terms of other input values/sources... For example, you could implement logic like use x if provided, else use y if provided or default to z. Or another common pattern would be to use jinja templating to process an input dict of dicts, dict of lists, or list of dicts to extract a list of the specific values from each entry or subentry so that you can interate over those items, possibly more than once, in an action in your tasks/ code without having to implement the ugly jinja templating for each action...

But just repeating the same `var: value` here as already specified in the defaults/main.yml has no benefit; just relying on the defaults/main.yml will to provide the value will work...

ansible/tests/test_playbook.yml
rtamalin: This has been addressed

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+  vars:
+    rmt:
+      RMT_ORGANIZATION_USERNAME: "test-user"
+      RMT_ORGANIZATION_PASSWORD: "test-pass"
+
+  tasks:
+    - name: Verify Ansible version
+      ansible.builtin.assert:
+        that:
+          - ansible_version.full is version('2.18', '>=')
```
rtamalin: One thing to consider here is putting these version values in the vars block above, and then referring to them by the variable name here and in the fail message on the next line...

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+        fail_msg: "Ansible version must be 2.18 or higher"
+        success_msg: "Ansible version check passed"
+
+    - name: Check if ansible.mysql collection is installed
```
rtamalin: Looks like `community.mysql` as been adopted into the main ansible collections as `ansible.mysql` going forward, with redirection from `community.mysql` to `ansible.mysql`, as of early May 2026, and will be dropped from the Ansible community releases starting with v17.

Currently `ansible.mysql` is not part of the standard collections included as part of the `ansible` package in SLE 16.0 (v11.3) or SLE 16.1 BETA (v13.3).

`ansible.mysql` is included as part of the current openSUSE Tumbleweed `ansible` v13.7 package, apparently leveraging the redirection mechanism there; not sure if it will make it into SLE 16.1 FCS...

As such we should probably aim to primarily support the `community.mysql` collection for now. We can add some conditional logic plumbing to manage which version we use; the simplest may be to implement two equivalent task list files, e.g. tasks/mysql.yml and tasks/community_mysql.yml, and conditionally include the appropriate task list file with an `include_tasks` or `import_tasks` action.

ansible/tests/test_playbook.yml
rtamalin: Thinking about this more the simplest way to do this may be to run `ansible-galaxy collection list --format=json` and use a `set_fact` action in conjunction with the `from_json` filter to import the output as "native" dict of dicts which we can easily check against in subsequent actions.

If we store the required collections + versions info as a dict in a variable in the vars section we could then iterate over that required versions dict checking for matching entries in the available collections dict...

ansible/tests/test_playbook.yml
```diff
@@ -0,0 +1,140 @@
...
+  hosts: localhost
+  connection: local
+  gather_facts: true
+  become: false
+
+  vars:
+    rmt:
+      RMT_ORGANIZATION_USERNAME: "test-user"
+      RMT_ORGANIZATION_PASSWORD: "test-pass"
+
```
rtamalin, line 21: If you wanted to we could list version requirements for ansible and the collections here, and then write the code below to verify against these versions.

```suggestion
    min_ansible_version: "2.18"
    min_collection_versions:
        "ansible.posix": "1.5.0"
        "community.general": "7.0.0"
        "ansible.mysql": "3.5.0"
        "ansible.crypto": "2.0.0"

```

ansible/requirements.yml
rtamalin: Note my discussion below related to `community.mysql` becoming `ansible.mysql` but this is a very bleeding edge change, and not currently available in the SLE 16.0 or SLE 16.1 BETA streams...

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+    state: latest
+
+- name: Ensure MariaDB service is started and enabled
+  ansible.builtin.systemd:
+    name: mariadb
+    state: started
+    enabled: true
+
+- name: Create RMT database
+  ansible.mysql.mysql_db:
```
rtamalin: See my discussion below in the test_playbook.yml comments re `community.mysql` transitioning to `ansible.mysql`...

Given that `ansible.mysql` is not currently available to the SLE 16.0 and SLE 16.1 streams as part of the collections included in the `ansible` package, we should revert to using the `community.mysql` actions here...

```suggestion
  community.mysql.mysql_db:
```

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+    name: "{{ rmt_db_name }}"
+    state: present
+    encoding: "{{ rmt_db_encoding }}"
+    login_unix_socket: /run/mysql/mysql.sock
+
+- name: Create RMT database user
+  ansible.mysql.mysql_user:
```
rtamalin: Similar to above we should revert to the community.mysql collection version of this action for now.

```suggestion
  community.mysql.mysql_user:
```

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+      - nginx
+      - python3-PyMySQL
+    state: latest
+
+- name: Ensure MariaDB service is started and enabled
+  ansible.builtin.systemd:
+    name: mariadb
+    state: started
+    enabled: true
+
```
rtamalin, line 19: My recommendation for these mysql actions would be to move them to a separate tasks file, e.g. `mysql_setup.yml` or similar...

Further to address the need to potentially support have both the ansible.mysql and community.mysql actions available to use, we could have a second `community_mysql_setup.yml` task list for those versions.

And then we could dynamically determine which task list we need to use, e.g. looking at the `ansible_version` or a combination of `ansible_distribution_version` and `ansible_distribution` facts, using a `set_fact` action to assign the appropriate task list name to a variable.

Next we could pull those actions in to the main task list using an `include_tasks` or `import_tasks` targetting the task list specified by the variable...

ansible/roles/rmt/vars/main.yml
ngetahun: Most of the variables are meant for cockpit to read and make changes to. But I have removed the placeholders.

Makefile
ngetahun: removed

package/obs/rmt-server.spec
ngetahun: done

ansible/requirements.yml
ngetahun: I have left the file and commented out the packages. If we want to target debian based distros, the file might come in handy in the future.

ansible/requirements.yml
ngetahun: keeping the file for posterity

Makefile
ngetahun: Let's pin this for the future as it is not needed to have OBS based tests right now.

ansible/roles/rmt/meta/main.yml
ngetahun: 16.1 doesn't exist yet

ansible/roles/rmt/files/rmt_host_http_ssl.conf
ngetahun: noted

ansible/roles/rmt/files/rmt_upstreams.conf
ngetahun: noted

ansible/roles/rmt/vars/main.yml
ngetahun: well leap and sles 16 have different repo structures, hence why I left it as a variable

ansible/roles/rmt/templates/rmt.conf.j2
ngetahun: actually, the original file `rmt.conf` is meant to be edited. What needs to be done is make the playbook aware of the user changes that exist already

ansible/roles/rmt/tasks/main.yml
ngetahun: From what I gather, `community.mysql` is dependent on `ansible.mysql` meaning that if we have `community.mysql` it will by default pull `ansible.mysql`. No need to split the plugin support yet until `community.mysql` is made obsolete

ansible/group_vars/all.yml
ngetahun: I will speak with alice on where to move the vars for better cockpit integration.

ansible/roles/rmt/tasks/ssl_setup.yml
```diff
@@ -0,0 +1,69 @@
...
+    path: /etc/rmt/ssl/rmt-ca.crt
+    privatekey_path: /etc/rmt/ssl/rmt-ca.key
+    provider: selfsigned
+    selfsigned_not_after: "+3650d"
+    mode: "0644"
+    owner: root
+    group: root
+
+- name: Provision HTTPS (SSL) nginx config
+  ansible.builtin.template:
```
rtamalin: So you moved these from templates/ to files/ and removed the `.j2` suffix, so this `ansible.builtin.template` so change to `ansible.builtin.copy`.

Sorry I missed this previously.

ansible/roles/rmt/tasks/ssl_setup.yml
```diff
@@ -0,0 +1,69 @@
...
+    src: rmt_host_http_ssl.conf.j2
+    dest: /etc/nginx/vhosts.d/rmt-server-https.conf
+    mode: "0644"
+    owner: root
+    group: root
+
+- name: Provision HTTP (Plain Text) nginx config
+  ansible.builtin.template:
```
rtamalin: Ditto - this needs to change to a copy referring to the file in the files/ area

rtamalin (changes requested):
Spotted a couple of ansible.builtin.template actions that need to change to ansible.builtin.copy as the associated input files have moved from the role's templates/ to files/ folder.

ansible/roles/rmt/tasks/ssl_setup.yml
```diff
@@ -0,0 +1,69 @@
...
+    privatekey_path: /etc/rmt/ssl/rmt-ca.key
+    provider: selfsigned
+    selfsigned_not_after: "+3650d"
+    mode: "0644"
+    owner: root
+    group: root
+
+- name: Provision HTTPS (SSL) nginx config
+  ansible.builtin.template:
+    src: rmt_host_http_ssl.conf.j2
```
rtamalin:
```suggestion
  ansible.builtin.copy:
    src: rmt_host_http_ssl.conf
```

ansible/roles/rmt/tasks/ssl_setup.yml
```diff
@@ -0,0 +1,69 @@
...
+    dest: /etc/nginx/vhosts.d/rmt-server-https.conf
+    mode: "0644"
+    owner: root
+    group: root
+
+- name: Provision HTTP (Plain Text) nginx config
+  ansible.builtin.template:
+    src: rmt_host_plain_text.conf.j2
```
rtamalin:
```suggestion
  ansible.builtin.copy:
    src: rmt_host_plain_text.conf
```

rtamalin (approved):
Looks great now!

rtamalin:
Added a patch to fix the feature test issue; ansible was not installed in the environment where the rpmbuild was running, and the spec file now has a BuildRequires for ansible >= 11, and ansible-core >= 2.18.

ansible/roles/rmt/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+  when: rmt_db_info is succeeded
+
+- name: Provision database
+  ansible.builtin.include_tasks: db_setup.yml
+  when: not (rmt_db_exists | default(false))
+
+- name: Provision SSL certificates and nginx configuration
+  ansible.builtin.include_tasks: ssl_setup.yml
+
+- name: Run database migrations
```
felixsch, line 64: I was running the ansible playbook on a 16.0 VM and got:

```
localhost:/usr/share/ansible/rmt # ansible-playbook site.yml
[WARNING]: No inventory was parsed, only implicit localhost is available

PLAY [Configure RMT server on localhost] *******************************************************

TASK [Validate required variables] *************************************************************
ok: [localhost]

TASK [rmt : Install rmt & dependencies] ********************************************************
ok: [localhost]

TASK [rmt : Ensure MariaDB service is started and enabled] *************************************
ok: [localhost]

TASK [rmt : Check if RMT database exists] ******************************************************
ok: [localhost]

TASK [rmt : Set fact for database existence] ***************************************************
ok: [localhost]

TASK [rmt : Provision database] ****************************************************************
included: /usr/share/ansible/rmt/roles/rmt/tasks/db_setup.yml for localhost

TASK [rmt : Ensure MariaDB service is started and enabled] *************************************
ok: [localhost]

TASK [rmt : Create RMT database] ***************************************************************
ok: [localhost]

TASK [rmt : Create RMT database user] **********************************************************
ok: [localhost]

TASK [rmt : Provision SSL certificates and nginx configuration] ********************************
included: /usr/share/ansible/rmt/roles/rmt/tasks/ssl_setup.yml for localhost

TASK [rmt : Create SSL directory] **************************************************************
ok: [localhost]

TASK [rmt : Generate RMT server private key] ***************************************************
ok: [localhost]

TASK [rmt : Generate RMT server certificate] ***************************************************
ok: [localhost]

TASK [rmt : Generate RMT CA private key] *******************************************************
ok: [localhost]

TASK [rmt : Generate RMT CA certificate] *******************************************************
ok: [localhost]

TASK [rmt : Provision HTTPS (SSL) nginx config] ************************************************
ok: [localhost]

TASK [rmt : Provision HTTP (Plain Text) nginx config] ******************************************
ok: [localhost]

TASK [rmt : Provision RMT Upstream section] ****************************************************
ok: [localhost]

TASK [rmt : Run database migrations] ***********************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to start service rmt-server-migration.service: Job for rmt-server-migration.service failed because the control process exited with error code.\nSee \"systemctl status rmt-server-migration.service\" and \"journalctl -xeu rmt-server-migration.service\" for details.\n"}

PLAY RECAP *************************************************************************************
localhost                  : ok=18   changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0

```

I edited `all.yml` before hand and set `rmt_db_username: rmt` and `rmt_db_password: rmt` + SCC credentials.

database is setup, user `rmt` exists with password `rmt`. Database is also created:
```

MariaDB [(none)]> SHOW DATABASES;
+--------------------+
| Database           |
+--------------------+
| information_schema |
| rmt                |
| test               |
+--------------------+
```

but the journal reveals:

```
Jun 10 11:15:46 localhost.localdomain rails[4635]: `/home/_rmt` is not a directory.
Jun 10 11:15:46 localhost.localdomain rails[4635]: Bundler will use `/tmp/bundler20260610-4635-cp2vy34635' as your home directory temporarily.
Jun 10 11:15:48 localhost.localdomain rails[4635]: There is an issue connecting to your database with your username/password, username: rmt.
Jun 10 11:15:48 localhost.localdomain rails[4635]: Please check your database configuration to ensure the username/password are valid.
Jun 10 11:15:48 localhost.localdomain rails[4635]: Couldn't create 'rmt' database. Please check your configuration.
Jun 10 11:15:48 localhost.localdomain rails[4635]: bin/rails aborted!
Jun 10 11:15:48 localhost.localdomain rails[4635]: ActiveRecord::DatabaseConnectionError: There is an issue connecting to your database with your username/password, username: rmt. (ActiveRecord::DatabaseConnectionError)
Jun 10 11:15:48 localhost.localdomain rails[4635]: Please check your database configuration to ensure the username/password are valid.
Jun 10 11:15:48 localhost.localdomain rails[4635]: /usr/share/rmt/bin/rails:9:in '<main>'
Jun 10 11:15:48 localhost.localdomain rails[4635]: Caused by:
Jun 10 11:15:48 localhost.localdomain rails[4635]: Mysql2::Error::ConnectionError: Access denied for user 'rmt'@'localhost' (using password: NO) (Mysql2::Error::ConnectionError)
Jun 10 11:15:48 localhost.localdomain rails[4635]: /usr/share/rmt/bin/rails:9:in '<main>'
Jun 10 11:15:48 localhost.localdomain rails[4635]: Tasks: TOP => db:create
Jun 10 11:15:48 localhost.localdomain rails[4635]: (See full trace by running task with --trace)
Jun 10 11:15:48 localhost.localdomain rails[4635]: INFO: RMT version 3.0
Jun 10 11:15:48 localhost.localdomain systemd[1]: rmt-server-migration.service: Main process exited, code=exited, status=1/FAILURE
```

felixsch: Did a fresh re-run and same outcome.

here the steps I took:

```
$ zypper in rmt-server-config-3.0-0.x86_64.rpm rmt-server-3.0-0.x86_64.rpm ansible-rmt-server-3.0-0.noarch.rpm
$ cd /usr/share/ansible/rmt
$ vim group_vars/all.yml
$ cat group_vars/all.yml
---
# Configuration for RMT server on localhost

# SCC credentials - REQUIRED
# Set your SUSE Customer Center organization credentials
rmt:
  RMT_ORGANIZATION_USERNAME: "<redacted>"
  RMT_ORGANIZATION_PASSWORD: "<redacted>"

rmt_db_username: rmt
rmt_db_password: "my-password"

$ ansible-playbook site.yml
# failed with the error above

$ cat /etc/rmt.conf
database:
  host: localhost
  database: rmt
  username: rmt
  password:
  adapter: mysql2
  encoding: utf8
  timeout: 5000
  pool: 5

scc:
  username:
  password:
  sync_systems: true
  metrics:
    enabled: false
    job_name: rmt
...
```

This looks to me that the variables are shadowed/read when creating the `/etc/rmt.conf` file

ansible/site.yml
```diff
@@ -0,0 +1,39 @@
...
+  become: true
+  gather_facts: true
+
+  pre_tasks:
+    - name: Validate required variables
+      ansible.builtin.assert:
+        that:
+          - rmt is defined
+          - rmt.RMT_ORGANIZATION_USERNAME is defined
+          - rmt.RMT_ORGANIZATION_PASSWORD is defined
```
felixsch, line 14: This always succeeds since: https://github.com/SUSE/rmt/pull/1470/changes#diff-b7ac6e16887ee45611114748e38dc438ad43b00269e4057dfc80e397b320a67bR7 already defines it by default.

ansible/roles/rmt/tasks/main.yml
ngetahun: I pushed changes to fix this.

felixsch (approved):
lgtm 👍 

```
localhost:/usr/share/ansible/rmt # rmt-cli sync
I, [2026-06-10T16:40:38.135001 #4723]  INFO -- : Downloading data from SCC
I, [2026-06-10T16:40:38.135168 #4723]  INFO -- : Loading product data from SCC
I, [2026-06-10T16:40:43.011533 #4723]  INFO -- : Updating products
I, [2026-06-10T16:41:50.950962 #4723]  INFO -- : Loading repository data from SCC
I, [2026-06-10T16:41:55.757002 #4723]  INFO -- : Updating repositories
I, [2026-06-10T16:42:02.953431 #4723]  INFO -- : Removing obsolete repositories
I, [2026-06-10T16:42:03.087663 #4723]  INFO -- : Loading subscription data from SCC
I, [2026-06-10T16:42:03.173717 #4723]  INFO -- : Updating subscriptions
localhost:/usr/share/ansible/rmt # 
```

**How I tested this merge request:**

- Checked the ansible playbook to make sense and if something is missing
- Built the rpm to see it working when building the sub package
- Run the ansible playbook on a SLES 16.0 to see it working correctly
- Validated the installation by syncing products with `rmt-cli sync`

nicely done @ngetahun!
cheers,

ansible/roles/rmt/tasks/main.yml
felixsch: fixed and is working now!

ansible/site.yml
```diff
@@ -0,0 +1,39 @@
...
+          - CA cert: /etc/rmt/ssl/rmt-ca.crt
+
+          Next steps:
+          1. Sync products: sudo rmt-cli sync
+          2. Enable products: sudo rmt-cli products enable <product_id>
+          3. Mirror repositories: sudo rmt-cli mirror
+
+          Access the RMT web interface at:
+          - HTTP: http://{{ ansible_fqdn }}
+          - HTTPS: https://{{ ansible_fqdn }} (self-signed certificate)
```
felixsch: just nitpick, ignore if needed: This does not print the new lines when done:

```
TASK [Display post-installation instructions] **************************************************
ok: [localhost] => {
    "msg": "RMT server installation complete!\n\nSSL certificates generated at: /etc/rmt/ssl/\n- Server cert: /etc/rmt/ssl/rmt-server.crt\n- Server key: /etc/rmt/ssl/rmt-server.key\n- CA cert: /etc/rmt/ssl/rmt-ca.crt\n\nNext steps:\n1. Sync products: sudo rmt-cli sync\n2. Enable products: sudo rmt-cli products enable <product_id>\n3. Mirror repositories: sudo rmt-cli mirror\n\nAccess the RMT web interface at:\n- HTTP: http://localhost.localdomain\n- HTTPS: https://localhost.localdomain (self-signed certificate)\n"
```

maybe there is an option to print the newlines

PR #1473, opened 2026-05-27
Fix race condition

jesusbv:
## Description

there are 3 siblings per region in the update infrastructure

each server with its own DB

there is an scenario where a system registered to the update infrastructure
runs the clean command, that command should remove the system from SCC
and from that sibling database and also trigger a regsharing event so the
system is also cleaned up in each siblings' DB

it could happen that while the registration sharing is being
triggered, another registration for the same system happens, and the
region server could return a different sibling than the one where the
system was removed from the database and not processed yet the
regsharing event to remove the system from its own DB, raising an exception

the way the client is implemented, the client would try a different
sibling, eventually producing a duplicate on SCC:

  Could not create offline certificate: Validation failed:
  System token This combination of system login, password and token is already taken

This Fixes bsc#1264275

* Related Issue / Ticket / Trello card: [bsc#1264275](https://bugzilla.suse.com/show_bug.cgi?id=1264275)

## How to test 

- register a BYOS system `registercloudguest -r FOO`
- de register the system `registercloudguest --clean`
- re-register the system immediately after `registercloudguest -r FOO` (no time to regsharing to catch up)

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -382,7 +382,7 @@ module SccProxy
             # retry to allow that regsharing info update this update infrastructure server
             message = "System could not be created with the following values #{system_values}: #{e.message}"
             if attempts < 3
-              logger.info "#{message}\nRetrying in 10 seconds"
+              logger.info "#{message}. Attempt #{attempts} of 3\nRetrying in 10 seconds"
               sleep 10
               retry
             else
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -96,15 +96,25 @@ module SccProxy
       scc_request
     end
 
-    def announce_system_scc(auth, params, system_token)
+    def announce_system_scc(auth, params, system_token, logger)
       uri = URI.parse(ANNOUNCE_URL)
       http = Net::HTTP.new(uri.host, uri.port)
       http.use_ssl = true
       scc_request = prepare_scc_announce_request(uri.path, auth, params, system_token)
       response = http.request(scc_request)
-      response.error! unless response.code_type == Net::HTTPCreated
+      begin
+        response_body = JSON.parse(response.body)
+      rescue JSON::ParserError => e
+        logger.info "Could not parse SCC response body: #{e.message}"
+        response_body = {}
+      end
+      return [response_body, response.to_hash] if response.code_type == Net::HTTPCreated
 
-      [JSON.parse(response.body), response.to_hash]
+      # if the system is in SCC it means it was created by a sibling
+      error = response_body['localized_error'] || response_body['error'] || ''
+      return [nil, nil] if error.include? 'system login, password and token is already taken'
+
+      response.error!
     end
 
     def scc_activate_product(system, product, auth, params, mode)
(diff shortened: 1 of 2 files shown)


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -311,20 +324,26 @@ def announce_system
...
+            if scc_response.present?
+              system_values[:scc_system_id] = scc_response['id']
+              system_values[:password] = scc_response['password']
+              profile_response_header_handling(scc_response_headers)
+            else
+              system_values[:password] = instance_identifier
+            end
+          end
+          @system, error = create_system(system_values, logger)
+          if @system.presence
```
jcomello: `Object#presence` is used in attributions operations with `||`. It works in this case, because `nil` is considered a `false` value in ruby, but I think since it's a `if` statement the best thing is to use `Object.present?`

```suggestion
          if @system.present?
```

jcomello (approved):

jcomello (approved):

rjschwei (approved):

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -96,15 +96,25 @@ def prepare_scc_upgrade_request(uri_path, product, auth, system_token, mode)
...
+        response_body = JSON.parse(response.body)
+      rescue JSON::ParserError => e
+        logger.info "Could not parse SCC response body: #{e.message}"
+        response_body = {}
+      end
+      return [response_body, response.to_hash] if response.code_type == Net::HTTPCreated
+
+      # if the system is in SCC it means it was created by a sibling
+      error = response_body['localized_error'] || response_body['error'] || ''
+      return [nil, nil] if error.include? 'system login, password and token is already taken'
```
digitaltom: This will break as soon as that message changes. Is there a specific HTTP code in that scenario? Maybe shorten it to error.include? 'already taken'. 
It would also be good to add a logger.warn() message here, to see in the log this happened.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -96,15 +96,25 @@ def prepare_scc_upgrade_request(uri_path, product, auth, system_token, mode)
...
-      response.error! unless response.code_type == Net::HTTPCreated
+      begin
+        response_body = JSON.parse(response.body)
+      rescue JSON::ParserError => e
+        logger.info "Could not parse SCC response body: #{e.message}"
+        response_body = {}
+      end
+      return [response_body, response.to_hash] if response.code_type == Net::HTTPCreated
+
+      # if the system is in SCC it means it was created by a sibling
```
digitaltom, line 113: Is there a bug# that describes the situation? Could be helpful to link it so that understanding the case in the future is easier.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -311,20 +324,26 @@ def announce_system
...
             )
-            profile_response_header_handling(scc_response_headers)
+            system_values[:proxy_byos_mode] = :byos
+            system_values[:proxy_byos] = true
+            if scc_response.present?
+              system_values[:scc_system_id] = scc_response['id']
+              system_values[:password] = scc_response['password']
+              profile_response_header_handling(scc_response_headers)
+            else
+              system_values[:password] = instance_identifier
```
digitaltom, line 337: Why is only the password set here?

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -346,9 +365,33 @@ def announce_system
...
+          begin
+            attempts += 1
+            [System.create!(system_values), nil]
+          rescue ActiveRecord::RecordInvalid => e
+            # if the system is also in this DB
+            # means that it is possible that the regsharing has not caught up to this sibling yet
+            # retry to allow that regsharing info update this update infrastructure server
+            message = "System could not be created with the following values #{system_values}: #{e.message}"
+            if attempts < 3
+              logger.info "#{message}\nRetrying in 10 seconds"
```
digitaltom: Please add the attempt # to the log.

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -346,9 +365,33 @@ def announce_system
...
+              sleep 10
```
digitaltom, line 386: Ho long does it take for system changes to be synced across siblings?

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -346,9 +365,33 @@ def announce_system
...
 
+        def create_system(system_values, logger)
+          attempts = 0
+          begin
+            attempts += 1
+            [System.create!(system_values), nil]
+          rescue ActiveRecord::RecordInvalid => e
+            # if the system is also in this DB
+            # means that it is possible that the regsharing has not caught up to this sibling yet
+            # retry to allow that regsharing info update this update infrastructure server
```
digitaltom:
```suggestion
            # The to-be-announced system is already in the database. 
            # This could mean it got de-registered on another update infrastructure server a short time ago. 
            # Let's wait 3*10 seconds for regsharing to catch up
```

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: done

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: well, SCC only returns `422`...so not possible to distinguish using the HTTP code

will update the include check for the shorter version and add the log warning

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: 30 seconds (max)

engines/scc_proxy/lib/scc_proxy/engine.rb
jesusbv: before this change, we set [all these fields](https://github.com/SUSE/rmt/blob/master/engines/scc_proxy/lib/scc_proxy/engine.rb#L315-L320) to create a `System`, after the fix, if we reach that point, it means another update server did create the `System` with all those fields in the DB, meaning it is similar to a regsharing, in the regsharing we do not have `scc_system_id` which is the missing field from that list, the `proxy_mode` and `proxy_byos` are still set 

IIRC, SCC returns the instance id as the password for `BYOS` so we could refactor that if agreed, but not in this PR

engines/scc_proxy/lib/scc_proxy/engine.rb

engines/scc_proxy/lib/scc_proxy/engine.rb

digitaltom (approved):

PR #1475, opened 2026-05-27
remove use of ActiveSupport::Configurable

jesusbv:
## Description

this change was done in master branch in #1402 PR
but not ported to rmt_3 branch, making a warning

DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.

show up, this warning shall be removed

this change fixes that

## How to test 

start RMT, the warning in the description must not show up

## Change Type

## Checklist

## Review

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -17,7 +17,7 @@ GEM
       erubi (~> 1.11)
       rails-dom-testing (~> 2.2)
       rails-html-sanitizer (~> 1.6)
-    active_model_serializers (0.10.15)
+    active_model_serializers (0.10.16)
       actionpack (>= 4.1)
       activemodel (>= 4.1)
       case_transform (>= 0.2)


jesusbv:
should we include all the gem changes from #1402  ?

jcomello (approved):

jeremy-moffitt:
> should we include all the gem changes from #1402 ?

I'm a little rusty with Ruby, but isnt that pinning the zeitwerk version to the latest supported by Ruby 2, which shouldn't be necessary for the 3.x branch? Looking at the [comparison between 2.6.18 and 2.7.0](https://github.com/fxn/zeitwerk/compare/v2.6.18...v2.7.0)  it looks like 2.7 dropped support for Ruby versions older than 3.2, which required the pin for RMT 2. So I don't *think* we need to pull those changes in, but I'm not positive.

jesusbv:
> I'm a little rusty with Ruby, but isnt that pinning the zeitwerk version to the latest supported by Ruby 2, which shouldn't be necessary for the 3.x branch? Looking at the [comparison between 2.6.18 and 2.7.0](https://github.com/fxn/zeitwerk/compare/v2.6.18...v2.7.0) it looks like 2.7 dropped support for Ruby versions older than 3.2, which required the pin for RMT 2. So I don't _think_ we need to pull those changes in, but I'm not positive.

yes, but there is also `rack` mentioned in that PR however, I do not see a `rack` gem change in the PR changes ? 
checking  the [Gemfile.lock](https://github.com/SUSE/rmt/blob/rmt_3/Gemfile.lock#L211) of `rmt_3` seems the gem version is higher than the change in said PR for RMT 2.X

I think we can merge this one

digitaltom:
> So I don't think we need to pull those changes in, but I'm not positive.

You are right. RMT2 has many gem dependencies pinned to versions that are compatible with Ruby 2.5.9. 
RMT3 does not have that constraint, and because of that uses the latest versions of those dependencies.

digitaltom (approved):

PR #1483, opened 2026-06-03
Fix purge

jesusbv:
## Description

when purging a large amount of systems, the command crashes,
these changes aim to remove the bottle-necks and to provide
more information on the process and a recovery path
    
add verbose option

* Related Issue / Ticket / Trello card: [bsc#1262706](https://bugzilla.suse.com/show_bug.cgi?id=1262706)

## How to test 

running purge in a database with ~30 million systems, the command finishes without any issue
with or without verbose option

## Change Type

## Checklist

## Review


lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,94 @@ def remove(target)
...
   protected
 
+  def delete_all_matching_systems(before)
+    n_systems_destroyed = 0
+    attempts = 0
+    begin
+      attempts += 1
+      System.where('last_seen_at < ?', before).in_batches(of: DELETE_BATCH_SIZE) do |batch|
+        n_systems_destroyed += batch.length
+        # pp batch.class
```
rtamalin: Is this a debugging artifact that was left in?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,94 @@ def remove(target)
...
+        batch.destroy_all
+        puts "#{n_systems_destroyed} systems destroyed"
+      end
+      n_systems_destroyed
+    rescue StandardError => e
+      # pp 'aaa'
```
rtamalin: Is this a debugging artifact that was left in?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,94 @@ def remove(target)
...
+    [systems_matched, false]
+  end
+
+  def delete_systems(systems)
+    attempts = 0
+    deleted_systems = []
+    begin
+      attempts += 1
+      systems.each_slice(DELETE_BATCH_SIZE) do |sliced_systems_ids|
+        System.where(id: sliced_systems_ids).in_batches(of: DELETE_BATCH_SIZE).destroy_all
```
rtamalin: It seems strange to use the same batch size here as for the systems.each_slice()... Did you mean to use VERBOSE_BATCH_SIZE?

lib/rmt/cli/systems.rb
jesusbv: yes, forcing the new test to fail

lib/rmt/cli/systems.rb
jesusbv: yes, forcing the new test to fail

lib/rmt/cli/systems.rb
jesusbv: no, the intention is to have a slice of the total number of systems and destroy them so `DELETE_BATCH_SIZE` is the right value here

rtamalin: Ok, just seemed a bit strange to double up on using the same size for both the slicing and the batching here.

rtamalin:
@jesusbv It looks good to me but it may best to get a Ruby/Rails expert to give a blessing

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,102 @@ def remove(target)
     $ rmt-cli systems purge --no-confirmation --before 2022-02-28
   PURGE
   def purge
-    ask, before = purge_options
-    systems = System.where('last_seen_at < ?', before).pluck(:id)
+    ask, before, verbose = purge_options
+    systems, fetch_ok = verbose ? get_all_matches(before) : [[], true]
```
digitaltom: I read this line as: When verbose is false, `systems` is set to `[]`. Is that right?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,102 @@ def remove(target)
...
 
-    if systems.empty?
-      warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.")
-      return
+    if verbose
```
digitaltom: Shouldn't we print at least the count of to be deleted systems in the non-verbose case?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,102 @@ def remove(target)
...
   protected
 
+  def handle_empty_systems(before, fetch_ok)
+    return unless fetch_ok
+
+    warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.")
+  end
+
+  def delete_systems(systems, before, verbose, fetch_ok)
+    n_deleted, all = verbose ? delete_systems_by_id(systems) : delete_all_matching_systems(before)
```
digitaltom: Why do we delete by different criteria depending on the `verbose` flag?

lib/rmt/cli/systems.rb
jesusbv: we do not know how many systems are going to be deleted in the non-verbose case

lib/rmt/cli/systems.rb
jesusbv: if the verbose flag is set, we need to print _all_ the matching systems,
then we need to delete them

if the verbose flag is not set, we can delete the matching systems directly

different operations

we could use the same deletion method for both, but we did the effort of finding the matching systems if verbose flag is set, I'd prefer not to do that twice

lib/rmt/cli/systems.rb
jesusbv: yes, it can be changed to
```ruby
systems, fetch_ok = get_all_matches(before) if verbose
```

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,102 @@ def remove(target)
...
+    attempts = 0
+    begin
+      attempts += 1
+      System.where('last_seen_at < ?', before).in_batches(of: DELETE_BATCH_SIZE) do |batch|
+        n_systems_destroyed += batch.length
+        batch.destroy_all
+        puts "#{n_systems_destroyed} systems destroyed"
+      end
+      [n_systems_destroyed, true]
+    rescue StandardError => e
```
digitaltom, line 133: In which kind of errors are you running into here on the production RMTs?

jesusbv: currently, time out errors but we shall not show a traceback if another error happens, right ?

lib/rmt/cli/systems.rb
digitaltom: But this pr changes the default to verbose=false, so this is what it will look like by default for the user: 

```
> rmt-cli systems purge
Do you want to delete all the matching systems? (y/n)
```

That's not a good UX, because the use has no idea what is going to happen. I suggest: 
* Keep the default with verbose=true
* Extend the output of the verbose=false case with the amount of systems, and the clear_before date. If counting the affected is not possible, let's add a comment to the code why it's skipped.

lib/rmt/cli/systems.rb
digitaltom: I think the "Error while purging systems" log below should include also the exception class, because `e.message` might not be meaningful in each case.

jesusbv: done

lib/rmt/cli/systems.rb
jesusbv: done, `quiet` instead of `verbose`

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
     $ rmt-cli systems purge --no-confirmation --before 2022-02-28
   PURGE
   def purge
-    ask, before = purge_options
-    systems = System.where('last_seen_at < ?', before).pluck(:id)
+    ask, before, quiet = purge_options
+    systems, fetch_ok = get_all_matches(before) unless quiet
```
digitaltom: Suggest to call this `system_ids` instead of `systems` for clarity.

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+    delete_systems(systems, before, quiet, fetch_ok)
   end
 
   protected
 
+  def handle_no_matching_systems(before, fetch_ok)
+    warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.") if fetch_ok
+  end
+
+  def delete_systems(systems, before, quiet, fetch_ok)
```
digitaltom: I think the `quiet` parameter is not necessary here, because the condition for which method to use for deleting the systems actually can depend on the presence of `systems`?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+    n_deleted, all = quiet ? delete_all_matching_systems(before) : delete_systems_by_id(systems)
+    return handle_no_matching_systems(before, true) if n_deleted.zero?
+
+    status = all ? 'all' : 'some'
+    puts "Purged #{status} systems that have not contacted this RMT since #{before}."
+    puts "Systems that have not contacted this RMT since #{before} may still be in this RMT" unless fetch_ok || all
+  end
+
+  def delete_all_matching_systems(before)
```
digitaltom: Suggest to call this `delete_inactive_systems`

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+      end
+      [deleted_systems.length, true]
+    rescue StandardError => e
+      systems -= deleted_systems
+      if attempts < 3
+        puts "Error while purging systems: #{e.class} #{e.message}. Attempt #{attempts}/3, retrying in 5 seconds"
+        sleep 5
+        retry
+      else
+        n_deleted = systems.length - deleted_systems.length
```
digitaltom: This calculation looks wrong, because it seems to set the n_deleted (number of deleted systems) to the number of still existing systems?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+        puts "Error while purging the systems: #{e.class} #{e.message}, " \
+          "all #{systems.length} systems could not be removed, #{n_deleted} systems still in the database"
```
digitaltom:
```suggestion
          "all #{systems.length} systems could not be removed, #{systems.length} systems still in the database"
```

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+
+  def delete_systems_by_id(systems)
+    attempts = 0
+    deleted_systems = []
+    begin
+      attempts += 1
+      systems.each_slice(DELETE_BATCH_SIZE) do |sliced_systems_ids|
+        System.where(id: sliced_systems_ids).in_batches(of: DELETE_BATCH_SIZE).destroy_all
+        deleted_systems += sliced_systems_ids
+        n_deleted = (systems.length - deleted_systems.length > 0) ? (systems - deleted_systems).length : systems.length
```
digitaltom: This is inconsistent: 
In the first iterations, this variable will carry the number of systems that are left to be deleted, in the last iteration it will carry the full amount of systems. And will print an incorrect output in the next line.

digitaltom: Why not just use `deleted_systems.length`?

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+    attempts = 0
+    begin
+      attempts += 1
+      # instead of getting all the systems to be shown
+      # before deletion, where we would know
+      # how many systems are going to be deleted
+      # query systems directly, thus we do not know how
+      # many systems before running the deletion
+      System.where('last_seen_at < ?', before).in_batches(of: DELETE_BATCH_SIZE) do |batch|
+        n_systems_destroyed += batch.length
```
digitaltom: The whole exception handling below is there to catch timeouts in the middle of the `batch.destroy_all` call. So setting n_systems_destroyed before that will be a wrong value in case of an exception. Additionally, the .length already triggers an sql load before the delete. 
Suggest: `n_systems_destroyed += batch.destroy_all.length`

lib/rmt/cli/systems.rb
digitaltom: After creating a slice of `DELETE_BATCH_SIZE` size, and loading those systems, I don't see what the `in_batches(of: DELETE_BATCH_SIZE)` would do, because the amount of queried systems always is <= `DELETE_BATCH_SIZE`.

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
-      warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.")
-      return
+    unless quiet
+      return handle_no_matching_systems(before, fetch_ok) if systems.empty?
+
+      print_rows(systems)
     end
-    print_rows(systems)
-    return if ask && !yesno(_('Do you want to delete these systems?'))
+    return if ask && !yesno(_('Do you want to delete all the matching systems?'))
```
digitaltom: We need to give the user at least some idea of which systems are going to get deleted.
If it's not possible to count the systems, we need a comment in the code explaining this. We could at least show the date that is used to detect to be purged systems.

lib/rmt/cli/systems.rb
```diff
@@ -73,6 +74,7 @@ def remove(target)
   desc 'purge', _('Removes inactive systems')
   option :before, aliases: '-b', type: :string, desc: _('Remove systems before the given date (format: "<year>-<month>-<day>")')
   option :confirmation, type: :boolean, default: true, desc: _('Ask for confirmation or do not ask for confirmation and require no user interaction')
+  option :quiet, aliases: '-q', type: :boolean, default: false, desc: _('do not show system details')
```
digitaltom:
```suggestion
  option :quiet, aliases: '-q', type: :boolean, default: false, desc: _('Do not show system details before deletion.')
```

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+  def handle_no_matching_systems(before, fetch_ok)
+    warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.") if fetch_ok
+  end
+
+  def delete_systems(systems, before, quiet, fetch_ok)
+    n_deleted, all = quiet ? delete_all_matching_systems(before) : delete_systems_by_id(systems)
+    return handle_no_matching_systems(before, true) if n_deleted.zero?
+
+    status = all ? 'all' : 'some'
+    puts "Purged #{status} systems that have not contacted this RMT since #{before}."
```
digitaltom: Missing i18n method `_()` on all many introduced user facing strings in this pr.

lib/rmt/cli/systems.rb
digitaltom: I'd suggest also a code comment explaining why we are doing this catch-retry handling here, because it's unexpected for 'normal' code.

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+        retry
+      end
+      puts "Could not delete all systems last seen before #{before}: #{e.message}"
+      [n_systems_destroyed, false]
+    end
+  end
+
+  def get_all_matches(before)
+    systems_matched = []
+    System.where('last_seen_at < ?', before).in_batches(of: DELETE_BATCH_SIZE, order: :desc) do |batch|
```
digitaltom, line 151: `order: :desc` is unnecessary db overhead in this case

lib/rmt/cli/systems.rb
jesusbv: that means to express the remaining systems, I updated the name for clarity

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,106 @@ def remove(target)
...
+  end
+
+  def delete_systems_by_id(systems)
+    attempts = 0
+    deleted_systems = []
+    begin
+      attempts += 1
+      systems.each_slice(DELETE_BATCH_SIZE) do |sliced_systems_ids|
+        System.where(id: sliced_systems_ids).in_batches(of: DELETE_BATCH_SIZE).destroy_all
+        deleted_systems += sliced_systems_ids
```
digitaltom: Why do we collect the ids of deleted systems here? When deleting 30M systems, this will be a giant array. Probably we're only interested in the count.

lib/rmt/cli/systems.rb
jesusbv: updated in [d677527](https://github.com/SUSE/rmt/pull/1483/commits/d67752753c8ee1451ea1ed703f9bdc11cd055785)

lib/rmt/cli/systems.rb
jesusbv: because I would like to show to the user how many systems are still to be deleted

digitaltom:
Please don't forget to update MANUAL.md and the RPM changelog

lib/rmt/cli/systems.rb
jesusbv: I added a comment in the code already ?

OK, will update the message to the user

lib/rmt/cli/systems.rb
jesusbv: it depends, the original  behaviour of showing the systems was to show them in descending order
from newest to oldest, if we want to keep that behaviour, we need to sort them at some point

jesusbv: maybe it's preferable to sort the array of ids once all are fetched

Thoughts ?

lib/rmt/cli/systems.rb
jesusbv: I will update the old (like [this](https://github.com/SUSE/rmt/blob/master/lib/rmt/cli/systems.rb#L99)) and the new added ones

rjschwei:
Appears to overlap with #1484 should this PR be closed?

jesusbv:
> Appears to overlap with #1484 should this PR be closed?

#1484 is to have the fix in `rmt_3`

lib/rmt/cli/systems.rb
jesusbv: maybe we do not care about the order when showing the systems ?

rjschwei (approved):

lib/rmt/cli/systems.rb
jesusbv: I understood that the value of the constant was wrong, instead of redundant `in_batches`

lib/rmt/cli/systems.rb
digitaltom: For the case where we show the systems, it makes sense to sort. For the quiet mode, no need to order.

lib/rmt/cli/systems.rb
```diff
@@ -85,28 +87,104 @@ def remove(target)
...
+  def handle_no_matching_systems(before, fetch_ok)
+    warn _("No systems to be purged on this RMT instance. All systems have contacted RMT after #{before}.") if fetch_ok
+  end
+
+  def delete_systems(system_ids, before, fetch_ok)
+    n_deleted, all = system_ids.blank? ? delete_inactive_systems(before) : delete_systems_by_id(system_ids)
+    return handle_no_matching_systems(before, true) if n_deleted.zero? && all
+
+    status = all ? 'all' : 'some'
+    puts _("Purged #{status} systems that have not contacted this RMT since #{before}.") unless n_deleted.zero?
```
digitaltom: Adding parameters into i18n strings doesn't work like this, they need to get marked in the string and get provided as an extra parameter, see eg: https://github.com/SUSE/rmt/blob/89a33f7173d705ae788b56549746d5636c270e6c/app/controllers/application_controller.rb#L90

digitaltom: Issue applies to many _() calls in this branch

lib/rmt/cli/systems.rb
```diff
@@ -73,6 +74,7 @@ def remove(target)
   desc 'purge', _('Removes inactive systems')
   option :before, aliases: '-b', type: :string, desc: _('Remove systems before the given date (format: "<year>-<month>-<day>")')
   option :confirmation, type: :boolean, default: true, desc: _('Ask for confirmation or do not ask for confirmation and require no user interaction')
+  option :quiet, aliases: '-q', type: :boolean, default: false, desc: _('do not show system details before deletion')
```
digitaltom:
```suggestion
  option :quiet, aliases: '-q', type: :boolean, default: false, desc: _('Do not show system details before deletion')
```

digitaltom (approved):

PR #1492, opened 2026-06-18
fix race condition for PAYG

jesusbv:
This Fixes bsc#1268149

## Description

there is an scenario when a PAYG system gets registered with sibling A, then the systems tries to access sibling B immediately after that registration and the registration sharing has not propagated the needed values, producing 'Invalid system credentials' error on the system

* Related Issue / Ticket / Trello card: [bsc#1268149](https://bugzilla.suse.com/show_bug.cgi?id=1268149)

## How to test 

Run the same tests that produced the error consistently and the error should not happen 

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -603,7 +603,7 @@ module SccProxy
         # as BYOS in the cloud does not use SYSTEM_TOKEN headers
         def authenticate_system(skip_on_duplicated: false)
           authenticate_or_request_with_http_basic('RMT API') do |login, password|
-            @systems = System.get_by_credentials(login, password)
+            fetch_systems_from_db(login, password)
             if @systems.present?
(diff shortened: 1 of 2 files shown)


engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -627,6 +627,21 @@ def authenticate_system(skip_on_duplicated: false)
...
         end
 
+        def fetch_systems_from_db(login, password)
+          attempts = 0
+          begin
+            attempts += 1
+            @systems = System.get_by_credentials(login, password)
+            raise ActionController::TranslatedError.new(N_('No systems found')) if @systems.blank?
+          rescue StandardError
+            if attempts < 5
```
rjschwei, line 636: I think we should be more clever here. We should only run this loop if siblings are configured. In a case where there is only 1 single system set up and no regsharing is configured waiting doesn't really help.

PR #1511, opened 2026-07-07
Synchronise V2.x stream historical changes to V3 stream

rtamalin:
This is intended to avoid historical discrepancies in the .changes file content for future V3.x stream submissions.

Relates: SCC-848


ngetahun (approved):
LGTM

PR #1480, opened 2026-05-28
Introduce RMT Release Management Skill and GEMINI.md

ngetahun:
## Description

This PR introduces a new agent skill `rmt-release-management` and a `GEMINI.md` file to streamline the release process and provide architectural context for the RMT project.

### Key Changes
- **GEMINI.md**: Added to provide a high-level architectural overview, tech stack details, and development workflows.
- **rmt-release-management Skill**:
    - `SKILL.md`: Core instructions for the release process, including OBS/IBS workflows and GitHub tagging.
    - `release-workflow.md`: Detailed 5-phase lifecycle documentation with a visual Mermaid graph.
    - `release.sh`: An idempotent bash script to automate version updates, source tarball generation, and OBS/IBS submissions.

* Related Issue / Ticket / Trello card: N/A

## How to test 

1. Review `GEMINI.md` for accuracy of RMT architecture and workflows.
2. Review `.agents/skills/rmt-release-management/SKILL.md` and its references for correctness of the release process.
3. (Optional) Run `.agents/skills/rmt-release-management/scripts/release.sh --version 0.0.0 --obs-path /tmp --dry-run` to verify the script logic.

## Change Type

- [ ] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [x] **Documentation Update** (a change which only updates documentation)

## Checklist

- [x] I have reviewed my own code and believe that it's ready for an external review.
- [x] I have provided comments for any hard-to-understand code.
- [ ] I have documented the `MANUAL.md` file with any changes to the user experience.
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

PR #1512, opened 2026-07-07
Chronologically re-order recent rmt-server.changes entries

rtamalin:
Relates: SCC-848

### DIFF
--- a/package/obs/rmt-server.changes
+++ b/package/obs/rmt-server.changes
@@ -1,4 +1,16 @@
 -------------------------------------------------------------------
+Fri Jun  5 16:15:34 UTC 2026 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
+
+- Version 2.28
+  * Set arch to unknown for Rancher based products (jsc#SCC-759)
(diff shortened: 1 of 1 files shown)


rtamalin:
Note that this re-ordering gets part the SLE 15 SP7 build failure due to the entries not being in chronological order...

But also means that the v2.28 entru is the most recent one...

jeremy-moffitt (commented):
If the corresponding changes are on the rmt_3 stream, then they should be in the changelog, I wonder if this entry should be versionless though since 2.28 is not going to publish from the rmt_3 branch

jeremy-moffitt (approved):
slightly skeptical about the versioning, but other projects have changes files that span multiple streams so I think this is probably ok in retrospect

rtamalin:
I have proposed an alternate approach of updating the timestamps of the V3.x stream entries to occur after the v2.28 entry in #1513 which also works to resolve the SLE 15 SP7 build failure.

rtamalin:
Equivalent changes were pulled into #1480

PR #1513, opened 2026-07-07
Update timestamps of recent rmt-server.changes v3 entries

rtamalin:
Update the timestamps to come chronologically after the v2.28 entry.

Relates: SCC-848


rtamalin:
This is an alternate approach from that proposed in #1512 to address the issue of the rmt-server.changes entries not being in chronological ordering, leading to SLE 15 SP7 build failures.

jeremy-moffitt (changes requested):
I prefer the other approach to this. Artificially changing the dates makes it hard to correlate changes with the implementation of those changes.

rtamalin:
Closed in favour of #1512

PR #1494, opened 2026-06-18
change data column type in profile table

jesusbv:
## Description
currently the data column type is TEXT

TEXT 65535 max length in bytes

and this changes the type to MEDIUMTEXT

MEDIUMTEXT 16777215 max length in bytes

which it could be less depending on the encoding

* Related Issue / Ticket / Trello card: [bsc#1268305](https://bugzilla.suse.com/show_bug.cgi?id=1268305)

## How to test 

Apply the change/migration, then inspect the table, that column should have the expected type (MEDIUMTEXT)

## Change Type

*Please select the correct option.*

- [X] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

## Checklist

*Please check off each item if the requirement is met.*

- [X] I have reviewed my own code and believe that it's ready for an external review.
- [X] I have provided comments for any hard-to-understand code.
- [X] I have documented the `MANUAL.md` file with any changes to the user experience.
- [X] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes`.

## Review

Please check out our [review guidelines](https://github.com/SUSE/scc-docs/blob/master/team/workflow/code_review.md) 
and get in touch with the author to get a shared understanding of the change.

PR #1522, opened 2026-07-28
Avoid registration and update request handling failure due to profiles handling errors

rtamalin:
## Description

Rework the announce_system request handling workflow to create the system record first, without including any profiles, and then, if profiles are provided, update the system record with any provided complete profiles. As before, if any issues are identified while processing the profiles the response header will be set. Additionally any failures that occur while updating the system record will be logged, but request processing will continue as normal.

As part of update request handling log any errors that occur during the system profiles updates, set the response header indicating that clients should send complete profiles next time, and continue with normal processing.

Refactored the update request handler to split out the processing of system profiles to a separate routine to address rubocop complexity concerns.

Updated spec tests for both announce_system and update requests to exercise the modified processing scenarios.

Created placeholder v3.2 entry in the rmt-server.changes to hold a changelog entry for these changes.

AI-Assistance-By: Qwen 3.6 35B MoE (A3B)

Fixes: SCC-822

## How to test 

To assist in testing this scenario I've SUSE/connect-ng's `public-api-demo` to add support for specifying profiles via a PROFILES_DIR env var. See [public-api-demo/README.md](https://github.com/SUSE/connect-ng/blob/main/cmd/public-api-demo/README.md) for more details.

However, there seems to be a limitation in the `public-api-demo` command, or possibly the SUSE/connect-ng client request setup where profiles larger than approx 1MiB in size cause an error when creating the client HTTP request. Therefore test this I had to temporarily reduce the max size of the data field in my local RMT's profiles table to a smaller value to allow me to submit profiles that were too large.

## Change Type

## Checklist

## Review

### DIFF
--- a/app/controllers/api/connect/v3/subscriptions/systems_controller.rb
+++ b/app/controllers/api/connect/v3/subscriptions/systems_controller.rb
@@ -1,16 +1,16 @@
 class Api::Connect::V3::Subscriptions::SystemsController < Api::Connect::BaseController
 
   def announce_system
-    # Construct the system creation parameters
-    create_params = {
-      hostname: params[:hostname],
-      system_information: info_params(:hwinfo)[:hwinfo].to_json
-    }
+    @system = System.create!(hostname: params[:hostname], system_information: info_params(:hwinfo)[:hwinfo].to_json)
 
-    # Check if any profiles have been provided
-    process_system_profiles(create_params)
+    # Extract any complete profiles, setting the profile response header
+    # if invalid or incomplete profiles are detected.
+    profile_params = {}
+    extract_complete_profiles(profile_params)
 
-    @system = create_system_retry_if_profiles_provided(create_params)
+    # Update with any provided profiles, just setting the response header
+    # if and logging a message if any errors occur.
+    update_system_profiles(profile_params)
 
     logger.info("System '#{@system.hostname}' announced")
     respond_with(@system, serializer: ::V3::SystemSerializer, location: nil)
@@ -18,24 +18,19 @@ class Api::Connect::V3::Subscriptions::SystemsController < Api::Connect::BaseCon
 
   private
 
-  def create_system_retry_if_profiles_provided(params)
-    System.create!(**params)
+  def update_system_profiles(profile_params)
+    # do nothing if profile_params is empty
+    return unless profile_params.any?
+
+    @system.update!(**profile_params)
   rescue StandardError => e
-    # Only retry if profiles were present or haven't already been stripped;
-    # prevents infinite retry loops for errors not related to profiles.
-    # In the event of a retry set the response header telling the client
-    # to send full profiles next time.
-    if params[:complete_profiles]
-      logger.warn("System creation failed with profiles, retrying without: #{e.message}")
-      params.delete(:complete_profiles)
-      response.headers['X-System-Profiles-Action'] = 'clear-cache'
-      System.create!(**params)
-    else
-      raise
-    end
+    # If any errors occur while updating the profiles, log a warning
+    # and set the response header but do not fail the request.
+    logger.warn("System '#{@system.hostname}' profile update failed: #{e.message}")
+    response.headers['X-System-Profiles-Action'] = 'clear-cache'
   end
 
-  def process_system_profiles(create_params)
+  def extract_complete_profiles(profile_params)
     if params.key?(:system_profiles)
       profiles = info_params(:system_profiles)[:system_profiles]
 
@@ -52,11 +47,11 @@ class Api::Connect::V3::Subscriptions::SystemsController < Api::Connect::BaseCon
         response.headers['X-System-Profiles-Action'] = 'clear-cache'
       end
 
-      # Include the complete profiles in create_params only if
+      # Include the complete profiles in profile_params only if
       # complete profiles were actually provided
       if complete.any?
         logger.debug("valid complete profiles detected: #{complete.keys}")
-        create_params[:complete_profiles] = complete
+        profile_params[:complete_profiles] = complete
       end
     end
   end
--- a/app/controllers/api/connect/v3/subscriptions/systems_controller.rb
+++ b/app/controllers/api/connect/v3/subscriptions/systems_controller.rb
@@ -10,7 +10,22 @@ class Api::Connect::V3::Subscriptions::SystemsController < Api::Connect::BaseCon
     # Check if any profiles have been provided
     process_system_profiles(create_params)
 
-    @system = System.create!(**create_params)
+    # Retry system creation without profiles if initial attempt fails
+    begin
+      @system = System.create!(**create_params)
+    rescue StandardError => e
+      # Only retry if profiles were present or haven't already been stripped;
+      # prevents infinite retry loops for errors not related to profiles.
+      # In the event of a retry set the response header telling the client
+      # to send full profiles next time.
+      if create_params[:complete_profiles]
+        logger.warn("System creation failed with profiles, retrying without: #{e.message}")
+        create_params.delete(:complete_profiles)
+        response.headers['X-System-Profiles-Action'] = 'clear-cache'
+        retry
+      end
+      raise
+    end
 
     logger.info("System '#{@system.hostname}' announced")
     respond_with(@system, serializer: ::V3::SystemSerializer, location: nil)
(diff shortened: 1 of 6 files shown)


rtamalin:
Created as a draft for now while waiting to see if feature tests pass

rtamalin:
Feature tests passed so draft has been removed for now...

rtamalin:
Have been able to test using enhanced `public-api-demo` command as proposed in https://github.com/SUSE/connect-ng/pull/418 but also needed to temporarily reduce the size of the profile table's data field due to sizing issues encountered when creating client requests with large profile blobs in `public-api-demo`.

jeremy-moffitt:
>  but also needed to temporarily reduce the size of the profile table's data field due to sizing issues encountered when creating client requests with large profile blobs in `public-api-demo`.

reduce from what to what? what makes it temporary?

rtamalin:
> reduce from what to what? what makes it temporary?

When the PCT team identified the original [bsc#1268305](https://bugzilla.suse.com/show_bug.cgi?id=1268305) issue @jesusbv proposed #1494 (corresponding #1506 for rmt_3) to increase the size of the data column to 16MiB from 64KiB.

It is temporary because it was only to be used for testing purposes; I don't want to reduce the profiles table data column size permanently as part of implementing this fix.

jeremy-moffitt:
> > reduce from what to what? what makes it temporary?
> 
> When the PCT team identified the original [bsc#1268305](https://bugzilla.suse.com/show_bug.cgi?id=1268305) issue @jesusbv proposed #1494 (corresponding #1506 for rmt_3) to increase the size of the data column to 16MiB from 64KiB.
> 
> It is temporary because it was only to be used for testing purposes; I don't want to reduce the profiles table data column size permanently as part of implementing this fix.

@gbuenodevsuse do we have good data on the package profile size? might be relevant to this

rtamalin:
> @gbuenodevsuse do we have good data on the package profile size? might be relevant to this
What may be more relevant is that I hit issues with the public-api-demo testing when a profile was approx 1.1MiB in size; that may indicate issues with large profile/payload sizes in SUSE/connect-ng in general, i.e. impacting suseconnect...

rtamalin:
> > @gbuenodevsuse do we have good data on the package profile size? might be relevant to this
> > What may be more relevant is that I hit issues with the public-api-demo testing when a profile was approx 1.1MiB in size; that may indicate issues with large profile/payload sizes in SUSE/connect-ng in general, i.e. impacting suseconnect...

That may possibly be an nginx config issue in the RMT deployment...

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
```diff
@@ -10,7 +10,22 @@ def announce_system
     # Check if any profiles have been provided
     process_system_profiles(create_params)
 
-    @system = System.create!(**create_params)
+    # Retry system creation without profiles if initial attempt fails
```
digitaltom: In pubcloud RMTs, the engine overwrites the announce_system methods with it's own version: https://github.com/SUSE/rmt/blob/master/engines/scc_proxy/lib/scc_proxy/engine.rb#L291 
So the retry here would not cover it. Maybe extract the retry'ed system create to a shared method?

rtamalin:
Switched this back to draft as I need to work on the issue with the PCT hooking that @digitaltom identified.

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
rtamalin: Done

jeremy-moffitt (approved):
looks ok to me, but someone with more Ruby familiarity should approve before merging

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
```diff
@@ -10,14 +10,31 @@ def announce_system
...
-    @system = System.create!(**create_params)
+    @system = create_system_retry_if_profiles_provided(create_params)
 
     logger.info("System '#{@system.hostname}' announced")
     respond_with(@system, serializer: ::V3::SystemSerializer, location: nil)
   end
 
   private
 
+  def create_system_retry_if_profiles_provided(params)
```
mssola: The name of this method should have a bang as it can raise exceptions.

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
```diff
@@ -10,14 +10,31 @@ def announce_system
...
+    System.create!(**params)
```
mssola: I'm honestly not the biggest fan of exceptions, but even with that, I find this to be a big anti-pattern. That is:

1. We call a method that might raise an exception.
2. Then we rescue "StandardError", which just means "any exception of any kind" and proceed with business logic.

To me we should do something like:

1. Does the system have "complete_profiles"? If so delete the params, headers, et al.
2. Call `System.create!` eitherway.

In other words: detect when that specifically can be an issue, react to it, and after all this prep, then just call `System.create!` as usual.

mssola (commented):
LGTM in general, just a question on the workflow on system creation.

mssola:
Small comment out of curiosity. I see that you are using tags on your git commit messages like `AI-Assistance-By`, which is cool. But you are not signing off your commits, which I believe is more important to denote that:

1. The usual meaning for signing off: You are fully responsible for each commit.
2. In conjunction with the AI tag: you have thoroughly reviewed the output of the LLM.

I'm not telling you to suddenly sign off your commits or anything, just curious by the whole thing 😉

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
rtamalin: Can do.

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
rtamalin: The goal is to try with any provided profiles, and if anything goes wrong and profiles were provided, try again without the profiles, as a last ditch effort to get the request to succeed for registrations in this case.

We attempt this because failing registrations can break instance creations for PAYG systems in PubCloud deployments, and one of the original mandates for adding profiles support to the PubCloud RMTs was to avoid breaking registrations if anything went wrong with profiles processing.

I'm open to checking for more finely scoped exceptions if there is a viable small list of high level exception classes, keeping in mind that if we miss something it takes an RMT release cycle and deployment to get a fix deployed to PubCloud or customer RMTs

I'm not sure I get your comment about proceeding with normal business logic in this case for the announce_system request handling.... Yes that is the case for the update request handling, though we do set the response header in both request handlers scenarios if exceptions are encountered processing profiles...

But in this case for announce_system, if the exception (re-)occurs when profiles aren't included, it will be re-raised by the retry routine, and normal Rails handling for the exception will occur, failing the request.

And we already make efforts to sanitize the profiles supplied in the request to construct the complete_profiles parameter, but in the recent case that @jesusbv fixed, the issue was a problem storing the profile data blob in the database, and that general problem scenario is the motivation for this catchall retry check.

Or did you mean something like the following processing workflow:

* extract any profiles from the request an construct an appropriate complete_profiles, which may set the response header if profile issues are detected
* perform the create!() without the complete_profiles parameter included
* update the system with the complete_profiles, catching any exceptions and logging them, and again setting the response header because a profile issue was encountered

rtamalin:
> Small comment out of curiosity. I see that you are using tags on your git commit messages like `AI-Assistance-By`, which is cool. But you are not signing off your commits, which I believe is more important to denote that:
> 
>     1. The usual meaning for signing off: You are fully responsible for each commit.
> 
>     2. In conjunction with the AI tag: you have thoroughly reviewed the output of the LLM.
> 
> 
> I'm not telling you to suddenly sign off your commits or anything, just curious by the whole thing 😉

So in this regard I was following the recommendation to indicate if AI assistance was used in generating the code in the PR. I avoid letting the AI assistant commit or submit anything though, and review and fix things up before I commit and submit... But yeah I can see the benefit of including a signed off by entry to clearly indicate that I have reviewed the code. So a good practice to assume going forward.

rtamalin:
Switching back to draft as my discussion with @mssola identified a viable, likely simpler, alternative approach for the announce_system handling.

app/controllers/api/connect/v3/subscriptions/systems_controller.rb
rtamalin: I've reworked to use the proposed workflow and it seems to simplify things.

rtamalin:
Removed draft status as generally happy with the updated solution, working on testing the PubCloud aspect of this.

mssola:
I see that this is still open. What's the status of it? Do you need another review? Or just more time to finish it? Thanks for your work!

jeremy-moffitt:
@rtamalin  please rebase and advise if this is ready for re-review, thanks!

rtamalin:
The delay has been in completing the verification testing for the modified code in the PubCloud engines which I have been trying to get to this week.

PR #1525, opened 2026-07-30
Heartbeat byos

jesusbv:
## Description

if the system is on SCC and the system gets updated (heartbeat),
the system information is sent to the SCC

* Related Issue / Ticket / Trello card: [bsc#1273155](https://bugzilla.suse.com/show_bug.cgi?id=1273155)

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/scc_proxy/lib/scc_proxy/engine.rb
+++ b/engines/scc_proxy/lib/scc_proxy/engine.rb
@@ -4,7 +4,7 @@ require 'net/http'
 ANNOUNCE_URL = 'https://scc.suse.com/connect/subscriptions/systems'.freeze
 SYSTEM_PRODUCTS_URL = 'https://scc.suse.com/connect/systems/products'.freeze
 SYSTEMS_ACTIVATIONS_URL = 'https://scc.suse.com/connect/systems/activations'.freeze
-DEREGISTER_SYSTEM_URL = 'https://scc.suse.com/connect/systems'.freeze
+SYSTEMS_URL = 'https://scc.suse.com/connect/systems'.freeze
 DEREGISTER_PRODUCT_URL = 'https://scc.suse.com/connect/systems/products'.freeze
 NET_HTTP_ERRORS = [
   Errno::EINVAL,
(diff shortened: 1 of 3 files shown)


jesusbv:
will also add the tests

rtamalin:
Conceptually looks good - will test it tomorrow hopefully

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -585,10 +600,30 @@ def make_system_payg(auth)
       end
 
       Api::Connect::V3::Systems::SystemsController.class_eval do
+        before_action :scc_update_system, only: %i[update], if: -> { (@system.byos? || @system.hybrid?) && request.headers.key?('HTTP_AUTHORIZATION') }
         before_action :scc_deregistration, only: %i[deregister]
 
         protected
 
+        def scc_update_system
+          system_up_params = {
```
digitaltom: Why don't we just take the original `params` here, and add `byos_mode` to it?
Then this code would not need to get changed when the params payload structure changes or gets extended.

jesusbv: yes, I was testing that before and it did work
but that change will be on Monday (day off here)

jesusbv: plus the unit tests

jesusbv:
once approved, I will squash the commits into one

engines/scc_proxy/lib/scc_proxy/engine.rb
```diff
@@ -117,6 +117,21 @@ def announce_system_scc(auth, params, system_token, logger)
...
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      # PUT /connect/systems
+      scc_request = Net::HTTP::Put.new(uri.path, headers(auth, system_token))
+      scc_request.body = params.to_json
+      response = http.request(scc_request)
+      unless response.is_a?(Net::HTTPSuccess)
+        error = JSON.parse(response.body) rescue {}
+        logger.info "Could not update system on SCC, error: (#{response.code}) #{error['error'] || response.message}"
+        response.error!
```
rjschwei, line 131: Does this send an error to the client? 

I am not convinced we want the heartbeat on the client to show a failure. After all heartbeat is supposed to be as invisible as possible on the user side. Sending the client an error makes it visible and potentially triggers an investigation loop. That in turn will raise the question why the user should spend time debugging an issue that is related to data we want and try to collect automatically. If I were the user I'd just turn it off, which is something we do not want.

jesusbv: > Does this send an error to the client?

it does not

> I am not convinced we want the heartbeat on the client to show a failure

agree, but we shall log the heartbeat request errors

jesusbv: @rjschwei ^

rjschwei (approved):

PR #1534, opened 2026-08-08
Update puma dependancy to 7.2.1

rtamalin:
## Description

The depfu #1477 moves forward to 8.0.2, but switching to 8.0.2 comes with a potentially breaking change to address setup/binding.

Updating 7.2.1 should be a non-breaking update to address the same security advisory concerns.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -205,7 +205,7 @@ GEM
       date
       stringio
     public_suffix (7.0.5)
-    puma (7.1.0)
+    puma (7.2.1)
       nio4r (~> 2.0)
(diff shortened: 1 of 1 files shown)


digitaltom (approved):
LGTM. Shall we add the Puma 8 update to one of the next RMT releases?

rtamalin:
> LGTM. Shall we add the Puma 8 update to one of the next RMT releases?

Yeah that was my plan, to do so once we have time to verify/confirm that we won't be negatively impacted by the switch to the 8.x branch.

PR #1542, opened 2026-08-14
Add a VM-based RMT 2.x to 3.x migration test harness

ngetahun:
## Description

Adds a VM-based test harness for the RMT 2.x → 3.x upgrade, plus documentation
for two defects that running it uncovered.

**Why a VM.** Testing this in a container means restoring a 2.x dump into a 3.x
tree and running `db:migrate` as root. That covers the migrations but skips
everything that distinguishes an *upgrade* from a schema change:

* the real RPM transaction, `%pre`/`%post` and the SUSE service macros, going
  from `rmt-server-2.28` to `rmt-server-3.1.0`
* the ruby2.5 → ruby3.4 interpreter switch that comes with it
* `rmt-server-migration.service` as an actual systemd unit — `User=_rmt`,
  `ProtectSystem=full`, `ProtectHome`, `PrivateDevices`
* whether the migration is triggered **automatically** by the upgrade at all,
  which is what an admin running `zypper up` actually experiences

That last point is the reason this exists. A container harness can only tell you
the migrations work when *you* run them; this tells you the upgrade runs them
for you. It does, in under a second, as `_rmt` rather than root.

SLE 15 SP7 is the target because it is the one distribution where both ends of
the upgrade exist: `SLE-Module-Server-Applications15-SP7` ships `rmt-server`
2.28, and `systemsmanagement:SCC:RMT/SLE_15_SP7` ships 3.1.0. The 2.x side is
deliberately the distribution package rather than an OBS rebuild, because that
is what a customer upgrading from the supported RMT actually starts from.

The harness does **not** manage VM lifecycle — it never defines, starts or stops
a domain. You bring a registered SLES 15 SP7 VM up however you normally do and
point the runner at it.

**`docs/upgrading-from-2x.md`** documents what the run found: on a database
migrated from 2.x, `rmt-cli systems list` aborts. These are two independent
defects sharing one trigger — a `systems` row with a `NULL` column that the 2.x
schema allowed and the 3.x migrations correctly preserve:

1. `System#init` calls `Time.zone.now` from an `after_initialize` hook, which
   fires on rows loaded from a `SELECT`, not only on new records. The CLI entry
   points do not boot Rails, so `Time.zone` is `nil` — but `||=` only evaluates
   it when `registered_at` is `NULL`, which is why a fresh install never sees it.
2. `systems_to_arrays` calls `.ljust` on the hostname unguarded, so a `NULL`
   hostname raises in both the table and the CSV renderer.

Neither is caused or cured by the upgrade, so the harness does not assert
`systems list` afterwards — holding the suite red over it would say nothing
about the transition it exists to test. The doc gives a verified backfill
workaround instead.

Neither defect is fixed here — this PR adds the harness and the documentation
only, and touches no runtime code.

* Related Issue / Ticket / Trello card: `<link reference>` — **TODO, no ticket filed yet**

## How to test

Requires a registered SLES 15 SP7 VM with root ssh key auth. The guest needs to
reach `updates.suse.com`; it does *not* need `download.opensuse.org`, since the
runner fetches the 3.x RPMs on the host and copies them in.

```bash
VM=<domain> ci/vm-migration-poc/run.sh      # or: ci/vm-migration-poc/run.sh root@<ip>
```

Re-runs, once the RPMs are cached and the guest has been refreshed once:

```bash
SKIP_FETCH=1 SKIP_REFRESH=1 VM=<domain> ci/vm-migration-poc/run.sh
```

Expected: `26 assertions passed (2.28 -> 3.1.0)`, exit 0, plus one `KNOWN` line
reporting the pre-existing 2.28 defect and four `NOTE` lines about the duplicate
migration timestamps. Full output lands in
`ci/vm-migration-poc/artifacts/last-run.log`.

Result classes are `PASS`/`FAIL` for the upgrade under test, `KNOWN` for a
pre-existing 2.x defect (reported but not fatal), and `NOTE` for things worth
seeing that are not pass/fail judgements. `ci/vm-migration-poc/README.md` covers
the setup, both env tables and the assertions in detail.

## Change Type

> None of the three is an exact fit: this is test tooling plus documentation and
> changes no runtime code. Nothing under `app/`, `lib/`, `bin/` or `config/` is
> touched — the diff is confined to `ci/` and `docs/`.

## Checklist

- [x] I have documented the `MANUAL.md` file with any changes to the user experience — *n/a, no user-facing behaviour changes*
- [ ] If my changes are non-trivial, I have added a changelog entry to notify users at `package/obs/rmt-server.changes` — **not added**; the harness is not shipped in the package. Happy to add one for `docs/upgrading-from-2x.md` if the documented workaround warrants it.

## Review


chcao:
Hi Experts, I'm testing this case, I got 'bash: ci/vm-migration-poc/run.sh: No such file or directory'
These are steps, please correct me if the testing steps have some problems, thanks.

**Environment:**
**Host:**
SLES15SP7, installed the Virtual Machine Manager, created a VM, the VM name is sle15sp7_rmt2_to_rmt3

**Guest** (the VM of sle15sp7_rmt2_to_rmt3): 
SLES15SP7, already setup a RMT server with rmt-server-2.28, enable and mirror some repos.

**Test steps:** 
**On Host:**
- run "zypper in rmt-server-2.28-1500700.3.23.1"
- run "yast rmt" to setup a rmt server
- run "rmt-cli sync" to sync with SCC
- run "zypper up rmt-server" to update the rmt-server package to 3.1.0 
- run VM=sle15sp7_rmt2_to_rmt3 ci/vm-migration-poc/run.sh 
  (and also tried with "SKIP_FETCH=1 SKIP_REFRESH=1 VM=sle15sp7_rmt2_to_rmt3  ci/vm-migration-poc/run.sh")

**The result is:** 
bash: ci/vm-migration-poc/run.sh: No such file or directory

Would you please hlep to take a look at if it is a bug or a problem of testing please?

PR #1543, opened 2026-08-14
Fix the RPM validation check for reserved paths

rtamalin:
## Description

The existing regexp check generates false positives as it will match any path ending with the reserved paths, not just the specific reserved paths.

Fixed by adding match at least one whitespace (`[[:space:]]`) character before the start of the path.

This fixes the CI feature test issue being hit by PR #1538.

### DIFF
--- a/ci/rmt-validate-packaging
+++ b/ci/rmt-validate-packaging
@@ -46,7 +46,7 @@ while IFS= read -r RPM; do
 
   # Check for forbidden directory ownership
   for dir_pattern in "${FORBIDDEN_DIRS[@]}"; do
-    if rpm -qplv "$RPM" 2>/dev/null | grep -E "^d.*${dir_pattern}"; then
+    if rpm -qplv "$RPM" 2>/dev/null | grep -e "^d.*[[:space:]]${dir_pattern}"; then
       echo "  ✗ ERROR: Package claims ownership of ${dir_pattern}"
       echo "    Top-level directories should be owned by filesystem package"
       ((ERRORS++))


ci/rmt-validate-packaging
```diff
@@ -46,7 +46,7 @@ while IFS= read -r RPM; do
 
   # Check for forbidden directory ownership
   for dir_pattern in "${FORBIDDEN_DIRS[@]}"; do
-    if rpm -qplv "$RPM" 2>/dev/null | grep -E "^d.*${dir_pattern}"; then
+    if rpm -qplv "$RPM" 2>/dev/null | grep -e "^d.*[[:space:]]${dir_pattern}"; then
```
jeremy-moffitt, line 49: is there always exactly 1 space or should this be suffixed with a `?` or `*` ?

rtamalin: There is always at least 1 whitespace character - there can be more than one from what I can see, and using `[[:space:]]` covers the eventuality of it being a space or a tab character. ;-)

rtamalin:
From my [comment](https://github.com/SUSE/rmt/pull/1538#issuecomment-5297425132) in #1538 where I demonstrated the fix:

To demonstrate, using a downloaded rmt-server-3.0.0 package from the 16.1 stream and checking for a path `/rmt` the old way yields the following:

```
30068d380cee:/ # rpm -qplv /var/cache/zypp/packages/SLE_BCI/x86_64/rmt-server-3.0.0-160099.1.10.x86_64.rpm | grep -E "^d.*/rmt$"
drwxr-xr-x    2 _rmt     nginx                       0 May 18 16:07 /etc/rmt
drwxr-xr-x    2 root     root                        0 May 18 16:07 /usr/lib64/rmt
drwxr-xr-x    2 root     root                        0 May 18 16:07 /usr/libexec/rmt
drwxr-xr-x    2 root     root                        0 May 18 16:07 /usr/share/rmt
drwxr-xr-x    2 root     root                        0 May 18 16:07 /usr/share/rmt/lib/rmt
drwxr-xr-x    2 _rmt     nginx                       0 May 18 16:07 /var/lib/rmt
30068d380cee:/ #
```

Adding a requirement to the regexp to match a whitespace before the target path yields the following:

```
30068d380cee:/ # rpm -qplv /var/cache/zypp/packages/SLE_BCI/x86_64/rmt-server-3.0.0-160099.1.10.x86_64.rpm | grep -e "^d.*[[:space:]]/rmt$"
30068d380cee:/ #
```

We also aren't using any extended regexp features, so we don't need the `-E`. ;-)

ci/rmt-validate-packaging
jeremy-moffitt: I had to lookup `[[:space:]]` for reviewing and all indications suggest it matches a single whitespace character (inclusive of tabs and other single-multispace characters):

```
[:space:]
matches a whitespace character, including a line break [ \t\r\n\v\f] (also written as \s)  
```
Are we sure there won't be multiple single spaces? when I play around with `[[:space:]]` on regex101 it shows multiple matches for multiple spaces.

brett060102 (approved):
Kooks good.

jeremy-moffitt (approved):

PR #1553, opened 2026-08-26
Fix ActiveRecord::Deadlocked on registration sharing push

jesusbv:
## Description

concurrent pushes for the same system deadlocked in
RmtToRmtController#create, returning 500 to the RMT sibling
(returning 500 is not allowed on RMT)

- add the new the activations instead of replacing them all
  An unchanged re-push now writes nothing at all, so the
  common case takes no activation locks
  Query count drops from N + 5 to 4 regardless of activation count

- run the transaction at read committed, which takes record locks only
  This also gives the loser of the race condition a fresh read view after
  lock! returns; under repeatable read it would wait

- retry the whole transaction up to 3 times with jittered backoff as a backstop
  beause InnoDB rolls a victim back whole

This Fixes bsc#1274092

* Related Issue / Ticket / Trello card: [bsc#1274092](https://bugzilla.suse.com/show_bug.cgi?id=1274092)

## How to test 

have two systems with the same login accesing the update server

## Change Type

## Checklist

## Review


rjschwei (approved):

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -3,32 +3,28 @@
...
-            service: product.service,
-            created_at: activation[:created_at]
-          )
-        end
+      attempts = 0
+      begin
+        attempts += 1
+        create_or_update_system
+      rescue ActiveRecord::Deadlocked, ActiveRecord::LockWaitTimeout => e
+        raise if attempts >= DEADLOCK_RETRIES
```
jcomello, line 20: Shouldn't we raise the error we rescue in the line above? 

```suggestion
        raise e if attempts >= DEADLOCK_RETRIES
```

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -40,25 +36,64 @@ def destroy
     protected
 
     def fetch_system
-      System.find_or_create_by(
+      credentials = {
```
jcomello, line 39: This is not wrong, but since we're creating another hash with the same keys we could use the `Hash#slice` method instead of recreating.

```suggestion
      credentials = params.slice(:login, :password, :system_token)
```
Or, if you prefer:


```suggestion
      credentials = params.slice(%i[login password system_token])
```

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
```diff
@@ -4,6 +4,20 @@
 # rubocop:disable Metrics/ModuleLength
 module RegistrationSharing
   RSpec.describe RmtToRmtController, type: :request do
+    # the controller opens its transaction with an explicit isolation level, and
+    # ActiveRecord refuses that inside the transaction that wraps each example
+    # ("cannot set transaction isolation in a nested transaction"), so this file
+    # cleans up after itself instead
+    self.use_transactional_tests = false
+
+    after do
```
jcomello, line 13: Rspec wasn't cleaning the database after every test? Is is necessary?

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
```diff
@@ -220,6 +234,102 @@ module RegistrationSharing
...
+          raise error if calls <= failures
+
+          original.call(*args)
+        end
+      end
+
+      {
+        'a deadlock' => ActiveRecord::Deadlocked,
+        'a lock wait timeout' => ActiveRecord::LockWaitTimeout
+      }.each do |description, error_class|
```
jcomello, line 275: Personally I don't like this kind of approach because of two reasons:
- It makes difficult to know the test in case it fails
- Rspec already provide shared_examples to do this.

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
```diff
@@ -220,6 +234,102 @@ module RegistrationSharing
...
+      # raises on the first 'failures' calls, then behaves normally
+      def fail_first(error, failures)
+        calls = 0
+        allow(System).to receive(:find_or_create_by).and_wrap_original do |original, *args|
+          calls += 1
+          raise error if calls <= failures
+
+          original.call(*args)
+        end
+      end
```
jcomello, line 270: I might not understand it correctly, but it can hide if a test fails, doesn't it?

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
```diff
@@ -40,25 +36,64 @@ def destroy
...
-        system_token: params.expect(:system_token)
-      )
+      System.find_by!(**credentials)
+    end
+
+    def create_or_update_system
+      # read_committed because under repeatable read the activations delete
+      # takes next-key locks covering the gaps a concurrent push then has to
+      # insert into; read_committed takes record locks only
+      System.transaction(isolation: :read_committed) do
```
jcomello, line 53: You're using System.transaction for a specific reason? Why not to use ActiveRecord::Base.transaction instead?

jcomello (commented):
Some questions and suggestions

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
jesusbv: if I dont, test failed

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
jesusbv: yes, I'm not a fan either, will use `shared_examples` which is also present on RMT, IIRC
(I blame claude)

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
jesusbv: I do not think so, I will double check again

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jesusbv: `system.lock!` creates or triggers `SELECT ... FOR UPDATE` through System's connection
On a different connection it would take the lock in a separate transaction

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jesusbv: yep, thought about that but that would generate a 

```ruby
ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash
```
i.e
```ruby
  1) RegistrationSharing::RmtToRmtController#create byos with correct credentials performs HTTP request successfully
     Failure/Error:
       post(
         '/api/regsharing',
         params: {
           login: login_byos,
           password: password,
           created_at: created_at,
           registered_at: registered_at,
           last_seen_at: last_seen_at,
           proxy_byos_mode: :byos,
           pubcloud_reg_code: pubcloud_reg_code,

     ActionController::UnfilteredParameters:
       unable to convert unpermitted parameters to hash
```
36 errors in our tests

same for 
```ruby
credentials = params.slice(:login, :password, :system_token)
```

and if I used any of those without `(**credentials)` but `(credentials)` that would also not work
```ruby
ActionController::Parameters {} permitted: false
```

jesusbv: now, `permit` would work i.e. 
```ruby
params.permit(:login, :password, :system_token)
```
but that would allow missing/absent keys, so if `system_token` is missing, it would search by `login` and `pass`
that is not unique by definition, see [here](https://github.com/SUSE/rmt/blob/rmt_3/db/schema.rb#L197)

now, we have changed that, making login unique per system but that change was introduced some time ago, meaning there could be systems in the DB using the previous non unique login and then that command would have unwanted consequences

jesusbv: to be explicit about using `permit`, we **potentially could** once we have purged all the systems from before the login unique change was introduced, again, potentially or we could be on the safe side and not thinking on changing this when purge and blablabla

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jesusbv: we do, bare `raise` inside a `rescue` block re-raises the same exception being handled in that block

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jcomello: the `permit` is not a good fit for this. To do that you'll need to use `.to_h`. Better keep the way we're now.

engines/registration_sharing/app/controllers/registration_sharing/rmt_to_rmt_controller.rb
jcomello: Ok, thanks

engines/registration_sharing/spec/requests/registration_sharing/rmt_to_rmt_controller_spec.rb
jcomello: This is weird because there `config.use_transactional_fixtures` is already set to `true` in `rails_helper`
In the future, we could consider using database cleaner or a similar library.

PR #1556, opened 2026-08-31
Fix mirror 1276939

jesusbv:
## Description

fix for mirroring

libcurl uses HTTP/2 as default, because of that
RMT resumes a Fiber from inside an FFI callback
that libcurl called while its own C frames are
live on the stack
the Fiber switch swaps the machine stack out
from under those frames, arbitrary/random Ruby runs
and then the next allocation trips a GC that finds a corrupt heap

that behaviour produces the traceback observed in the bugzilla entry

force HTTP/1.1 for Typhoeus

Fixes bsc#1276939

* Related Issue / Ticket / Trello card: [bsc#1276939](https://bugzilla.suse.com/show_bug.cgi?id=1276939)

## How to test 

re run the mirror command, no bug should be raised and mirroring should complete fine

## Change Type

## Checklist

## Review

### DIFF
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -23,6 +23,7 @@ class RMT::HttpRequest < Typhoeus::Request
     options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) || 512
     options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) || 120
     options[:accept_encoding] = 'gzip'
+    options[:http_version]    = :httpv1_1
   end
 
 end
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -23,7 +23,12 @@ class RMT::HttpRequest < Typhoeus::Request
     options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) || 512
     options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) || 120
     options[:accept_encoding] = 'gzip'
-    options[:http_version]    = :httpv1_1
+    # HTTP/2 multiplexes several streams into one curl_multi_perform frame, so
+    # libcurl can fire callbacks for multiple easy handles while its own frames
+    # are live, RMT::FiberRequest resumes fibers from those callbacks, which
+    # corrupts the heap
+    # Set the HTTP version to 1.1 until that is fixed, see bsc#1276939
+    options[:http_version] = (Settings.try(:http_client).try(:http_version) || 'httpv1_1').to_sym
   end
 
 end
--- a/lib/rmt/http_request.rb
+++ b/lib/rmt/http_request.rb
@@ -28,7 +28,7 @@ class RMT::HttpRequest < Typhoeus::Request
     # are live, RMT::FiberRequest resumes fibers from those callbacks, which
     # corrupts the heap
     # Set the HTTP version to 1.1 until that is fixed, see bsc#1276939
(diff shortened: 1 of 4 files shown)


lib/rmt/http_request.rb
```diff
@@ -23,6 +23,7 @@ def set_defaults
     options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) || 512
     options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) || 120
     options[:accept_encoding] = 'gzip'
+    options[:http_version]    = :httpv1_1
```
jesusbv: maybe we should make this setting configurable to be read from `rmt.conf` / `Settings`

for when we want to switch to HTTP/2

Thoughts ?

jesusbv: done

package/obs/rmt.conf
```diff
@@ -30,6 +30,7 @@ http_client:
   proxy_password:
   low_speed_limit: 512
   low_speed_time: 120
+  http_version: httpv1_1
```
rtamalin: The one question I would have is whether we need this change for the RMT server itself? What we have seem so far is that rmt-cli mirror (and possibly purge?) are hitting problems for libcurl managed connections...

Which I don't think is leveraged by the core RMT server?

rtamalin: So the associated code is leveraged by the main RMT server code, but as per other discussions the impact (slightly longer connection establishment time, and managing extra connections) are relatively low/trivial in impact.

lib/rmt/http_request.rb
```diff
@@ -23,6 +23,12 @@ def set_defaults
     options[:low_speed_limit] = Settings.try(:http_client).try(:low_speed_limit) || 512
     options[:low_speed_time] = Settings.try(:http_client).try(:low_speed_time) || 120
     options[:accept_encoding] = 'gzip'
+    # HTTP/2 multiplexes several streams into one curl_multi_perform frame, so
+    # libcurl can fire callbacks for multiple easy handles while its own frames
+    # are live, RMT::FiberRequest resumes fibers from those callbacks, which
+    # corrupts the heap
+    # Set the HTTP version to 1.1 until that is fixed, see bsc#1276939
+    options[:http_version] = :httpv1_1
```
jesusbv, line 31: we could set this only in the downloader but this issue may happen in other places where it is used
so better be safe

rtamalin (commented):
Looks good to me now, and is fundamentally the same solution I identified in my investigations last Friday.

I added @digitaltom and @ngetahun as reviewers in case they are aware of any subtleties that we missed.

digitaltom:
@jesusbv @rtamalin I did some further debugging into this issue with help of Claude. 
I don't think downgrading to http 1.1 is the proper fix, it probably just narrows the chances for it to happen. 

Instead, I'd like to try simplifying our Fiber handling, dropping the `on_headers` callback and only resuming on `on_complete`. Branch: https://github.com/SUSE/rmt/compare/rmt_3...no_fiber_resume_on_headers?expand=1

jesusbv:
> @jesusbv @rtamalin I did some further debugging into this issue with help of Claude. I don't think downgrading to http 1.1 is the proper fix, it probably just narrows the chances for it to happen.
> 

yes, I was under the believe that Fergal test on Friday last week was a confirmation of this change
apparently it is a false positive, so I will close this PR

> Instead, I'd like to try simplifying our Fiber handling, dropping the `on_headers` callback and only resuming on `on_complete`. Branch: https://github.com/SUSE/rmt/compare/rmt_3...no_fiber_resume_on_headers?expand=1

that LGTM

PR #1569, opened 2026-09-09
Eliminate use of Fibers in RMT::Downloader

rtamalin:
## Description

Simplify the multi-file downloader, replacing the existing fiber-based RMT::FiberRequest with a solution that leverages RMT::HttpRequest in the callbacks for a Typheous::Hydra managed queue of download requests.

Implementation Details:
- Limit initial queue size based on concurrency settings to avoid file descriptor starvation issues, queuing additional requests when an existing one finishes
- Refactor the download flow to use on_body/on_complete callbacks on RMT::HttpRequest instances queued via process_queue
- Fail immediately on the first error when ignore_errors is false, rather than continuing to process remaining files in the queue
- Add test for stale cache re-download scenario
- Consolidate and deduplicate downloader tests (stub_logger helper, merged describe blocks for private methods)

Delete the existing FiberRequest implementation:
- Delete defunct lib/rmt/fiber_request.rb and associated tests.
- Update .rubocop_todo.yml to remove deleted fiber_request.rb entries

Fixes: bsc#1276939

AI-Generated-By: Qwen 3.6 35B Moe (A3B)
Co-Authored-By: Thomas Schmidt <tschmidt@suse.com>
Signed-off-by: Fergal Mc Carthy <fmccarthy@suse.com>

## How to test 

It should be possible to mirror a reasonable amount of RPMs such as those associated with products 2609 (SLE 15 SP6), 2793 (SLE 15 SP7), and 2930 (SLE 16.0) repeatedly and reliably without coredumping or hanging.

## Change Type

## Checklist

## Review


rtamalin:
Local testing on my workstation and in one of my reproducer VM environments is looking good, however it sometimes took 4+ runs to see failures with the v3.1.0 version of the code, so I'll keep running testing iterations for now.

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,93 @@ def download_multi(files, ignore_errors: false)
...
 
-  def copy_from_cache(file)
-    unless (file.cache_path == file.local_path)
-      make_file_dir(file.local_path)
-      FileUtils.cp(file.cache_path, file.local_path, preserve: true)
+  def handle_response(response, downloaded_file, file, retries)
+    if invalid_response?(response)
+      downloaded_file.unlink
+      if retries.zero? || response.code == 404
+        if @failed_downloads
```
felixsch:
```suggestion
        unless @failed_downloads.nil?
```

I think this makes it more clear what it is doing.

lib/rmt/downloader.rb
```diff
@@ -234,6 +155,9 @@ def handle_checksum_verification!(checksum_type, checksum_value, download_path)
   end
 
   def invalid_response?(response)
+    return true if response.nil?
```
felixsch: This makes me wonder what is going on, did you run into this issue or is this something the model added itself. If a `response.nil? == true` then the whole following workflow is breaking since it assumes that the `response` object is actually available.

lib/rmt/downloader.rb
```diff
@@ -234,6 +155,9 @@ def handle_checksum_verification!(checksum_type, checksum_value, download_path)
...
+    return false if response.code == 0 && response.return_code == :ok
```
felixsch, line 174: What is the case where `code = 0` and `return_code = :ok` can you maybe document why this is added.

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,93 @@ def download_multi(files, ignore_errors: false)
...
-    failed_files = []
-    cache_requests.each do |file, request|
-      next downloads_needed << file if request.nil?
-      next downloads_needed << file unless valid_cached_file?(file, request.response)
-
-      copy_from_cache(file)
-    rescue RMT::Downloader::Exception => e
-      next failed_files << file.local_path if ignore_errors
 
+    request.on_complete do |response|
```
felixsch: Maybe this is a chance to move away from the `on_complete` implementation. I wonder why it was implemented this way, while Typhoeus implements the full Promise workflow (`on_success` and `on_failure`) and the current implementation does nothing else than the provided implementation using `invalid_response?`

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,93 @@ def download_multi(files, ignore_errors: false)
...
-  end
-
-  def valid_cached_file?(file, response)
-    RMT::Downloader::Exception.raise_request_error(file.remote_path, response, @logger) if invalid_response?(response)
-
-    # response.headers returns Typhoeus::Response::Headers, which takes care of
-    # case-sensitive concerns with the header's key
-    last_modified_header = response.headers['Last-Modified']
-    return false unless last_modified_header
+  def process_queue
```
felixsch: How about renaming this to `enqueue_next` instead of `process_queue` to make it more clear what the action is doing?

digitaltom (approved):

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,114 @@ def download_multi(files, ignore_errors: false)
...
-
-    file.cache_timestamp == Time.parse(last_modified_header).utc
+  def handle_response(response, downloaded_file, file, retries)
+    if invalid_response?(response)
+      downloaded_file.close!
+      begin
+        raise_request_error(file.remote_path, response)
+      rescue RMT::Downloader::Exception => e
+        handle_failure(file, retries, e)
+      end
```
mssola: I'm lost. What is this block of code? I'm not at all that familiar with this part of RMT, but to me it reads like:

1. Call a method that raises an exception.
2. Catch it

Isn't it weird?

lib/rmt/downloader.rb
mssola: Agreed. This confused me along the review 👍🏻

mssola (commented):
To be honest, I'm not that familiar with this part of RMT. So I'm hesitant on giving a thumbs up/down.

That being said, do we have benchmarks on this? Also, at a quick glance I actually did not see the code trying to download multiple files concurrently (I've seen it sequentially). Hence, even dumber question, do we even need all this concurrency? :D

lib/rmt/downloader.rb
rtamalin: So this was the name in the fiber implementation and my stated goal to the AI assistant was to create a drop in replacement for the existing fiber based implementation, so it tended to keep the names of things much the same, even for purely internal stuff. ;-)

Happy to rename.

lib/rmt/downloader.rb
rtamalin: So this is partly my fault as this code evolved through many iterations of trying to restructure the code to use fibers "safely" before giving up, and this guard was added because at one point it was relevant during one of those iterations when one of the calling paths could have a nil response.

Looks like it can be safely removed now.

lib/rmt/downloader.rb
rtamalin: And the return false in this case was one that the AI assistance added unnecessarily.
So it can also be removed.

lib/rmt/downloader.rb
rtamalin: This one we probably want @digitaltom to weigh in on... His [updates to restore the error handling that fall outside the response](https://github.com/SUSE/rmt/commit/a5bee15c54ad5dd113d0bdb81901c0189a970eaa) created raise then catch structure.

Looking at what is effectively happening it here it seems that the construct is being used to redirect specific exceptions to be handled by the handle_failure() routine... For the `if invalid_response?(response)` case it looks like the raise and catch is being used to create a new RMT::Downloader::Exception for a failed response, and log it, but then catch it so that we can pass it to the handle_failure...

For the `if invalid_response?()` case that could be replaced by changing RMT::Downloader::Exception to have a create_request_error() that does the logging, message construction and returns the result of the .new() call, and then just call that to create the error to pass to handle_failure(), rather than raising and catching it to create it.

Will look into cleaning it up that way...

lib/rmt/downloader.rb
rtamalin: Actually there is a valid reason for this guard, so I will add a comment explaining why it is there.

lib/rmt/downloader.rb
rtamalin: Not sure why this got resolved, as the associated code still exists though Tom's commits did change the context around it. If I did it, I certainly didn't mean to. ;-)

As for the if vs unless, coming from a primarily non-Ruby background, unless checks are usually less clear to me that if checks.

However, in this case I think you are right now that I review the subtleties of what `if @failed_downloads` is actually checking.

I'll incorporate it into my upcoming revisions to address review feedback.

rtamalin: Ok, so rubocop didn't like it. Will make the if check more explicit ;-)

lib/rmt/downloader.rb
rtamalin: Renamed.

lib/rmt/downloader.rb
rtamalin: I restructured the code to make the intent more explicit and keep Rubocop happy ;-)

lib/rmt/downloader.rb
rtamalin: Comment added.

lib/rmt/downloader.rb
rtamalin: I've implemented the create_request_error() in RMT::Downloader::Exception and updated the RMT::Downloader code to leverage it, which allowed me to simplify this strange pattern of catching the exception raised by raise_request_error() to create an error. ;-)

lib/rmt/downloader.rb
rtamalin: Done.

rtamalin:
I have updated the PR to address review feedback.

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,106 @@ def download_multi(files, ignore_errors: false)
...
   end
 
-  def copy_from_cache(file)
-    unless (file.cache_path == file.local_path)
-      make_file_dir(file.local_path)
-      FileUtils.cp(file.cache_path, file.local_path, preserve: true)
+  # retries the file, or records/raises the failure, depending on 'ignore_errors'
+  def handle_failure(file, retries, error)
+    if retries.zero? || error.try(:http_code) == 404
+      if @failed_downloads.nil?
```
ngetahun, line 105: The rails-y way would be to use `@failed_downloads.present?`

lib/rmt/downloader.rb
ngetahun, line 20: Should we also change this to detect the number of cores with something like
```ruby
require 'concurrent-ruby'
Concurrent.processor_count
```

lib/rmt/downloader.rb
```diff
@@ -36,171 +35,106 @@ def download_multi(files, ignore_errors: false)
...
+    handle_failure(file, retries, e)
   end
 
-  def cache_head_request(file)
-    # RMT must not make HEAD requests when importing repos (file://)
-    return nil unless %w[http https].include?(file.remote_path.scheme)
-    return nil if file.cache_timestamp.nil?
+  def enqueue_next
+    queue_item = @queue.shift
+    return unless queue_item
```
ngetahun:
```suggestion
    return unless queue_item.present?
```

ngetahun (changes requested):
I think the file caching mechanism should be refactored into its own class/concern. One code smell is that there are similar looking `handle_*`  and `process_*`  functions working with different objects.

jesusbv (approved):
LGTM

jesusbv:
@rtamalin I saw the tests changes, I guess we do have a test that covers the scenario of the bug ?

lib/rmt/downloader.rb
rtamalin: Yeah I was checking into this yesterday as the concurrency setting is hardcoded and not linked to anything; this is the same as it was previously.

Just mapping directly to the processor count however would be a poor idea, as on a single core system you would have just a single downloader running, so you would want to have it set up a lower bound of 4, so that you get a similar level of concurrency to what we currently have. But on the same not if it is running on a system with lots of cores you may not want to run with a concurrency setting of 32 if you have 32 cores as that could generate excessive network and disk activity.

I think this is something that we should look into after we get this bug fix out the door.

rtamalin:
> I think the file caching mechanism should be refactored into its own class/concern. One code smell is that there are similar looking `handle_*` and `process_*` functions working with different objects.

@ngetahun given that this is a priority bugfix that we are trying to get released, and the existing names are pretty much the same as they were before, I'd prefer to do any refactoring like this as part of a follow-up effort.

lib/rmt/downloader.rb
rtamalin: That isn't what this is checking for though... The value of @failed_downloads is set to nil if ignore_errors is false when setting things up in download_multi(), so this check is explicitly checking for nil to see if we are in ignore_errors=false mode, with the else case handling things if we are ignoring errors.

lib/rmt/downloader.rb
rtamalin: Rubocop was unhappy with the `return unless queue_item.present?` and wants `return if queue_item.blank?` ;-)

rtamalin:
> @rtamalin I saw the tests changes, I guess we do have a test that covers the scenario of the bug ?

@jesusbv The bug scenario turned out to be the use of fibers; our original code flow "tickled" memory corruption issues associated with the fiber pool scheduler code and the Fiber.yield call, partially aggravated by some missing Garbage Collection (GC) guard checks in the Fiber implementation, and the fact that the original code was recursively creating new fibers to handle retries, rather the handling them within the original fiber context, as well as potentially Fiber.yield'ing unnecessarily for the headers handler, all potentially contributed to the corruption issues.

However, even when testing with the latest unreleased fix for Fiber GC guard issues in a patch Ruby, and having eliminated the recursive fiber creation and unnecessary extra yielding, we still saw the memory corruption issues (it happened much faster/more reliably) so the solution settled upon was to get rid of Fibers. ;-)

lib/rmt/downloader.rb
ngetahun: Now I am confused, this line https://github.com/SUSE/rmt/blob/dca74eb876e731f2a98e6ff84db7f677fb50ff59/lib/rmt/downloader.rb#L109-L110 suggests `@failed_downloads` is a list. If not then we need to rework the approach as this kind of internals is a code smell.

lib/rmt/downloader.rb
ngetahun: I dont mean that we need to occupy all cores to run the download. But we should ensure a dynamic value that properly utilizes the existing multi-processing capabilities. This will be noticable for the cloud team.

lib/rmt/downloader.rb
ngetahun: yeah, present? and blank? are basically the same. But it is much better than relying on the object comparison methods,

ngetahun (approved):
LGTM!

lib/rmt/downloader.rb
rtamalin: Yeah my thought is to also make it configurable via config setting.
The PubCloud team seem to have been OK with the existing 4, and if we define a new sensible initial default value that is derived from system capabilities that can be overridden by a config setting if desired, that will likely cover all the bases.
I've added it to our jsc#SCC-1008 follow-up work ticket.

lib/rmt/downloader.rb
rtamalin: This is pretty much how the original code worked... I've added this to the jsc#SCC-1008 follow up work ticket.

PR #1570, opened 2026-09-09
Do not return 500s registry

jesusbv:
## Description

registry in RMT must not return 500 to the client

## How to test 

try the registry with a misconfigured update server or wrong registry command and it should not return 500

## Change Type

## Checklist

## Review

### DIFF
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -26,6 +26,11 @@ module Registry
       render json: { code: :bad_request, error: error.message }, status: error.status
     end
 
+    rescue_from Registry::Exceptions::RegistryUnavailable do |error|
+      logger.error("Registry is unavailable: #{error.message}")
+      render json: { code: :unauthorized, error: 'Registry is unavailable' }, status: error.status
+    end
+
     # AuthZ handler
     # AuthZ will validate which of the requested scope policies are fulfilled
     # with the current login access and prepare the token to be sent back to the client
--- a/engines/registry/app/controllers/registry/registry_controller.rb
+++ b/engines/registry/app/controllers/registry/registry_controller.rb
@@ -8,6 +8,24 @@ module Registry
     before_action :basic_auth, except: [ :catalog ]
     before_action :catalog_token_auth, only: [ :catalog ]
 
+    rescue_from StandardError do |error|
+      logger.error("Registry request failed: #{error.class}: #{error.message}")
+      logger.error(error.backtrace.join("\n")) if error.backtrace
+      render json: { code: :unauthorized, error: 'Registry authentication failed' }, status: :unauthorized
+    end
+
+    rescue_from Registry::Exceptions::RegistryUnavailable do |error|
+      logger.error("Registry is unavailable: #{error.message}")
+      render json: { code: :unauthorized, error: 'Registry is unavailable' }, status: error.status
+    end
+
+    # the scope comes straight off the query string, so a malformed one is the
+    # client's mistake to correct and the message is safe to hand back
+    rescue_from Registry::Exceptions::InvalidScope do |error|
+      logger.info("Rejected registry scope: #{error.message}")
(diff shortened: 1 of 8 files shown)


jesusbv:
separated or split into commits for easier review, tests have their own commit, again for reviewing purposes

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -8,15 +8,41 @@ class RegistryController < Registry::ApplicationController
     before_action :basic_auth, except: [ :catalog ]
     before_action :catalog_token_auth, only: [ :catalog ]
 
+    rescue_from StandardError do |error|
```
rtamalin, line 11: Minor suggestion, not required but may be nice to have, given that rescue_from precedence is bottom to top of file, rather than top to bottom, in precedence.

Potentially add a comment along the lines of:

```suggestion
    # default rescue handler if a more specific one below doesn't match
    rescue_from StandardError do |error|
```

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -8,15 +8,41 @@ class RegistryController < Registry::ApplicationController
...
+      logger.error("Registry request failed: #{error.class}: #{error.message}")
+      logger.error(error.backtrace.join("\n")) if error.backtrace
+      render json: { code: :unauthorized, error: 'Registry authentication failed' }, status: :unauthorized
```
rtamalin, line 14: This is the same failure as for the RegistryAuthError below... Do we maybe want the wording of the error message to be somewhat different so that we can differentiate the two scenarios from the requesting side?

Perhaps something like this?
```suggestion
      render json: { code: :unauthorized, error: 'Registry request failed' }, status: :unauthorized
```

engines/registry/spec/app/controllers/registry_controller_spec.rb
```diff
@@ -95,9 +149,27 @@ module Registry
       let(:system) { create(:system) }
       let(:auth_headers) { { 'Authorization' => ActionController::HttpAuthentication::Basic.encode_credentials(system.login, system.password) } }
 
-      it 'raise an error with a clear message' do
+      it 'refue the request instead of raising' do
```
rtamalin, line 152:
```suggestion
      it 'refuse the request instead of raising' do
```

rtamalin (commented):
Looks good - spotted a minor typo and made a couple of minor suggestions.

engines/registry/app/controllers/registry/registry_controller.rb
jesusbv: yes, it should match the error message logged, copypasta is dangerous

engines/registry/app/controllers/registry/registry_controller.rb
```diff
@@ -8,15 +8,41 @@ class RegistryController < Registry::ApplicationController
...
+    end
+
+    rescue_from Registry::Exceptions::RegistryUnavailable do |error|
+      logger.error("Registry is unavailable: #{error.message}")
+      render json: { code: :unauthorized, error: 'Registry is unavailable' }, status: error.status
```
rjschwei, line 19: When the registry is not available then this is a `bad request`

rjschwei (changes requested):

engines/registry/app/controllers/registry/registry_controller.rb
jesusbv: this was or is a tricky one, because I did consider `bad request` but wouldn't that be confusing for the user ?
like the request is actually OK, and yes, I think similar argument can be made for unauth, credentials are fine so unauth could be confusing

rjschwei: Well the client is asking for something that does not exist, as such it is a `bad request`. However, I can see where this could lead the client down the wrong path. As such `not found` is probably better here.

rjschwei (approved):

